PR #5, opened 2021-03-15
Add CI using Github Actions and Windows-related fixes to setup.py and DLL loading

brendan-ward:
This enables fast testing in Linux using the GDAL Docker image.

This build and manual testing on Windows, but automated testing via pytest is still failing due to what appears to be a corrupt environment.  DLLs that load properly in manual testing fail when loaded within a pytest environment.

On Windows with Python >= 3.8, this uses a similar trick as Fiona to look for the GDAL DLL in the `PATH` and use `os.add_dll_directory` to temporarily add that to the search path used for loading DLLs.

Note: this makes geopandas an optional dependency.


pyogrio/geopandas.py
```diff
@@ -62,9 +61,10 @@ def read_dataframe(
     GeoDataFrame or DataFrame (if no geometry is present)
     """
     try:
-        import pandas as pd
-        import geopandas as gp
-        from geopandas.array import from_wkb
+        with GDALEnv():
+            import pandas as pd
+            import geopandas as gp
+            from geopandas.array import from_wkb
```
jorisvandenbossche, line 67: @brendan-ward do you remember why geopandas also needs to be imported in a GDALEnv context? 
(I understand that we need this for our own _ogr functions, but don't understand that for external packages)

I suppose there might be some interference with the fact that geopandas will try to import fiona? (we could also change that to make that in geopandas a really lazy import only when doing read_file/to_file the first time)

PR #10, opened 2021-04-02
Autodetect GDAL_VERSION on Windows using gdalinfo

### LINKED ISSUE #9: Attempt to determine GDAL_VERSION automatically on Windows
On Linux, we use `gdal-config` to obtain info about GDAL: version, location of libraries, etc.

On Windows, there is no counterpart, and the user needs to provide these via env variables and command line parameters.

We could use an attempt to execute `gdalinfo --version` to try and determine `GDAL_VERSION` within setup.py to obtain this automatically; where this fails user will have to provide `GDAL_VERSION` as a variable.

Inspired by: https://github.com/conda-forge/pyogrio-feedstock/pull/9

brendan-ward:
Resolves #9

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import os
 from pathlib import Path
 import platform
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -78,18 +79,44 @@ def read_response(cmd):
...
+
+                    if gdalinfo_path:
+                        print(f"Found gdalinfo at {gdalinfo_path}")
+                        raw_version = read_response([gdalinfo_path, "--version"]) or ""
+                        # gdal_version_str = raw_version.split(',')[0].split(' ')[1]
+                        m = re.search("\d+\.\d+\.\d+", raw_version)
+                        if m:
+                            gdal_version_str = m.group()
+                    else:
+                        print("Unable to find gdalinfo")
```
jorisvandenbossche, line 105: just a question: do you have some rationale for when to use print and when to use logging?

brendan-ward: Oops - these were for debugging that I forgot to remove before merging.

For logging in setup.py I was roughly following what Fiona was doing.

PR #18, opened 2021-09-15
DOC: basic API docs

martinfleis:
Setting up basic API documentation hosted on ReadTheDocs.

### DIFF
--- a/docs/source/api.md
+++ b/docs/source/api.md
@@ -1,0 +1,17 @@
+# API reference
+
+## Core
+
+```{eval-rst}
+.. automodule:: pyogrio
+   :members: list_drivers, list_layers, read_bounds, read_info, set_gdal_config_options, get_gdal_config_option, __gdal_version__, __gdal_version_string__
+```
+
+
+## GeoPandas integration
+
+```{eval-rst}
+.. automodule:: pyogrio
+    :members: read_dataframe, write_dataframe
+    :noindex:
+```
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -25,8 +25,8 @@ autodoc_mock_imports = [
 # -- Project information -----------------------------------------------------
 
 project = "pyogrio"
-copyright = "2021, Brendan C. Ward"
-author = "Brendan C. Ward"
+copyright = "2020-2021 Brendan C. Ward and pyogrio contributors"
+author = "Brendan C. Ward and pyogrio contributors"
 
 # The full version, including alpha/beta/rc tags
 release = __version__
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,0 +1,67 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+import sphinx_rtd_theme
+import pyogrio  # noqa
+
+
+autodoc_mock_imports = [
+    "geopandas",
+    "pygeos",
+]
+
+# -- Project information -----------------------------------------------------
+
+project = "pyogrio"
+copyright = "2021, Brendan C. Ward"
+author = "Brendan C. Ward"
+
+# The full version, including alpha/beta/rc tags
+release = pyogrio.__version__
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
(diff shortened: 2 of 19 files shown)


martinfleis:
This wasn't as straightforward as I hoped but it builds now https://pyogrio--18.org.readthedocs.build/en/18/index.html

docs/source/conf.py
```diff
@@ -0,0 +1,67 @@
...
+autodoc_mock_imports = [
+    "geopandas",
+    "pygeos",
+]
+
+# -- Project information -----------------------------------------------------
+
+project = "pyogrio"
+copyright = "2021, Brendan C. Ward"
+author = "Brendan C. Ward"
```
brendan-ward, line 29: Can this be left out?  I don't see that this is used in Geopandas' conf.py, so we probably don't need it here.  Otherwise could be expanded to `Brendan C. Ward and pyogrio contributors`

docs/source/index.md
```diff
@@ -0,0 +1,22 @@
+# pyogrio - Vectorized spatial vector file format I/O using GDAL/OGR
+
+Pyogrio provides a
+[GeoPandas](https://github.com/geopandas/geopandas)-oriented API to OGR vector
+data sources, such as ESRI Shapefile, GeoPackage, and GeoJSON. This converts to
+/ from `GeoPandas` `GeoDataFrame`s when the data source includes geometry and `Pandas`
+`DataFrame`s otherwise.
```
brendan-ward, line 7: Consider adding something about the value-add of using pyogrio:

```
Pyogrio uses a vectorized approach for reading and writing DataFrames, resulting in >5-10x speedups reading files and >5-20x speedups writing files compared to using non-vectorized approaches (Fiona and current I/O support in GeoPandas).
```

docs/source/install.md
```diff
@@ -0,0 +1,99 @@
...
+
+This package is available on [conda-forge](https://anaconda.org/conda-forge/pyogrio)
+for Linux, MacOS, and Windows.
+
+```bash
+conda install -c conda-forge pyogrio
+```
+
+This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
+raw I/O support and `geopandas`, `pygeos` and their dependencies for GeoDataFrame
```
brendan-ward, line 22:
```suggestion
raw I/O support and `geopandas`, `pygeos`, and their dependencies for GeoDataFrame
```

docs/source/api.md
```diff
@@ -0,0 +1,17 @@
+# API reference
```
brendan-ward, line 1: What do you think about using RST files as the standard under `docs/source` similar to GeoPandas?  I don't have a strong preference, but it seems like this would sidestep the need to add the `{eval-rst}` context here.

brendan-ward (commented):
Thanks @martinfleis for working on this!  Having docs available is a great addition; it also highlights some of the areas where our docstrings need a bit of work.  :)

I'm not very familiar with setting up Sphinx, so I'm not able to provide much feedback about that here other than that it appears to be working properly in most cases.

It looks like perhaps some of our docstrings are triggering weird formatting?  See the return statement [here](https://pyogrio--18.org.readthedocs.build/en/18/api.html#pyogrio.read_info)
- it shouldn't be highlighting the opening curly brace, right?  (not that you should fix docstrings here, just wanted to flag in case it was a Sphinx config issue).

Do you see any easy ways to reduce the duplication of content between markdown files at the project root and markdown / RST files used for building docs?  For instance, could we instead have a `installing.md` at the project root and link to that from README (and drop that section from README), and then read from that when building docs, instead of having a dedicated docs/sourdce/install.md file?   I'm not familiar enough with Sphinx to know what is easy / possible - just looking to centralize the updates we'd have to make in maintaining these...

docs/source/install.md
```diff
@@ -0,0 +1,99 @@
...
+I/O support.
+
+### PyPi
+
+This package is not yet available on PyPi because it involves compiled binary
```
jorisvandenbossche, line 27:
```suggestion
This package is not yet available on PyPI because it involves compiled binary
```

jorisvandenbossche:
> It looks like perhaps some of our docstrings are triggering weird formatting? See the return statement [here](https://pyogrio--18.org.readthedocs.build/en/18/api.html#pyogrio.read_info)
> 
> * it shouldn't be highlighting the opening curly brace, right?  (not that you should fix docstrings here, just wanted to flag in case it was a Sphinx config issue).

Yes, that's because of the formatting (the way the dictionary with curly braces is formatted now gives unexpected whitespace/indentation for text interpreted as rst (not as code)). I think doing something like the following should fix that:

```
    Returns
    -------
    dict
        A dictionary with the following keys::

            {
                "crs": "<crs>",
                "fields": <ndarray of field names>,
                "encoding": "<encoding>",
                "geometry": "<geometry type>",
                "features": <feature count>
            }
```

docs/source/api.md
martinfleis: I have changed to have this one a pure RST. Since there's no text it doesn't matter so much.

martinfleis:
I usually prefer succinct ReadMe with details in the docs. As the project grows, documentation gives you more options on how to format and split the information. So in the long term, I would move everything from ReadMe to docs and keep it much shorter, similarly to what we have in geopandas, xyzservices or contextily.

brendan-ward (approved):
This looks great - thank you so much @martinfleis !  🎉 

We can update the note about availability on PyPI (i.e., as a source distribution) in docs / readme after this is merged.

Issue #21, opened 2021-11-05
ENH: support URI schemes like zip:// (in addition to GDAL's vsi)


jorisvandenbossche:
Currently reading a file like the following fails (file path is the output of `geopandas.datasets.get_path("nybb")`):

```
In [8]: pyogrio.read_dataframe('zip:///home/joris/scipy/geopandas/geopandas/datasets/nybb_16a.zip')
...
DriverError: zip:///home/joris/scipy/geopandas/geopandas/datasets/nybb_16a.zip: No such file or directory
```

but using GDAL's native vsi pattern works:

```
In [10]: pyogrio.read_dataframe('/vsizip//home/joris/scipy/geopandas/geopandas/datasets/nybb_16a.zip')
Out[10]: 
   BoroCode       BoroName     Shape_Leng    Shape_Area                                           geometry
0         5  Staten Island  330470.010332  1.623820e+09  MULTIPOLYGON (((970217.022 145643.332, 970227....
1         4         Queens  896344.047763  3.045213e+09  MULTIPOLYGON (((1029606.077 156073.814, 102957...
2         3       Brooklyn  741080.523166  1.937479e+09  MULTIPOLYGON (((1021176.479 151374.797, 102100...
3         1      Manhattan  359299.096471  6.364715e+08  MULTIPOLYGON (((981219.056 188655.316, 980940....
4         2          Bronx  464392.991824  1.186925e+09  MULTIPOLYGON (((1012821.806 229228.265, 101278...

```

Given that those VSI paths are quite uncommon for general users, I think it would be nice to support the more general URI schemes as well. 
In Fiona this is handled by their `path` module: https://github.com/Toblerity/Fiona/blob/885dbd11c5dcb2c61862faefc28bfbbff99954de/fiona/path.py

jorisvandenbossche:
(automatically detecting a zip file would also be nice, but that's a separate feature request)

jamesmyatt:
Would be good to implement fsspec support for general URIs.

jorisvandenbossche:
@jamesmyatt yes indeed! I was actually looking at that as well, and the main issue there is to allow reading from byte blobs / file like object (instead of just string paths). This is not yet implemented I think, and was writing an issue for that. 

I think it is still good to use GDAL for those types it supports, but have fsspec available as the general solution (either included in pyogrio, or either explicitly used by the user, which is to be discussed)

jamesmyatt:
[...]

PR #29, opened 2022-01-15
ENH: pass through vsi schemas like S3

martinfleis:
GDAL natively supports [more](https://gdal.org/user/virtual_file_systems.html) vsi schemes than `vsizip`. If we don't necessarily check for that one, you can already use storages like S3 out of the box.

```py
pyogrio.read_dataframe("/vsis3/{bucket}/{filename}")
pyogrio.write_dataframe(df, "/vsis3/{bucket}/{filename}", driver="GeoJSON")
```

I think that it would also be nice to support the `s3://{bucket}/{filename}` format as well, as we do in GeoPandas (Fiona) (ref #21). 

I just don't know how to test this. I tried the `moto` way of mocking S3 but it seems that GDAL somehow ignores that. I'd rather avoid setting up a bucket as[ Fiona did](https://github.com/Toblerity/Fiona/blob/885dbd11c5dcb2c61862faefc28bfbbff99954de/tests/test_vfs.py#L182-L187).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,6 @@
 *.dll
 
 */__pycache__/*
-build/*
-*.egg-info/*
 .benchmarks/*
 .coverage
 .pytest_cache/*
@@ -19,3 +17,21 @@ benchmarks/fixtures/*
 .libs
 
 docs/build
+
+# Distribution / packaging
+.Python
+env/
+venv*/
+build/
+dist/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -84,7 +84,7 @@ def read_dataframe(
     path = str(path)
 
     if not "://" in path:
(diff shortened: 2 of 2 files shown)


martinfleis:
Note: I wanted this to be able to properly test dask-geopandas IO by reading a large file from my S3 bucket on a coiled cluster but I quickly figured out that it is not possible at the moment.

brendan-ward (approved):
Thanks @martinfleis !

I'm a bit confused by 
> not possible at the moment

Do you mean simply the missing support you added here, or did you run into other issues once you made these changes?

Since we're not yet enabling any special logic for S3 data sources, perhaps we can defer the need for testing those to GDAL for the time being?  Then if / when we add variants of `zip://` and `s3://` we add tests for those more specifically, since more of that would be custom logic on our end..

martinfleis:
> Do you mean simply the missing support you added here, or did you run into other issues once you made these changes?

Yes, I was just missing this. Otherwise I haven't run into any issues with the change in this PR.

> Since we're not yet enabling any special logic for S3 data sources, perhaps we can defer the need for testing those to GDAL for the time being? Then if / when we add variants of zip:// and s3:// we add tests for those more specifically, since more of that would be custom logic on our end..

Okay, that sounds fine.

PR #30, opened 2022-01-17
ENH: Add dtypes to read_info

### LINKED ISSUE #23: ENH: add data type information to read_info

brendan-ward:
Resolves #23

This adds an array of numpy dtype names to the output of `read_info`.  This intentionally does not refactor the `fields` entry, because that is also used as input when writing files and easiest to use as a simple list-like.

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -75,6 +75,7 @@ first layer.
   'crs': 'EPSG:4326',
   'encoding': 'UTF-8',
   'fields': array(['featurecla', 'scalerank', 'LABELRANK', ...], dtype=object),
+  'dtypes': array(['int64', 'object', 'object', 'object', 'float64'], dtype=object),
   'geometry_type': 'Polygon',
   'features': 255
 }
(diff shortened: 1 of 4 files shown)


brendan-ward:
This is intended to support chunking in [dask-geopandas #123](https://github.com/geopandas/dask-geopandas/pull/123)

martinfleis:
Thanks! This is great for dask-geopandas! 

I think that the only missing bit is an update of the `read_info` docstring.

martinfleis (approved):
Thanks! Do you envisage a potential release soon so we can use it is dask?

brendan-ward:
Do you mean a release via conda, or do you just need a release tagged here or source-only distribution pushed to PyPI?

If possible, I'd suggest testing dask-geopandas #123 against these latest changes, and a bit more investigation into why multiple processes in dask do not yield a speedup (in case it reveals anything else that needs to be added here first) before we cut another release.

martinfleis:
That sounds reasonable. 

I meant mostly to ensure that when we cut dask-geopandas 0.1, which we wanted to do once the open PRs there will be merged, we can rely on releases version of pyogrio. On conda-forge (and PyPI).

jorisvandenbossche (approved):
Thanks!

PR #38, opened 2022-01-28
ENH: support empty dataframes

### LINKED ISSUE #27: Handle empty GeoDataFrames
This is bit obscure issue but it seems that there is a demand in GeoPandas to allow writing empty GeoDataFrames to files (as GeoJSON). See https://github.com/geopandas/geopandas/issues/2234

Fiona seems to allow that for most drivers. It might be worth trying to achieve the same here. At the moment, we raise an error as we don't expect such a situation to happen.

```py
import pyogrio
import geopandas as gpd

df = gpd.GeoDataFrame(geometry=[])

driver = "GeoJSON"
pyogrio.write_dataframe(df, f"file_{driver}", driver=driver)

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/var/folders/2f/fhks6w_d0k556plcv3rfmshw0000gn/T/ipykernel_11567/3607813978.py in <module>
      1 driver = "GeoJSON"
----> 2 pyogrio.write_dataframe(df, f"file_{driver}", driver=driver)

~/Git/pyogrio/pyogrio/geopandas.py in write_dataframe(df, path, layer, driver, encoding, **kwargs)
    170 
    171     # TODO: validate geometry types, not all combinations are valid
--> 172     geometry_type = geometry.type.unique()[0]
    173 
    174     crs = None

IndexError: index 0 is out of bounds for axis 0 with size 0
```

martinfleis:
Closes #27

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -168,8 +168,11 @@ def write_dataframe(df, path, layer=None, driver=None, encoding=None, **kwargs):
     # TODO: may need to fill in pd.NA, etc
     field_data = [df[f].values for f in fields]
 
-    # TODO: validate geometry types, not all combinations are valid
-    geometry_type = geometry.type.unique()[0]
+    if not df.empty:
+        # TODO: validate geometry types, not all combinations are valid
+        geometry_type = geometry.type.unique()[0]
+    else:
+        geometry_type = "Unknown"
 
     crs = None
     if geometry.crs:
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -164,9 +164,7 @@ def test_read_fids_force_2d(test_fgdb_vsi):
         assert len(df) == 1
         assert df.iloc[0].geometry.has_z
 
-        df = read_dataframe(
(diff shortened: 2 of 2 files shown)


pyogrio/geopandas.py
```diff
@@ -168,8 +168,11 @@ def write_dataframe(df, path, layer=None, driver=None, encoding=None, **kwargs):
     # TODO: may need to fill in pd.NA, etc
     field_data = [df[f].values for f in fields]
 
-    # TODO: validate geometry types, not all combinations are valid
-    geometry_type = geometry.type.unique()[0]
+    if not df.empty:
+        # TODO: validate geometry types, not all combinations are valid
+        geometry_type = geometry.type.unique()[0]
+    else:
+        geometry_type = "Unknown"
```
martinfleis, line 175: `"Unknown"` is what OGR returns in `read_info` if you try to read an empty file saved by fiona and ensures that the behaviour of both is the same. For some reason, shapefile defaults to LineString (in both fiona and pyogrio) but the rest is read back correctly as "Unknown" geometry type.

brendan-ward (approved):
Thanks @martinfleis - you beat me to it!

PR #39, opened 2022-01-30
ENH: Better wrap GDAL errors and pass through GDAL exceptions 

### LINKED ISSUE #24: ENH: wrap GDAL errors
As mentioned in #19, we should wrap certain read operations against a dataset with `exc_wrap_pointer` so that we can provide the underlying GDAL error back to the user.

brendan-ward:
Closes #24

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,20 @@
 # CHANGELOG
 
+## 0.4.0 (in progress)
+
+### Major enhancements
+
+-   generalize check for VSI files from `/vsizip` to `/vsi` (#29)
+-   add dtype for each field to `read_info` (#30)
+-   support writing empty GeoDataFrames (#38)
+-   attempting to read features with an out of bound FID will now raise
+    `DriverIOError` instead of `ValueError`
+
+### Bug fixes
+
+-   use dtype `object` instead of `numpy.object` to eliminate deprecation warnings (#34)
+-   raise error if layer cannot be opened (#35)
+
 ## 0.3.0
 
 ### Major enhancements
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,11 +7,12 @@
 -   generalize check for VSI files from `/vsizip` to `/vsi` (#29)
 -   add dtype for each field to `read_info` (#30)
 -   support writing empty GeoDataFrames (#38)
+
+### Potentially breaking changes
+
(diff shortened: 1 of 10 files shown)


jorisvandenbossche (commented):
What's the general difference between DriverError and DriverIOError? (or what's the reason to have  both and differentiate?)

brendan-ward:
These came from [Fiona  error classes](https://github.com/Toblerity/Fiona/blob/maint-1.9/fiona/errors.py)

```
class DriverError(FionaValueError):
    """Encapsulates unsupported driver and driver mode errors."""

class DriverIOError(IOError):
    """A format specific driver error."""
```

(which also doesn't shed a lot of clarity)

I was trying to apply them here to indicate errors related to opening data sources as `DriverError` and errors related to reading from / writing to opened data sources as `DriverIOError`.  But yeah, this isn't obvious enough, and if it isn't obvious, then these don't make for good error handlers on the caller side either.

I'm not particularly a fan of using error classes in this way; I think we want to simplify and clarify errors that are related to the underlying data sources not directly under the user's control (e.g., failure to open a seemingly valid data sources) from errors directly related to a user-provided value (e.g., you tried to pass a value that is invalid, so `ValueError` / `TypeError` are likely more appropriate).  Will revisit our error classes here before finishing up this PR.

brendan-ward:
See `docs/src/errors.md` for a brief description of each error class I used here.  I tried to make them more clear, but also better distinguish those errors that come from the GDAL/OGR side and those that we can directly relate back to user inputs (`ValueError` / `TypeError`, etc).

Several of these are not yet easy to test; they require various types of invalid input files that we'll need to collect over time.  But this should follow the general pattern of wrapping GDAL to check for errors, if GDAL raises an error message, pass that back, otherwise try to create our own error message.

I used the pattern of `raise OurError(...) from None` here to help prevent issues around `During handling of the above exception, another exception occurred:`.  Not sure I got these all correct but we can always tune later once we have better test cases to trigger these.

martinfleis (approved):
Looks good to me! Thanks!

jorisvandenbossche (approved):
That looks like a thorough error handling now, looks good!

The one thing I was wondering is whether RuntimeError is the good error to subclass, but looking at the overview of the builtin exception classes, I also don't see another that would be better suited (except just base Exception).

PR #41, opened 2022-02-04
ENH: Use FID to set GeoDataFrame index

### LINKED ISSUE #14: Read FIDs into index of data frame
Return ndarray of FIDs from `read()` in addition to `fields` and `geometry`.  Use this for the index in `read_dataframe`.

brendan-ward:
Resolves #14

This expands the Cython layer to read in FID while reading features, unless `fids` are provided as input, in which case those are copied and returned.

Questions:
* do we want to make this the default behavior or opt-in?
* if opt-in, do we want this to be based solely on a keyword parameter to enable it, or do we also want to opt-in when `fids` are passed in?

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -16,6 +16,7 @@ def read_dataframe(
     where=None,
     bbox=None,
     fids=None,
+    set_fid_index=False,
 ):
     """Read from an OGR data source to a GeoPandas GeoDataFrame or Pandas DataFrame.
     If the data source does not have a geometry column or ``read_geometry`` is False,
@@ -67,6 +68,9 @@ def read_dataframe(
         specific (e.g. typically 0 for Shapefile and 1 for GeoPackage, but can
         still depend on the specific file). The performance of reading a large
         number of features usings FIDs is also driver specific.
+    set_fid_index : bool, optional (default: False)
+        If True, will return use the FIDs of the features that were read as the
+        index of the GeoDataFrame.  May start at 0 or 1 depending on the driver.
 
     Returns
     -------
@@ -99,11 +103,17 @@ def read_dataframe(
         where=where,
         bbox=bbox,
         fids=fids,
+        return_fids=set_fid_index,
     )
 
     columns = meta["fields"].tolist()
     data = {columns[i]: field_data[i] for i in range(len(columns))}
-    df = pd.DataFrame(data, columns=columns, index=pd.Series(index, name='fid'))
+    if set_fid_index:
+        index = index = pd.Series(index, name="fid")
+    else:
+        index = None
+
+    df = pd.DataFrame(data, columns=columns, index=index)
 
     if geometry is None or not read_geometry:
         return df
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -87,7 +87,7 @@ def read_dataframe(
         if not "/vsi" in path.lower() and not os.path.exists(path):
             raise ValueError(f"'{path}' does not exist")
 
-    meta, geometry, field_data = read(
+    meta, index, geometry, field_data = read(
         path,
         layer=layer,
         encoding=encoding,
@@ -103,7 +103,7 @@ def read_dataframe(
 
     columns = meta["fields"].tolist()
     data = {columns[i]: field_data[i] for i in range(len(columns))}
-    df = pd.DataFrame(data, columns=columns)
+    df = pd.DataFrame(data, columns=columns, index=pd.Series(index, name='fid'))
 
     if geometry is None or not read_geometry:
         return df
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
Code looks good!

My feeling is that this should be optional. I think depending on the file format, this can be just a counter and not necessarily informative?

Also, especially if we read this by default, we should also check how this integrates with the writing side. Do we set field ids from the index / a column named "fid"?

brendan-ward:
> Do we set field ids from the index / a column named "fid"?

I wasn't thinking so.  I was thinking that if you are creating a brand new file, FID should be sequential and based on the underlying driver (FIDs that start at 0 or 1), esp. since this would run into issues if you read from a format that starts FIDs at 0 then write to a format that starts FIDs at 1.

To me, reading FIDs was mostly to help correlate features that were read against the original source, so it only had meaning on the reading side.

It also seems mostly useful if you are reading a subset; otherwise in theory it is equivalent to a counter that starts at 0 or 1 (depending on format) - unless the producer of that file wrote fids differently.  But a subset may read FIDs 1,5,7, but the resulting sequential index of GeoDataFrame (without considering FID) would be 0,1,2 - which makes it a bit harder to correlate back to the source records if you needed to do so.

Will proceed with making this optional, toggled by a keyword.

brendan-ward:
Reading FIDs and setting those to GeoDataFrame index is now optional, toggled by `return_fids` in the `read()` function and `set_fid_index` in the `read_dataframe()` function.

pyogrio/geopandas.py
```diff
@@ -16,6 +16,7 @@ def read_dataframe(
     where=None,
     bbox=None,
     fids=None,
+    set_fid_index=False,
```
martinfleis, line 19: I find this keyword a bit cryptic at first sight. What about `fid_as_index` instead?

pyogrio/geopandas.py
```diff
@@ -67,6 +68,9 @@ def read_dataframe(
         specific (e.g. typically 0 for Shapefile and 1 for GeoPackage, but can
         still depend on the specific file). The performance of reading a large
         number of features usings FIDs is also driver specific.
+    set_fid_index : bool, optional (default: False)
+        If True, will return use the FIDs of the features that were read as the
```
martinfleis, line 72:
```suggestion
        If True, will use the FIDs of the features that were read as the
```
Not sure which of those verbs were intended to be there but likely not both.

pyogrio/geopandas.py
```diff
@@ -99,11 +103,17 @@ def read_dataframe(
...
         bbox=bbox,
         fids=fids,
+        return_fids=set_fid_index,
     )
 
     columns = meta["fields"].tolist()
     data = {columns[i]: field_data[i] for i in range(len(columns))}
-    df = pd.DataFrame(data, columns=columns)
+    if set_fid_index:
+        index = index = pd.Series(index, name="fid")
```
martinfleis, line 112: Can we create `pandas.Index` here directly instead of going via `Series`? Do we know the `dtype`? Can we pass it or we just let pandas figure it out?
```suggestion
        index = pd.Index(index, name="fid")
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -100,6 +100,21 @@ def test_read_null_values(test_fgdb_vsi):
     assert df.loc[df.SEGMENT_NAME.isnull()].SEGMENT_NAME.iloc[0] == None
 
 
+def test_read_set_fid_index(naturalearth_lowres):
+    kwargs = {"skip_features": 2, "max_features": 2}
+
+    # default is to not set FIDs as index
+    df = read_dataframe(naturalearth_lowres, **kwargs)
+    assert np.array_equal(df.index.values, pd.RangeIndex(0, 2).values)
```
martinfleis, line 108:
```suggestion
    assert_index_equal(df.index, pd.RangeIndex(0, 2)
```
I think it is better to use dedicated `assert_index_equal` here. It also needs to be imported above 

```py
from pandas.testing import assert_index_equal
```

Same below and we can even skip `assert df.index.name == "fid"` here because that gets tested implicitly.

martinfleis (commented):
Thanks! Couple of suggestions in the code.

pyogrio/geopandas.py
brendan-ward: Incoming values should be `int64` type, so I think we can use `index = pd.Int64Index(index, name="fid")` to be explicit about that.

jorisvandenbossche: `pd.Int64Index` will be deprecated in the future, so you can use just `pd.Index`. I suppose the data is already a numpy array of type int64? In that case, both are exactly equivalent and will preserve the dtype of the array.

martinfleis (approved):
Thanks!

jorisvandenbossche (approved):

Issue #40, opened 2022-02-01
Packaging: include original cython files (not cythonized .c files) in sdist


jorisvandenbossche:
Nowadays, I think it is best practice to not cythonize the pyx files and put the .c files in an sdist, but rather the .pyx files itself. It is then assumed that if you install such an sdist from source, that cython is a build dependency (as it is correctly listed in pyproject.toml).

jorisvandenbossche:
And something else packaging related: we should probably use `oldest-supported-numpy` in pyproject.toml

PR #45, opened 2022-02-10
GHA: add release action

martinfleis:
This workflow creates both github release and PyPI release everytime we create a new tag starting `v` (as `v0.4.0`. 

To make PyPI work, I need @brendan-ward or @jorisvandenbossche to add a repository secret called `PYPI_API_TOKEN` containing PyPI access token (or access to PyPI so I can do it myself).

Edit: token added

### DIFF
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v2
@@ -22,8 +24,8 @@ jobs:
       - name: Build a source tarball
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools
-          python setup.py sdist
+          pip install build setuptools
+          python -m build --sdist
 
       - name: Publish distribution to PyPI
         uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -1,0 +1,60 @@
+name: Publish pyogrio to PyPI / GitHub
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build-n-publish:
+    name: Build and publish pyogrio to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build a source tarball
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools
+          python setup.py sdist
+
+      - name: Publish distribution to PyPI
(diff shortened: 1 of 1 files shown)


.github/workflows/release_to_pypi.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build a source tarball
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools
```
martinfleis, line 25: @brendan-ward we probably need also Cython in here, don't we?

martinfleis:
I am trying to figure out the proper workflow in https://github.com/martinfleis/pyogrio/tree/gha_rls_check so do not merge this yet, it wouldn't work.

.github/workflows/release_to_pypi.yml
jorisvandenbossche: In principle, cython is not needed, because that is specified in the pyproject.toml

jorisvandenbossche:
@martinfleis we should probably first fix https://github.com/geopandas/pyogrio/issues/40, then you don't need cython and gdal to build an sdist

martinfleis:
Ah, okay. I'll put this on hold then, it will be much easier.

jorisvandenbossche:
I am taking a look at that now

.github/workflows/release_to_pypi.yml
jorisvandenbossche: We should also all the `build` package here, instead of directly invoking setup.py

martinfleis:
This should work assuming #46 is merged. I have [tested the build on another branch ](https://github.com/martinfleis/pyogrio/actions/runs/1829098548) and it seems to work as expected now.

brendan-ward (commented):
Thanks for working on this @martinfleis !

What do you think about changing this so that it gets triggered on `release` events instead of tags, and have it upload the sdist assets to the associated release?  That would mean remving the release creation action here, and modifying the upload of assets; not sure if those are easy changes.

I personally prefer creating the releases manually and having GH create a release tag as part of that process, so that I can provide hand-edited release notes (almost always copy-pasted from relevant section in `CHANGES.md`).

But since changes are well-documented in `CHANGES.md`, I'm not sure that is strictly necessary, I just thought it was helpful (I often refer to them where available for other projects).

martinfleis:
@brendan-ward In other geopandas repositories we let GHA create the release and then manually edit the release notes and copy&paste the relevant part of the changelog there. 

But it shouldn't be an issue changing this to wait for a manual release and then build sdist, upload it to that release and upload it to PyPI. It is up to you, I can make the changes.

brendan-ward (approved):
>  we let GHA create the release and then manually edit the release notes

Ah, didn't know that.  We should do the same here, to have a consistent process.

martinfleis:
More on that here https://github.com/geopandas/pyogrio/wiki/Release-checklist . I have copied the same from geopandas.

brendan-ward:
Thanks for the release checklist @martinfleis !

Feel free to merge this whenever you are ready; wasn't quite sure from your above comments when you wanted this merged.

PR #51, opened 2022-03-07
Add test for custom CRS read / write

brendan-ward:
This uses a custom (completely made up) CRS centered over Belgium to verify that a custom CRS can be round-tripped.

This should also ensure that GDAL data files / PROJ data files are used when reading / writing the file.

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -236,3 +236,26 @@ def test_write_empty_dataframe(tmpdir, driver, ext):
     df = read_dataframe(filename)
 
     assert_geodataframe_equal(df, expected)
+
+
+@pytest.mark.filterwarnings(
+    "ignore: You will likely lose important projection information"
+)
+def test_custom_crs_io(tmpdir, naturalearth_lowres):
+    df = read_dataframe(naturalearth_lowres)
+    # project Belgium to a custom Albers Equal Area projection
+    expected = df.loc[df.name == "Belgium"].to_crs(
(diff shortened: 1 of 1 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -236,3 +236,26 @@ def test_write_empty_dataframe(tmpdir, driver, ext):
...
+    write_dataframe(expected, filename)
+
+    assert os.path.exists(filename)
+
+    df = read_dataframe(filename)
+
+    crs = df.crs.to_dict()
+    assert crs["lat_1"] == 49.5
+    assert crs["lat_2"] == 51.5
+    assert crs["lon_0"] == 4.3
```
martinfleis, line 260:
```suggestion
    crs = df.crs.to_dict()
    assert crs["lat_1"] == 49.5
    assert crs["lat_2"] == 51.5
    assert crs["lon_0"] == 4.3
    assert df.crs.equals(expected.crs)
```

We can directly check if the expected and read CRS are equal.

martinfleis: I theory, we don't need that `dict`-based check at all, but happy to keep it as it is nicely explicit.

martinfleis (approved):
Thanks!

brendan-ward:
@martinfleis thanks for the review

@jorisvandenbossche this might be useful for testing the Windows wheels, though I think the conclusion there is that we need to package and then correctly detect GDAL Data / Proj Data similar to Fiona.

jorisvandenbossche:
Thanks, this certainly increases the coverage of our CRS handling code!

(I have to say it's not really clear to me when GDAL / PROJ would use the database and when not ..)

PR #52, opened 2022-03-09
Search for GDAL and PROJ data folders during startup

brendan-ward:
This follows the same approach as used by Fiona for searching for GDAL and PROJ data folders.

GDAL data folders are searched in the following precedence:
- wheel copy of gdal_data
- auto detection by GDAL (including `GDAL_DATA` env var)
- other well-known paths under `sys.prefix`

PROJ data folders are searched in the following precedence:
- wheel copy of proj
- default detection by GDAL (including `PROJ_LIB` env var)
- other well-known paths under `sys.prefix`

### DIFF
--- a/pyogrio/_ogr.pyx
+++ b/pyogrio/_ogr.pyx
@@ -185,7 +185,7 @@ def init_gdal_data():
             raise ValueError("Could not correctly detect GDAL data files installed by pyogrio wheel")
         return
 
-    # GDAL correctly found data files from compiled-in paths
+    # GDAL correctly found data files from GDAL_DATA or compiled-in paths
     if has_gdal_data():
         return
 
@@ -201,7 +201,6 @@ def init_gdal_data():
 
 def init_proj_data():
     """Set Proj search directories in the following precedence:
-    - PROJ_LIB env var
     - wheel copy of proj_data
     - default detection by GDAL, including PROJ_LIB (detected automatically by GDAL)
     - search other well-known paths under sys.prefix
@@ -218,7 +217,7 @@ def init_proj_data():
             raise ValueError("Could not correctly detect PROJ data files installed by pyogrio wheel")
         return
 
-    # GDAL correctly found PROJ based on compiled-in paths
+    # GDAL correctly found data files from PROJ_LIB or compiled-in paths
     if has_proj_data():
         return
 
--- a/pyogrio/_ogr.pyx
+++ b/pyogrio/_ogr.pyx
@@ -168,7 +168,6 @@ cdef char has_proj_data():
             OSRRelease(srs)
 
 
-
 def init_gdal_data():
     """Set GDAL data search directories in the following precedence:
     - GDAL_DATA env var
@@ -185,7 +184,7 @@ def init_gdal_data():
             raise ValueError("GDAL_DATA does not resolve to a directory that contains GDAL data files")
         return
 
-    # wheels are packaged to include PROJ data files at pyogrio/gdal_data
+    # wheels are packaged to include GDAL data files at pyogrio/gdal_data
     wheel_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "gdal_data"))
     if os.path.exists(wheel_path):
         set_gdal_config_options({"GDAL_DATA": wheel_path})
@@ -205,8 +204,6 @@ def init_gdal_data():
         return
 
     warnings.warn("Could not detect GDAL data files.  Set GDAL_DATA environment variable to the correct path.", RuntimeWarning)
-
-
 
 
 def init_proj_data():
--- a/pyogrio/_ogr.pyx
+++ b/pyogrio/_ogr.pyx
@@ -170,19 +170,12 @@ cdef char has_proj_data():
 
 def init_gdal_data():
     """Set GDAL data search directories in the following precedence:
-    - GDAL_DATA env var
     - wheel copy of gdal_data
-    - default install of GDAL data files
+    - default detection by GDAL, including GDAL_DATA (detected automatically by GDAL)
     - other well-known paths under sys.prefix
 
     Adapted from Fiona (env.py, _env.pyx).
     """
-
-    if "GDAL_DATA" in os.environ:
-        set_gdal_config_options({"GDAL_DATA": os.environ["GDAL_DATA"]})
-        if not has_gdal_data():
-            raise ValueError("GDAL_DATA does not resolve to a directory that contains GDAL data files")
-        return
 
     # wheels are packaged to include GDAL data files at pyogrio/gdal_data
     wheel_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "gdal_data"))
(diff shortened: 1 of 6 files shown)


brendan-ward:
@jorisvandenbossche this assumes that GDAL and Proj data files have been packaged into the wheels just like in Fiona, at
* `pyogrio/gdal_data`
* `pyogrio/proj_data`

Once those are in place, this _should_ fix the issues finding `proj.db` in #49

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,130 @@ def ogr_list_drivers():
...
+    - GDAL_DATA env var
+    - wheel copy of gdal_data
+    - default install of GDAL data files
+    - other well-known paths under sys.prefix
+
+    Adapted from Fiona (env.py, _env.pyx).
+    """
+
+    if "GDAL_DATA" in os.environ:
+        set_gdal_config_options({"GDAL_DATA": os.environ["GDAL_DATA"]})
```
jorisvandenbossche, line 183: Not that it gives that much of simplification code-wise, but I am wondering to what extent it is needed to set the option in this case, because GDAL itself already will check for the GDAL_DATA env variable.

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,130 @@ def ogr_list_drivers():
...
+        if not has_gdal_data():
+            raise ValueError("GDAL_DATA does not resolve to a directory that contains GDAL data files")
+        return
+
+    # wheels are packaged to include PROJ data files at pyogrio/gdal_data
```
jorisvandenbossche, line 188:
```suggestion
    # wheels are packaged to include GDAL data files at pyogrio/gdal_data
```

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,130 @@ def ogr_list_drivers():
...
+        set_gdal_config_options({"GDAL_DATA": wk_path})
+        if not has_gdal_data():
+            raise ValueError(f"Found GDAL data directory at {wk_path} but it does not appear to correctly contain GDAL data files")
+        return
+
+    warnings.warn("Could not detect GDAL data files.  Set GDAL_DATA environment variable to the correct path.", RuntimeWarning)
+
+
+
+
```
jorisvandenbossche, line 211: Tiny styling nit: 2 blank lines is enough

jorisvandenbossche (approved):
Thanks a lot for looking into this, looks great!

pyogrio/_ogr.pyx
jorisvandenbossche: (but it also doesn't hurt, so my comment doesn't matter that much)

brendan-ward: This was from Fiona, so I assume there was a good reason for it.

My guess is that it is related to precedence: we want to check this before falling back to checking the wheel location to give greatest control over how it gets set (otherwise, if we left it to GDAL to use env var, it would be after the check for the wheel location).

That said, I'm wondering now if the wheel location should take top precedence; if we deliberately put the data files there, we know they are correct for the version of GDAL pyogrio is compiled against, so it really doesn't make sense to let the user override that.  So then we could fall back to letting GDAL detect this env var.

jorisvandenbossche: > otherwise, if we left it to GDAL to use env var, it would be after the check for the wheel location).

Yes, we should still return early here if GDAL_DATA is properly set (so my suggestion was only about removing this single line, and not the whole block)

> That said, I'm wondering now if the wheel location should take top precedence

That's a good point, I also can't think of any way that you wouldn't want to use those files _if_ installed from a wheel.

One use case I am thinking about where this could actually matter is the (certainly not recommended! :)) situation of using a conda environment which has GDAL/PROJ installed (eg from installing fiona with conda), and then installing pyogrio from a wheel in that environment. In that case GDAL_DATA will be set and pointing to conda's GDAL, and thus it could give problems if first using the env variable (assuming it is still using the actual libraries included in the wheel).

brendan-ward:
Latest changes remove specific check for `GDAL_DATA`, 'PROJ_LIB` (just let GDAL detect these env vars instead of us); promotes wheel path to highest precedence.

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,113 @@ def ogr_list_drivers():
...
+    # wheels are packaged to include GDAL data files at pyogrio/gdal_data
+    wheel_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "gdal_data"))
+    if os.path.exists(wheel_path):
+        set_gdal_config_options({"GDAL_DATA": wheel_path})
+        if not has_gdal_data():
+            raise ValueError("Could not correctly detect GDAL data files installed by pyogrio wheel")
+        return
+
+    # GDAL correctly found data files from compiled-in paths
```
jorisvandenbossche, line 188:
```suggestion
    # GDAL correctly found data files from GDAL_DATA or compiled-in paths
```

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,113 @@ def ogr_list_drivers():
...
+        if not has_gdal_data():
+            raise ValueError(f"Found GDAL data directory at {wk_path} but it does not appear to correctly contain GDAL data files")
+        return
+
+    warnings.warn("Could not detect GDAL data files.  Set GDAL_DATA environment variable to the correct path.", RuntimeWarning)
+
+
+def init_proj_data():
+    """Set Proj search directories in the following precedence:
+    - PROJ_LIB env var
```
jorisvandenbossche, line 204:
```suggestion
```

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,113 @@ def ogr_list_drivers():
...
+    # wheels are packaged to include PROJ data files at pyogrio/proj_data
+    wheel_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "proj_data"))
+    if os.path.exists(wheel_path):
+        set_proj_search_path(wheel_path)
+        # verify that this now resolves
+        if not has_proj_data():
+            raise ValueError("Could not correctly detect PROJ data files installed by pyogrio wheel")
+        return
+
+    # GDAL correctly found PROJ based on compiled-in paths
```
jorisvandenbossche, line 221:
```suggestion
    # GDAL correctly found PROJ based on PROJ_LIB or compiled-in paths
```

pyogrio/_ogr.pyx
```diff
@@ -113,3 +121,112 @@ def ogr_list_drivers():
...
+            raise ValueError(f"Found GDAL data directory at {wk_path} but it does not appear to correctly contain GDAL data files")
+        return
+
+    warnings.warn("Could not detect GDAL data files.  Set GDAL_DATA environment variable to the correct path.", RuntimeWarning)
+
+
+def init_proj_data():
+    """Set Proj search directories in the following precedence:
+    - wheel copy of proj_data
+    - default detection by GDAL, including PROJ_LIB (detected automatically by GDAL)
```
jorisvandenbossche, line 205: Not too important since this is only a comment :), but I think it is actually PROJ itself that detects this (it seems GDAL just lets PROJ do this themselves, unless you set something with `OSRSetPROJSearchPaths`)

jorisvandenbossche:
Thanks Brendan! (and indeed just checking (after setting the wheel path) if GDAL already has the data is a nice simplification!)
I updated https://github.com/geopandas/pyogrio/pull/49 to make use of this.

PR #53, opened 2022-03-10
Add direct tests for GDAL and PROJ data discovery

brendan-ward:
Exposes check for valid GDAL / PROJ data file discovery to Python, and adds associated tests.

These should fail if GDAL and PROJ data are not correctly discovered by GDAL / PROJ.

In the case of PROJ, this verifies that PROJ data files are actually used, because it initializes a CRS object, which would fail if data files are not present.

In the case of GDAL data, this verifies that a [sentinal file is in the expected location](https://github.com/rasterio/rasterio/issues/1631#issuecomment-462395938).


jorisvandenbossche (approved):
Thanks!

PR #67, opened 2022-04-03
Fix: kwargs not passed on to OGR in write_dataframe

### LINKED ISSUE #62: ENH: support writing geopackage files without index
Writing a geopackage without index is 20% faster based on a test I did using gdal. So if you don't need an index in the output file, this is an interesting option.

Also, based on my experience, if you need to eg. merge several geopackages by appending them one by one to a new file, it is also faster to first add them all one by one without index, and then add the index in one go. Because pyogrio doesn't support appending yet this isn't a usable case yet, but... might become one in the future.

theroggy:
proposal to fix #62

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -216,4 +216,5 @@ def write_dataframe(df, path, layer=None, driver=None, encoding=None, **kwargs):
         crs=crs,
         geometry_type=geometry_type,
         encoding=encoding,
+        **kwargs
     )


jorisvandenbossche:
Thanks a lot for the catch!

Would you like to add a test for this? (use an easily verifyable option to ensure it was actually passed)

theroggy:
It was a bit of a search to find a gdal creation param that I could test without introducing new (explicit) library dependencies... but I think I found one :-).

martinfleis (approved):
Thanks!

brendan-ward (approved):
Thanks @theroggy - good catch!

PR #68, opened 2022-04-06
TST: Add test to verify GDAL detects zip archives

brendan-ward:
[GeoPandas crossref](https://github.com/geopandas/geopandas/pull/2225#issuecomment-1086589092)

GDAL already detects ZIP archives without needing `/vsizip/` or `zip://` prefix.  This adds a test to demonstrate that.

This does not add Fiona's full parsing of URLs, which includes splitting on `!` to separate out archive vs file within archive components; I don't have a good test string for that kind of path, and from [Fiona #742](https://github.com/Toblerity/Fiona/issues/742) I'm not sure if we should support it.

### DIFF
--- a/pyogrio/tests/test_path.py
+++ b/pyogrio/tests/test_path.py
@@ -120,6 +120,14 @@ def test_zip_path(naturalearth_lowres_vsi):
     assert len(df) == 177
 
 
+@pytest.mark.filterwarnings("ignore: Measured")
+def test_detect_zip_path(data_dir):
+    path = data_dir / "test_fgdb.gdb.zip"
+
+    df = pyogrio.read_dataframe(path, layer="test_lines")
+    assert len(df) == 2
+
+
 @pytest.mark.network
 def test_url():
     df = pyogrio.read_dataframe(
@@ -143,5 +151,5 @@ def aws_env_setup(monkeypatch):
 
 @pytest.mark.network
 def test_uri_s3(aws_env_setup):
-    df = pyogrio.read_dataframe('zip+s3://fiona-testing/coutwildrnp.zip')
+    df = pyogrio.read_dataframe("zip+s3://fiona-testing/coutwildrnp.zip")
     assert len(df) == 67


jorisvandenbossche:
Ah, I didn't know that GDAL does that (maybe only on recent versions?)

But, testing it locally, it also doesn't seem to work in general, so it would be good to understand the exact limitations. 

For example, reading a zipped shapefile does not seem to work naively:

```python
>>> df = pyogrio.read_dataframe(geopandas.datasets.get_path("naturalearth_lowres"))
>>> pyogrio.write_dataframe(df, "test_zipped_shapefile.shp", driver="ESRI Shapefile")
# zip the written files manually, try to read back
>>> pyogrio.read_dataframe("test_zipped_shapefile.zip")
...
DataSourceError: 'test_zipped_shapefile.zip' not recognized as a supported file format.
# rename extension of .zip to .shp.zip -> now it works ...
>>> pyogrio.read_dataframe("test_zipped_shapefile.shp.zip")
       pop_est      continent                      name iso_a3  gdp_md_est                                           geometry
0       920938        Oceania                      Fiji    FJI      8374.0  MULTIPOLYGON (((180.00000 -16.06713, 180.00000...
1     53950935         Africa                  Tanzania    TZA    150600.0  POLYGON ((33.90371 -0.95000, 34.07262 -1.05982...
...

```

So it seems that the file name (the part before the `.zip` extension) still matters. 
Of course, if you know that, you can ensure this is done properly so it will read automatically. But at least for zipped shapefiles, this is not the name you get by default when zipping the different files together using your OS file manager (at least on ubuntu, it will by default give a name with just ".zip").

brendan-ward:
> it would be good to understand the exact limitations

I'm not finding a lot of documentation on this, so it looks like I was hasty in assuming it worked.  

I reworked the implementation so that we always attempt to parse paths and thus can auto-detect zip files in either the archive component (preceding `!`) or in full path.  I tried to add enough test cases to make sure that this works as expected, but it is possible there are still edge cases we aren't handling correctly.

pyogrio/tests/test_path.py
```diff
@@ -120,6 +121,66 @@ def test_zip_path(naturalearth_lowres_vsi):
...
+    df = pyogrio.read_dataframe(f"zip://{path}!/a/b/test2.shp")
+    assert df.iso_a3[0] == "PER"
+
+    # specifying /vsizip/ should also work but path must already be in GDAL ready
+    # format without the "!"" archive specifier
+    df = pyogrio.read_dataframe(f"/vsizip/{path}/a/b/test2.shp")
+    assert df.iso_a3[0] == "PER"
+
+
+def test_prefix_vsizip():
```
jorisvandenbossche, line 160: Maybe move this next to `test_vsi_path` above, since that also tests directly `vsi_path` (without going through an actual read)

docs/source/introduction.md
```diff
@@ -279,6 +279,24 @@ using `zip://..`:
 >>> read_dataframe("zip://ne_10m_admin_0_countries.zip")
 ```
 
+If your archive contains multiple datasets, you need to specify which one to use;
+otherwise GDAL will default to the first one found.
+
+```python
+>>> read_dataframe("/vsizip/ne_10m_admin_0_countries.zip/a/b/test.shp")
```
jorisvandenbossche, line 286: Maybe use a dummy name instead of "ne_10m_admin_0_countries.zip", because this name denotes a specific naturalearth dataset, for which it would be a bit strange to have multiple shapefiles

jorisvandenbossche (commented):
This is looking great!

brendan-ward:
@jorisvandenbossche any ideas on the Windows test failures?  I don't have a Windows machine to test the paths on, so I can't tell if it is because when we use the `zip://` prefix that we have to reformat the path from `\` to `/` for GDAL, or if it is some other issue.

jorisvandenbossche:
Hmm, in geopandas.datasets.get_path, we also naively prepend `"zip://"`, and AFAIK we never got issues about that not working on windows. 

If needed, I can fire up my old windows laptop to diagnose.

pyogrio/util.py
```diff
@@ -11,21 +11,22 @@ def vsi_path(path: str) -> str:
     # Windows drive letters (e.g. "C:\") confuse `urlparse` as they look like
     # URL schemes
     if sys.platform == "win32" and re.match("^[a-zA-Z]\\:", path):
+        # try to detect archive component without url parsing
+        if path.split("!")[0].endswith(".zip"):
+            return f"zip://{path}"
```
jorisvandenbossche, line 16: Should we actually be prepending `/vsizip/` instead of `zip://`, since this is directly returned and doesn't go through `_construct_vsi_path`?

jorisvandenbossche: I suppose so, because `zip://` prefixed paths (which should work I think) would result in that: 

```
n [27]: pyogrio.util.vsi_path("zip://C:\layer.shp")
Out[27]: '/vsizip/C:\\layer.shp'
```

brendan-ward:
Turned out the simplest solution was simply to prefix with `zip://` then allow it to proceed with the rest of parsing, which sidesteps (I think) the issue with the windows drive letter.

pyogrio/tests/test_path.py
```diff
@@ -49,6 +50,20 @@ def change_cwd(path):
...
+        ("/a/b/test.zip", "/vsizip//a/b/test.zip"),
+        ("a/b/test.zip", "/vsizip/a/b/test.zip"),
+        # archives using ! notation should be prefixed by vsizip
+        ("test.zip!item.shp", "/vsizip/test.zip/item.shp"),
+        ("test.zip!/a/b/item.shp", "/vsizip/test.zip/a/b/item.shp"),
+        ("test.zip!a/b/item.shp", "/vsizip/test.zip/a/b/item.shp"),
+        ("/vsizip/test.zip/a/b/item.shp", "/vsizip/test.zip/a/b/item.shp"),
+        ("zip:///test.zip/a/b/item.shp", "/vsizip//test.zip/a/b/item.shp"),
+        (
+            "zip+https://s3.amazonaws.com/fiona-testing/coutwildrnp.zip",
```
jorisvandenbossche, line 64:
```suggestion
            "https://s3.amazonaws.com/fiona-testing/coutwildrnp.zip",
```

Does this also work? (and the version of zip+https is already included just above the new code)

brendan-ward: It uncovered an error!  Great!  We needed to put the `zip` scheme first rather than just appending it, otherwise the `https://` scheme got clobbered.  Fixed and added more tests.

pyogrio/util.py
```diff
@@ -82,29 +84,35 @@ def _parse_uri(path: str):
...
+            "vsi{0}".format(SCHEMES[p]) for p in schemes if p and p != "file"
+        )
+
+        if schemes[-1] in CURLSCHEMES:
+            suffix = f"{schemes[-1]}://"
 
-    prefix = "/".join(
-        "vsi{0}".format(SCHEMES[p]) for p in scheme.split("+") if p != "file"
-    )
+    print(f"schemes: {schemes}, prefix: {prefix}, suffix: {suffix}, archive: {archive}")
```
jorisvandenbossche, line 110: Small debugging leftover

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks a lot!

PR #69, opened 2022-04-08
CI: Try to fix MacOS vcpkg libgeotiff build

brendan-ward:


brendan-ward:
Something about the update of libgeotiff from `1.7.0#1` to `1.7.1` is breaking on MacOS on CI, even though it builds fine for me on local MacOS 10.15:

Contents of `libgeotiff/config-x64-osx-dynamic-rel-err.log`:

```
CMake Error at /usr/local/share/vcpkg/installed/x64-osx-dynamic/share/unofficial-sqlite3/unofficial-sqlite3-targets.cmake:55 (set_target_properties):
  The link interface of target "unofficial::sqlite3::sqlite3" contains:

    Threads::Threads

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/local/share/vcpkg/installed/x64-osx-dynamic/share/unofficial-sqlite3/unofficial-sqlite3-config.cmake:7 (include)
  /usr/local/share/vcpkg/scripts/buildsystems/vcpkg.cmake:816 (_find_package)
  /usr/local/Cellar/cmake/3.23.0/share/cmake/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /usr/local/share/vcpkg/installed/x64-osx-dynamic/share/proj/proj-config.cmake:13 (find_dependency)
  /usr/local/share/vcpkg/scripts/buildsystems/vcpkg.cmake:816 (_find_package)
  CMakeLists.txt:118 (FIND_PACKAGE)

```

jorisvandenbossche:
I can't directly help with the build failure on Mac (I would maybe report the build failure upstream at https://github.com/microsoft/vcpkg, since it worked before; it's a bit a pity that the dynamic osx build it not one of the built-in triplets, which would make it less likely to get broken).

But, maybe we should try to pin the build to an exact "baseline" for which we know it works (https://devblogs.microsoft.com/cppblog/take-control-of-your-vcpkg-dependencies-with-versioning-support/#baselines), and then only from time to time manually update the version.

brendan-ward:
This should now be using the vcpkg manifest approach for all builds, which includes a baseline commit in the vcpkg repository (set to just before libgeotiff 1.7.1 was merged), and a pinned version of libgeotiff that works.  We'll occasionally need to bump this version / baseline commit.

The instructions weren't particularly clear; given the basline that I set, it shouldn't have tried to install libgeotiff 1.7.1 at all, but it did so until I added an override to pin that to a prior version.

I couldn't reproduce the libgeotiff error locally, so hopefully that will be fixed at a later point.  I'm not seeing anything obvious to call attention to in the vcpkg setup for it, so I'm not sure what we would log out there as an issue other than it doesn't build in our specific environment.

brendan-ward:
Since this worked on last test of building the wheels, I'm going to go ahead and merge this so we can get closer to cutting an 0.4.0 release that builds wheels.  We can add future PRs to fix things here if needed.

jorisvandenbossche:
Sorry for the late review here, and thanks a lot for fixing this!

.github/workflows/release.yml
```diff
@@ -120,14 +118,17 @@ jobs:
       - name: Install GDAL
         env:
           VCPKG_DEFAULT_TRIPLET: ${{ matrix.triplet }}
+
         run: |
-          vcpkg install gdal[core] --overlay-triplets=./ci/custom-triplets
+          echo "VCPKG_INSTALLATION_ROOT: ${VCPKG_INSTALLATION_ROOT}"
```
jorisvandenbossche, line 123: Is this a debugging left-over? (the variable is documented by github actions about what it is)

ci/manylinux2014_x86_64-vcpkg-gdal.Dockerfile
```diff
@@ -13,6 +13,13 @@ RUN bootstrap-vcpkg.sh && \
...
+COPY ci/vcpkg.json opt/vcpkg/
 
 ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/opt/vcpkg/installed/x64-linux-dynamic/lib"
-RUN vcpkg install gdal[core]:x64-linux-dynamic --overlay-triplets=opt/vcpkg/custom-triplets --overlay-ports=opt/vcpkg/custom-ports
+RUN vcpkg install --overlay-triplets=opt/vcpkg/custom-triplets \
+    --triplet=x64-linux-dynamic \
+    --overlay-ports=opt/vcpkg/custom-ports \
+    --feature-flags="versions,manifests" \
+    --x-manifest-root=opt/vcpkg \
+    --x-install-root=opt/vcpkg/installed && \
```
jorisvandenbossche, line 24: Question: the reason that it is now needed to mention the install-root, is that because using manifest mode changes that? (it would now try to install inside the project, instead of in the vcpkg root?)

ci/vcpkg.json
```diff
@@ -0,0 +1,17 @@
...
+        {
+            "name": "gdal",
+            "default-features": false
+        }
+    ],
+    "builtin-baseline": "601d7f32059066ac229d4efbd9179790ccc57b0f",
+    "overrides": [
+        {
+            "name": "libgeotiff",
+            "version": "1.7.0"
```
jorisvandenbossche, line 14: I would have expected that only the baseline would be sufficient? (assuming the baseline is from a commit where libgeotiff is still at 1.7.0?)

jorisvandenbossche:
In any case it is a nice solution!

brendan-ward:
Until it broke on CI, at time of merge.  Because CI is... fun... (investigating now, further fixes will be a in a different PR)

jorisvandenbossche:
> Until it broke on CI, at time of merge. Because CI is... fun... (investigating now, further fixes will be a in a different PR)

But the Mac build passed :)

jorisvandenbossche:
The GDAL data files probably changes again location on windows ..

ci/manylinux2014_x86_64-vcpkg-gdal.Dockerfile
brendan-ward: Yes, using manifest mode changes this, and would have put it in `ci/vcpkg_installed/` which then would have required other updates here to use that path.

ci/vcpkg.json
brendan-ward: Provided I did it correctly (I picked commit immediately prior to 1.7.0 being merged), I would have expected the same.  However, I found it went ahead and tried to install 1.7.1 anyway, without setting this override.

I don't know that it is necessarily a problem for us to explicitly list it as a pinned version though - if only for documentation purposes, since this was the root of the problem.

jorisvandenbossche: > I don't know that it is necessarily a problem for us to explicitly list it as a pinned version though - if only for documentation purposes, since this was the root of the problem.

Indeed

sgillies: @jorisvandenbossche @brendan-ward I am finding vcpkg manifests difficult to use. Do you know if there is a vcpkg mode that will fail if no vcpkg.json is found? I suspect it's continuing without one in my case, but can't verify that.

jorisvandenbossche: @sgillies I think you are missing a "manifests" feature flag. Assuming this is about https://github.com/rasterio/rasterio-wheels/pull/90, will comment over there

PR #70, opened 2022-04-14
Add option to read_dataframe to use sql statement

theroggy:
I like to use SQL... but I don't know of an elegant/efficient way to get the result from an SQL statement on a geopackage to a GeoDataFrame at the moment. As a workaround, I execute the sql statement with gdal.VectorTranslate() (~python binding to ogr2ogr) to a temp file and then I read the result with geopandas.read_file(). It works, but it feels slightly suboptimal :-).

Because GDAL supports executing sql statements it seemed like an interesting and possibly realistic option to built in support for it in pyogrio.

To explore the idea I had a better look at the pyogrio code... and it seemed easier than I anticipated, so I immediately went for a POC. However, I don't have any experience in/knowledge about writing cython... so it's pure copy/paste from the code around it, so I hope it doesn't look too ugly :-(.
I added a TODO in ogr_read in a spot where I think it would be safer to use a try/finally to be sure everything gets cleaned up... but because it isn't there now I wonder if there is a good reason for doing it the way it is now?

If you can spare the time it would be great if you could have a look...

Remark: as it is a POC, I didn't make any changes to inline documentation yet.


pyogrio/_io.pyx
```diff
@@ -757,15 +796,24 @@ def ogr_read(
...
 
+    if sql is not None:
+        if where is not None or bbox is not None or fids is not None: # or skip_features or max_features:
+            raise ValueError(
+                "cannot set both 'sql' and any of 'where', 'bbox', 'fids'"
+            )
+
     ogr_dataset = ogr_open(path_c, 0, kwargs)
-    ogr_layer = get_ogr_layer(ogr_dataset, layer)
+    if sql is None:
```
brendan-ward, line 813: Defaulting the layer to 0 (lines 794-796) can move to this block since it is not used if `sql` is provided.

pyogrio/_io.pyx
```diff
@@ -757,15 +796,24 @@ def ogr_read(
...
+        if where is not None or bbox is not None or sql is not None or skip_features or max_features:
             raise ValueError(
-                "cannot set both 'fids' and any of 'where', 'bbox', "
+                "cannot set both 'fids' and any of 'where', 'bbox', 'sql', "
                 "'skip_features' or 'max_features'"
             )
         fids = np.asarray(fids, dtype=np.intc)
 
+    if sql is not None:
+        if where is not None or bbox is not None or fids is not None: # or skip_features or max_features:
```
brendan-ward, line 807:
```suggestion
        if where is not None or bbox is not None or fids is not None:
```

pyogrio/_io.pyx
```diff
@@ -834,6 +882,9 @@ def ogr_read(
     }
 
     if ogr_dataset != NULL:
+        # TODO: shouldn't this happen in an exception handler?
```
brendan-ward, line 885: No - my understanding is that `GDALDatasetReleaseResultSet` should always be called to release SQL layer.

The check `if ogr_dataset != NULL` can be removed; if we got to this point, it is always non-null.

pyogrio/_io.pyx
```diff
@@ -171,6 +171,43 @@ cdef OGRLayerH get_ogr_layer(GDALDatasetH ogr_dataset, layer) except NULL:
...
 
+cdef OGRLayerH execute_sql(GDALDatasetH ogr_dataset, sql, sql_dialect) except NULL:
+    """Execute an SQL statement on a dataset.
+
+    Parameters
+    ----------
+    ogr_dataset : pointer to open OGR dataset
+    sql : str
+        The sql statement to execute
+    sql_dialect : str
```
brendan-ward, line 182:
```suggestion
    sql_dialect : str, optional (default: None)
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+
+    # use spatialite function (needs SQLITE dialect)
+    sql = """SELECT ST_Buffer(geometry, 5) AS geometry, name, pop_est, iso_a3 
+               FROM naturalearth_lowres 
+              WHERE ISO_A3 = 'CAN'"""
+    df = read_dataframe(naturalearth_lowres, sql=sql, sql_dialect="SQLITE")
+    assert len(df) == 1
+    assert df.iloc[0].geometry.area > area_canada
+
+    # the geometry column doesn't need/cannot be specified when using the
```
brendan-ward, line 212:
```suggestion
    # the geometry column cannot be specified when using the 
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
         assert not df.iloc[0].geometry.has_z
 
 
+#@pytest.mark.filterwarnings("ignore: Layer")
+def test_read_sql(naturalearth_lowres):
+    # empty filter should return full set of records
```
brendan-ward, line 191: This can be removed since it is already covered by other tests, given default of `sql=None`.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+
+    # should return singular item
+    sql = "SELECT * FROM naturalearth_lowres WHERE iso_a3 = 'CAN'"
+    df = read_dataframe(naturalearth_lowres, sql=sql)
+    assert len(df) == 1
+    assert len(df.columns) == 6
+    assert df.iloc[0].iso_a3 == "CAN"
+    area_canada = df.iloc[0].geometry.area
+
+    # use spatialite function (needs SQLITE dialect)
```
brendan-ward, line 204: Suggest putting `sqlite` dialect test into a separate test case `test_read_sql_dialect`

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    sql = """SELECT ST_Buffer(geometry, 5) AS geometry, name, pop_est, iso_a3 
+               FROM naturalearth_lowres 
+              WHERE ISO_A3 = 'CAN'"""
+    df = read_dataframe(naturalearth_lowres, sql=sql, sql_dialect="SQLITE")
+    assert len(df) == 1
+    assert df.iloc[0].geometry.area > area_canada
```
brendan-ward, line 210: add 
```Python
assert len(df.columns) == 4
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+
+    # the geometry column doesn't need/cannot be specified when using the 
+    # default OGRSQL dialect but is returned nonetheless, so 4 columns
+    sql = """SELECT name, pop_est, iso_a3
```
brendan-ward, line 214: Suggest splitting this into a test that just reads subset of columns and verifies that those are present, and add another test of the `ORDER BY` clause.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+
+    # the geometry column doesn't need/cannot be specified when using the 
+    # default OGRSQL dialect but is returned nonetheless, so 4 columns
+    sql = """SELECT name, pop_est, iso_a3  
+               FROM naturalearth_lowres 
+              WHERE iso_a3 IN ('CAN', 'USA', 'MEX')
+              ORDER BY name"""
+    df = read_dataframe(naturalearth_lowres, sql=sql)
+    assert len(df.columns) == 4
+    assert len(set(df.iso_a3.unique()).difference(["CAN", "USA", "MEX"])) == 0
```
brendan-ward, line 220: Replace lines 220-222 with 

```Python
    assert len(df) == 3
    assert df.iso_a3.tolist() == ["CAN", "MEX", "USA"]
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    names_found = df['name'].to_list()
+    #assert len(names_found) == 4
+
+    # the geometry column doesn't need/cannot be specified when using the
```
brendan-ward, line 224: Only need this comment on first instance, don't need to repeat it

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+
+    # the geometry column doesn't need/cannot be specified when using the 
+    # default OGRSQL dialect but is returned nonetheless, so 4 columns
+    sql = """SELECT name, pop_est, iso_a3  
+               FROM naturalearth_lowres 
+              WHERE iso_a3 IN ('CAN', 'USA', 'MEX')
+              ORDER BY name"""
+    df = read_dataframe(naturalearth_lowres, sql=sql, skip_features=1, max_features=1)
+    assert len(df.columns) == 4
+    assert len(set(df.iso_a3.unique()).difference(["CAN", "USA", "MEX"])) == 0
```
brendan-ward, line 232: Replace lines 232-235 with 

```Python
    assert len(df) == 1
    assert df.iso_a3.tolist() == ["MEX"]
```

Suggest adding another couple tests that compounds the sql `LIMIT` with `skip_features` / `max_features`:

```Python
    sql = "SELECT name, pop_est, iso_a3 FROM naturalearth_lowres LIMIT 1"
    with pytest.raises(ValueError, match="'skip_features' must be between 0 and 0"):
        df = read_dataframe(naturalearth_lowres, sql=sql, skip_features=1)
```
(can't skip when there is only 1 result)

```Python
sql = "SELECT name, pop_est, iso_a3 FROM naturalearth_lowres LIMIT 1"
    df = read_dataframe(naturalearth_lowres, sql=sql, max_features=3)
    assert len(df) == 1
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,79 @@ def test_read_fids_force_2d(test_fgdb_vsi):
         assert not df.iloc[0].geometry.has_z
 
 
+#@pytest.mark.filterwarnings("ignore: Layer")
```
brendan-ward, line 189:
```suggestion
@pytest.mark.filterwarnings("ignore: Layer")
```

(needed for `LIMIT` / `skip_features` test below)

brendan-ward (changes requested):
Thanks for working on this @theroggy !  This is a good addition and may make it eaier for pyogrio to stand in as a faster substitute for `read_sql` in GeoPandas (though not a drop-in replacement).

Overall the Cython part looks pretty good, and most of my comments were focused on the tests.

It makes me think we should expand our test fixtures to include a tiny geopackage with a few layers, so that we can test interaction with `sql` and `where` since there are known issues with GPKG behaving differently on invalid SQL.  I added #73 for that, no need to address that in this PR.

Given that this is a good addition in its current form, please go ahead and add this to the documentation.  In addition to examples, which can largely be drawn from your test cases, we'll want to make sure to address incompatible parameter combinations, mention that you don't specify the geometry column when subsetting columns using the default dialect, and other SQL-specific gotchas (or a link to the GDAL pages for such).

pyogrio/_io.pyx
theroggy: I mean in the finally block of an exception handler, as there is a lot of code in between that can throw an exception...

pyogrio/tests/test_geopandas_io.py
theroggy: I put the sqlite test in a seperate function, but I already had a go at testing on different file formats and for that I had to specify the OGRSQL dialect on the other tests. If the dialect is not specified (dialect=None), it will depend on the database/file you connect to which dialect will be used:
- shapefile -> OGRSQL
- gpkg -> SQLite
(- postgress -> postgress)
(- Oracle -> oracle)
(- ...)

So I named the test function test_read_sql_dialect_sqlite.

pyogrio/_io.pyx
brendan-ward: Yes, that's correct - we probably should add one.  In which case, we should keep the `if ogr_dataset != NULL` within it.

pyogrio/tests/test_geopandas_io.py
brendan-ward: Ok - that will be a good thing to note in the documentation, since I would have naively expected it to default to `OGRSQL` unless directed otherwise.

pyogrio/geopandas.py
```diff
@@ -84,13 +86,39 @@ def read_dataframe(
         specific (e.g. typically 0 for Shapefile and 1 for GeoPackage, but can
         still depend on the specific file). The performance of reading a large
         number of features usings FIDs is also driver specific.
+    sql : str, optional (default: None)
+        The sql statement to execute. Look at the sql_dialect parameter for 
+        more information on the syntax to use for the query. When combined 
+        with other keywords like columns, skip_features, max_features, where 
+        or bbox, those are applied after the sql query. The ``layer`` keyword
```
brendan-ward, line 93: Also `fids` keyword cannot be combined; add a note here and add to list of of unsupported parameters under `fids` parameter.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,124 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    df = read_dataframe(
+        naturalearth_lowres, 
+        sql=sql, 
+        sql_dialect="OGRSQL", 
+        columns=["iso_a3_renamed", "name"],
+        where="iso_a3_renamed IN ('CAN', 'USA', 'MEX')",
+    )
+    assert len(df.columns) == 3
+    assert len(df) == 3
+    assert len(set(df.iso_a3_renamed.unique()).difference(["CAN", "MEX", "USA"])) == 0
```
brendan-ward, line 209:
```suggestion
    assert df.iso_a3_renamed.tolist == ["CAN", "MEX", "USA"]
```

Suggest similar changes below to sidestep set intersections.

(update order, if needed)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,124 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    # The geometry column cannot be specified when using the 
+    # default OGRSQL dialect but is returned nonetheless, so 4 columns.
+    sql = "SELECT iso_a3 AS iso_a3_renamed, name, pop_est FROM naturalearth_lowres"
+    df = read_dataframe(naturalearth_lowres, sql=sql, sql_dialect="OGRSQL")
+    assert len(df.columns) == 4
+    assert len(df) == 177
+
+    # Columns and/or where filters are applied after the sql query, so alias 
+    # needs to be specified instead of original column name.
+    df = read_dataframe(
```
brendan-ward, line 200: Suggest splitting out each combination of SQL and other parameters into their own individual test cases

pyogrio/tests/test_geopandas_io.py
```diff
@@ -186,6 +186,124 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    df = read_dataframe(naturalearth_lowres, sql=sql, sql_dialect="SQLITE")
+    assert len(df) == 1
+    assert len(df.columns) == 4
+    assert df.iloc[0].geometry.area > area_canada
+
+
+def test_read_sql_invalid(naturalearth_lowres):
+    with pytest.raises(Exception, match="SQL Expression Parsing Error"):
+        read_dataframe(naturalearth_lowres, sql="invalid")
+    with pytest.raises(ValueError, match="'sql' paramater cannot be combined with 'layer'"):
```
brendan-ward, line 303: Add space above

brendan-ward (changes requested):
Thanks for your continued work on this @theroggy , this is shaping up nicely!

Looks like you found it possible to combine `sql` with some of the other parameters (bbox, etc) that get applied after the SQL query.  Do you know if there are any performance implications lurking there that we should state in docstring / docs?  E.g., if you have a massive dataset that leverages a spatial index in its database, that using SQL followed by `bbox` may not be able to leverage that (not sure, just speculating).

docs/source/introduction.md
```diff
@@ -181,6 +182,41 @@ with the bbox.
...
+(the default) and the '[SQLITE](https://gdal.org/user/sql_sqlite_dialect.html#sql-sqlite-dialect)' 
+dialect. For SELECT statements the 'SQLITE' dialect tends to provide more 
+features as all [spatialite](https://www.gaia-gis.it/gaia-sins/spatialite-sql-latest.html) 
+functions can be used.
+
+This sql query returns the 5 Western European countries with the most 
+neighbours with their "circularity":
+
+```python
+>>> sql = """SELECT geometry, name, Circularity(geometry) AS circularity,
```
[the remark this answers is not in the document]
theroggy: It's indeed a bit of a balance between giving inspiration and complexity, but I agree the "Circularity" probably doesn't add a lot while increasing the complexity, so I'll remove it.

pyogrio/_io.pyx
```diff
@@ -752,90 +791,98 @@ def ogr_read(
...
+            'crs': crs,
+            'encoding': encoding,
+            'fields': fields[:,2], # return only names
+            'geometry_type': geometry_type
+        }
+    finally:
+        if ogr_dataset != NULL:
+            if sql is not None:
+                GDALDatasetReleaseResultSet(ogr_dataset, ogr_layer)
+            GDALClose(ogr_dataset)
```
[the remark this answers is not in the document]
theroggy: Do you mean a newline above `if ogr_dataset != NULL:`? or?

docs/source/introduction.md
```diff
@@ -181,6 +182,41 @@ with the bbox.
...
+              LIMIT 5"""
+>>> read_dataframe('ne_10m_admin_0_countries.shp', sql=sql, sql_dialect='SQLITE')
+          NAME  circularity  nb_neighbours                            geometry
+0       France     0.086359             11  MULTIPOLYGON (((-54.11153 2.114...
+1      Germany     0.123962             10  MULTIPOLYGON (((13.81572 48.766...
+2      Austria     0.216613              9  POLYGON ((16.94504 48.60417, 16...
+3  Switzerland     0.221956              6  POLYGON ((10.45381 46.86443, 10...
+4      Belgium     0.268855              5  POLYGON ((2.52180 51.08754, 2.5...
+```
+
```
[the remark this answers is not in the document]
theroggy: OK. I was wondering how elaborate the documentation in introduction.md should be compared with the information in the API reference, but I'll expand it a bit...

pyogrio/_io.pyx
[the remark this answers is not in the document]
brendan-ward: Above line 84 `GDALClose(ogr_dataset)`

pyogrio/_io.pyx
```diff
@@ -752,90 +791,98 @@ def ogr_read(
...
-    if ogr_dataset != NULL:
-        GDALClose(ogr_dataset)
-    ogr_dataset = NULL
+        meta = {
+            'crs': crs,
+            'encoding': encoding,
+            'fields': fields[:,2], # return only names
+            'geometry_type': geometry_type
+        }
+    finally:
```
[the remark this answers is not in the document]
brendan-ward: Yes

theroggy:
> Thanks for your continued work on this @theroggy , this is shaping up nicely!
> 
> Looks like you found it possible to combine `sql` with some of the other parameters (bbox, etc) that get applied after the SQL query. Do you know if there are any performance implications lurking there that we should state in docstring / docs? E.g., if you have a massive dataset that leverages a spatial index in its database, that using SQL followed by `bbox` may not be able to leverage that (not sure, just speculating).

I did some performance tests on how the combination of SQL + bbox behaves, and added the conclusions to the documentation. Also did some extra textual changes in general as well + added some more links to other relevant documentation.

theroggy:
Hmm... it's possible that spatialite is not available on the "Docker GDAL" test images?

brendan-ward:
> spatialite is not available on the "Docker GDAL" test images

Indeed.  We're using the [small image](https://github.com/OSGeo/gdal/tree/master/docker#full-osgeogdalubuntu-full-latest-aliased-to-osgeogdal), and it looks like spatialite is only enabled for the [full image](https://github.com/OSGeo/gdal/tree/master/docker#full-osgeogdalubuntu-full-latest-aliased-to-osgeogdal)

Pretty big difference in base sizes, though on CI this may only impact the transfer / startup time of the container.

A couple thoughts:
* we could switch to the full container [here](https://github.com/geopandas/pyogrio/blob/main/.github/workflows/docker-gdal.yml#L16-L20) but would want to see if that negatively impacts test execution time
* we could add the full container for latest released GDAL, and then toggle that test based on something (not sure: environment variable? feature detection? something else?)

theroggy:
I found a way to check if spatialite is available in sqlite... so the test is now skipped if it is not.

docs/source/introduction.md
```diff
@@ -181,6 +182,63 @@ with the bbox.
...
+doesn't natively support sql (e.g. 
+[ESRI Shapefile](https://gdal.org/drivers/vector/shapefile.html), 
+[FlatGeobuf](https://gdal.org/drivers/vector/flatgeobuf.html)), you can choose 
+between '[OGRSQL](https://gdal.org/user/ogr_sql_dialect.html#ogr-sql-dialect)' 
+(the default) and  
+'[SQLITE](https://gdal.org/user/sql_sqlite_dialect.html#sql-sqlite-dialect)'. 
+For SELECT statements the 'SQLITE' dialect tends to provide more spatial 
+features as all 
+[spatialite](https://www.gaia-gis.it/gaia-sins/spatialite-sql-latest.html) 
+functions can be used.
```
brendan-ward, line 204: Seems like given your experience with the CI not supporting this, it would be good to note that your GDAL needs to have spatialite support built in, as it is an optional component?

pyogrio/_io.pyx
```diff
@@ -738,6 +771,8 @@ def ogr_read(
     object where=None,
     tuple bbox=None,
     object fids=None,
+    object sql=None,
```
brendan-ward, line 774:
```suggestion
    str sql=None,
    str sql_dialect=None,
```

(I think this should work?)

pyogrio/geopandas.py
```diff
@@ -80,17 +80,46 @@ def read_dataframe(
...
         specific (e.g. typically 0 for Shapefile and 1 for GeoPackage, but can
         still depend on the specific file). The performance of reading a large
         number of features usings FIDs is also driver specific.
+    sql : str, optional (default: None)
+        The sql statement to execute. Look at the sql_dialect parameter for
+        more information on the syntax to use for the query. When combined
+        with other keywords like ``columns``, ``skip_features``,
+        ``max_features``, ``where`` or ``bbox``, those are applied after the
+        sql query. Be aware that this can have an impact on performance,
+        because e.g. filtering with the ``bbox`` keyword won't trigger use of
```
brendan-ward, line 93:
```suggestion
        (e.g. filtering with the ``bbox`` keyword may not use
```

pyogrio/geopandas.py
```diff
@@ -80,17 +80,46 @@ def read_dataframe(
...
+        spatial indexes anymore.
```
brendan-ward, line 94:
```suggestion
        spatial indexes).
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -207,6 +218,144 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+def test_read_sql_dialect_sqlite(naturalearth_lowres_all_ext):
+    # Should return singular item
+    sql = "SELECT * FROM naturalearth_lowres WHERE iso_a3 = 'CAN'"
+    df = read_dataframe(naturalearth_lowres_all_ext, sql=sql, sql_dialect="SQLITE")
+    assert len(df) == 1
+    assert len(df.columns) == 6
+    assert df.iloc[0].iso_a3 == "CAN"
+    area_canada = df.iloc[0].geometry.area
+
+    # Use spatialite function
```
brendan-ward, line 345: Suggest splitting this into a separate test that can be executed or skipped in whole, and then limit that test to GPKG inputs.

brendan-ward (commented):
This is looking great @theroggy , thanks for the updates.

Please also add to the changelog.

A few minor comments, and suggest splitting a test case for better skip-ability.

docs/source/introduction.md
theroggy: This is also specified in the GDAL documentation, but I'll mention it here as well.

pyogrio/tests/test_geopandas_io.py
theroggy: I now remembered a way to get this working anyway. GDAL is actually always build with spatialite support as far as I know, but for some odd reason spatialite is not activated in sqlite on those images. So as long as you don't run such queries on non-native-sqlite files, the spatialite used will be the one "inside" GDAL. If you run such a query on a GPKG, by default GDAL will just pass the sql through to sqlite... and if spatialite isn't activated there, it doesn't work.

However, you can force gdal not to pass through the SQLITE query by specifying "INDIRECT_SQLITE" as sql_dialect, and then it works, even though performance might be impacted.

I'll change the documentation accordingly, but regarding the tests: do you still like the tests separated in this case? When it is detected that spatialite is not available within sqlite, the won't be skipped anymore but the sqlite_dialect will be changed to INDIRECT_SQLITE.

brendan-ward: > do you still like the tests separated in this case

Yes, I think smaller focused test cases are better.

> When it is detected that spatialite is not available within sqlite, the won't be skipped anymore but the sqlite_dialect will be changed to INDIRECT_SQLITE

This is a good solution, thank you for the additional investigation.

> request for the gdal-ubuntu-small image that spatialite would be activated within sqlite as well

Certainly could be raised as an issue in the GDAL repo; I'm not sure of the implications for image size / complexity / license (could well be the latter).

theroggy: OK, I kept it in a separate test...

CHANGES.md
```diff
@@ -11,6 +11,7 @@
 -   generalize check for VSI files from `/vsizip` to `/vsi` (#29)
 -   add dtype for each field to `read_info` (#30)
 -   support writing empty GeoDataFrames (#38)
+-   support to use a sql statement in read_dataframe (#70)
```
brendan-ward, line 14:
```suggestion
-   support use of a sql statement in read_dataframe (#70)
```

docs/source/introduction.md
```diff
@@ -42,8 +42,9 @@ You can certainly try to read or write using unsupported drivers that are
 available in your installation, but you may encounter errors.
 
 Note: different drivers have different tolerance for mixed geometry types, e.g.,
-MultiPolygon and Polygon in the same dataset. You will get exceptions if you
-attempt to write mixed geometries to a driver that does not support them.
+MultiPolygon and Polygon in the same dataset. You will get exceptions or
```
brendan-ward, line 45: Given all the work on this topic in other PRs at the moment, I'd suggest deleting this.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -207,6 +219,165 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+    assert df.iloc[0].geometry.area > area_canada
+
+
+@pytest.mark.parametrize(
+        "naturalearth_lowres", [".gpkg"],
+        indirect=["naturalearth_lowres"])
+def test_read_sql_dialect_sqlite_gpkg(naturalearth_lowres):
+    # Should return singular item
+    sql = "SELECT * FROM naturalearth_lowres WHERE iso_a3 = 'CAN'"
+    df = read_dataframe(naturalearth_lowres, sql=sql, sql_dialect="INDIRECT_SQLITE")
```
brendan-ward, line 365: Suggest adding a comment indicating that `INDIRECT_SQLITE` uses spatialite functionality within GDAL rather than depending on libspatialite.

brendan-ward (approved):
Thanks for all the updates here @theroggy 

A couple very minor comments, then this is good to go.

brendan-ward (approved):
Thanks @theroggy 🚀

theroggy:
For the issue about spatialite not available in image osgeo/gdal:ubuntu-small-latest, the following issue was opened:
OSGeo/gdal#5657

Issue #76, opened 2022-04-20
Unable to install with pip (ValueError: 'pyogrio/_err.pyx' doesn't match any files)


kylebarron:
Seems like the cython compilation is trying to find a file that doesn't exist? This might be an issue with the `MANIFEST.in` not including the right files?

Here are the files included in the sdist from pip:
```
> extract pyogrio-0.3.0.tar.gz
x pyogrio-0.3.0/
x pyogrio-0.3.0/LICENSE
x pyogrio-0.3.0/MANIFEST.in
x pyogrio-0.3.0/PKG-INFO
x pyogrio-0.3.0/README.md
x pyogrio-0.3.0/pyogrio/
x pyogrio-0.3.0/pyogrio/__init__.py
x pyogrio-0.3.0/pyogrio/_env.py
x pyogrio-0.3.0/pyogrio/_err.c
x pyogrio-0.3.0/pyogrio/_geometry.c
x pyogrio-0.3.0/pyogrio/_io.c
x pyogrio-0.3.0/pyogrio/_ogr.c
x pyogrio-0.3.0/pyogrio/_version.py
x pyogrio-0.3.0/pyogrio/core.py
x pyogrio-0.3.0/pyogrio/errors.py
x pyogrio-0.3.0/pyogrio/geopandas.py
x pyogrio-0.3.0/pyogrio/raw.py
x pyogrio-0.3.0/pyogrio.egg-info/
x pyogrio-0.3.0/pyogrio.egg-info/PKG-INFO
x pyogrio-0.3.0/pyogrio.egg-info/SOURCES.txt
x pyogrio-0.3.0/pyogrio.egg-info/dependency_links.txt
x pyogrio-0.3.0/pyogrio.egg-info/requires.txt
x pyogrio-0.3.0/pyogrio.egg-info/top_level.txt
x pyogrio-0.3.0/pyproject.toml
x pyogrio-0.3.0/setup.cfg
x pyogrio-0.3.0/setup.py
x pyogrio-0.3.0/versioneer.py
```

```
> pip install pyogrio
Collecting pyogrio
  Using cached pyogrio-0.3.0.tar.gz (322 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [24 lines of output]
      Traceback (most recent call last):
[...]

PR #80, opened 2022-04-21
CI: Add specific test for sdist

brendan-ward:
Adds a specific test to the release builds to ensure that the source distribution can be successfully built (just using the GDAL container to keep this easy) and prevent future occurrences of broken sdist builds (xref #76).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,6 +57,13 @@ jobs:
         with:
           path: wheelhouse
 
+      - name: Log paths
+        run: |
+          echo "wheelhouse/*"
+          ls wheelhouse/*
+          echo "/wheelhouse/*"
+          ls /wheelhouse/*
+
       - name: Install python
         run: |
           apt-get update && apt-get install -y python3 python3-pip
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,129 +42,14 @@ jobs:
           path: ./dist/*.tar.gz
           retention-days: 5
 
-  build-wheels-linux:
-    name: Build wheels on Linux
-    runs-on: "ubuntu-20.04"
+  test-sdist:
+    name: Test sdist
+    needs: [build-sdist]
+    runs-on: ubuntu-latest
+    container:
+      image: "osgeo/gdal:ubuntu-small-3.4.0" # python 3.8.10
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-
-      - name: Set up Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v1
-        with:
-          install: true
-          buildkitd-flags: --debug
-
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,6 +42,38 @@ jobs:
...
+
+      - name: Install python
+        run: |
+          apt-get update && apt-get install -y python3 python3-pip
+
+      - name: Build from sdist
+        shell: bash
+        run: |
+          python3 -m pip install --no-cache-dir cython~=0.29
+          python3 -m pip install --no-cache-dir --no-build-isolation wheelhouse/artifact/*.tar.gz
```
jorisvandenbossche, line 66: This should also work without the `--no-build-isolation` (and installing cython upfront)? 
(which then also tests that the pyproject.toml build deps are correct)

.github/workflows/release.yml
```diff
@@ -42,6 +42,38 @@ jobs:
...
+          python3 -m pip list
+
+      - name: Run tests
+        shell: bash
+        run: |
+          cd ..
+          python3 -m pip install --no-cache-dir pytest pandas pyproj shapely
```
jorisvandenbossche, line 73: I wanted to say you can do `-r ci/requirements-wheel-test.txt`, but that's not possible in the docker I suppose?

.github/workflows/release.yml
```diff
@@ -42,6 +42,38 @@ jobs:
...
+          python3 -m pip install --no-cache-dir --no-deps geopandas
```
jorisvandenbossche, line 74: Maybe move those two install lines to the step above, so those are included in the `pip list` output?

jorisvandenbossche (commented):
Yes, this was one of the remaining things I still wanted to add to the release workflow. Thanks!

Maybe change the "needs" for the release job from "build-sdist" to "test-sdist"

.github/workflows/release.yml
```diff
@@ -42,6 +42,38 @@ jobs:
...
+
+    steps:
+      - name: Download sdist from artifacts
+        uses: actions/download-artifact@v2
+        with:
+          path: wheelhouse
+
+      - name: Install python
+        run: |
+          apt-get update && apt-get install -y python3 python3-pip
```
jorisvandenbossche, line 60: python is already installed on the docker image, so this step might not be needed? (I noticed because the build log said that numpy was already installed) Although pip might still be needed to install

.github/workflows/release.yml
brendan-ward: This folder is not present without first doing a code checkout, which we're deliberately omitting here; hence the need to hard-code them.

.github/workflows/release.yml
brendan-ward: Right!  `python3-pip` came from our existing docker GDAL test, but I went ahead and added `python3` to be safe without thinking it through...

jorisvandenbossche (approved):

PR #93, opened 2022-04-30
Handle nulls for additional integer types now present because of GDAL subtypes

brendan-ward:
#83 exposed a bug that wasn't covered in our test suite.

When reading a dataset, if a field claims to be an integer type but has a null value, we have to re-cast it to float64 in order to hold `np.nan`.

This expands the check to any integer type.  I didn't add bool to the list: should we?

I'm not 100% confident in the test; we're preparing the data as floating point in order to include null values, but then sending different types during write.  I'm not sure that this actually tricks the driver into actually storing those values as those types, but what seems to matter here is that the driver claims they are integer types on read.

Detected reading from a FGDB with null values in an int16 field.


jorisvandenbossche:
> I didn't add bool to the list: should we?

I would do that as well. It might not be common, but I suppose nothing in the GDAL data model prevents having nulls in a bool subtype column (I don't know to what extent file formats support this though). 

For preparing the data, I don't think the current test actually is testing it. Because we infer the OGR field type based on the numpy dtype of the data (`cdef infer_field_types(list dtypes)`), so that solely depend on the passed `field_data` list of ndarrays. And in the case of the test, that are all float64 arrays (the `fields` variable in the test is only the field names). 

So I don't think it's currently possible to write integer data with nulls (since it is not possible to have that with numpy / pandas). Short term we could maybe create a small file with fiona (or some other software like QGIS) and include that. 

Longer term, one option could be to let the user override the inferred field types (although not sure I am a fan of this). Another option could be to support "masked" arrays as input (so where you have a second boolean array that indicates whether there is a null or not), which is the approach that pandas' nullable dtypes use (which would also enable us to support those more properly).

jorisvandenbossche:
Small example:

```python
import geopandas
import fiona
import pyogrio

df = geopandas.GeoDataFrame(
    {"col_bool": [True, False, True]}, geometry=geopandas.points_from_xy([0, 1, 2], [0, 1, 2]),
    crs="EPSG:4326"
)
df.to_file("test_null.gpkg")

# append a record with a null using fiona
with fiona.open("test_null.gpkg") as c:
    rec = next(iter(c))

# >>> rec
# {'type': 'Feature',
#  'id': '1',
#  'properties': OrderedDict([('col_bool', True)]),
#  'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)}}

rec["properties"]["col_bool"] = None
with fiona.open("test_null.gpkg", "a") as c:
    c.write(rec)

# result with geopandas vs pyogrio

In [16]: geopandas.read_file("test_null.gpkg")
Out[16]: 
  col_bool                 geometry
0     True  POINT (0.00000 0.00000)
1    False  POINT (1.00000 1.00000)
2     True  POINT (2.00000 2.00000)
3     None  POINT (0.00000 0.00000)

In [18]: pyogrio.read_dataframe("test_null.gpkg")
Out[18]: 
   col_bool                 geometry
0      True  POINT (0.00000 0.00000)
1     False  POINT (1.00000 1.00000)
2      True  POINT (2.00000 2.00000)
3      True  POINT (0.00000 0.00000)
```

So that is with the main branch, reading the null incorrectly. With this branch (with "b" added to the `dtype.kind` check) I get:

```
In [2]: pyogrio.read_dataframe("test_null.gpkg")
Out[2]: 
   col_bool                 geometry
0       1.0  POINT (0.00000 0.00000)
1       0.0  POINT (1.00000 1.00000)
2       1.0  POINT (2.00000 2.00000)
3       NaN  POINT (0.00000 0.00000)
```

jorisvandenbossche:
> I'm assuming that in the case of bool that we should instead treat nulls as `None` instead of `np.nan`, and set the output dtype to `object` instead of `bool`?

I think both are valid choices (None with object dtype, or NaN with float dtype). Using object dtype preserves better the type information, using floats make it more efficient to handle. In both cases you will first need to do something with the missing value and cast to bool, before you can actually use it in a boolean context (like using it as mask in a filtering operation)

With fiona we currently get object dtype, because in geopandas we pass the actual values from fiona as a sequence to the pandas constructor, which defaults to object dtype in this case:

```
In [41]: pd.Series([True, False, None])
Out[41]: 
0     True
1    False
2     None
dtype: object
```

Using float is more in line with how integers are handled (and will be more performant). 

(I don't have a strong opinion on which option to choose)

pyogrio/tests/test_raw_io.py
```diff
@@ -390,16 +378,27 @@ def test_read_write_data_types_numeric(tmp_path, ext):
...
         assert np.array_equal(result, np.array([1, 2, 3]))
         assert result.dtype == result_dtype
+
+
+def test_read_data_types_numeric_with_null(test_gpkg_nulls):
+    fields = read(test_gpkg_nulls)[3]
+
+    # all fields should be cast to float64 and last value should be np.nan
+    for field in fields:
+        assert field.dtype == "float64"
```
jorisvandenbossche, line 403: float32 should be preserved as float32?

brendan-ward: I had mistakenly assumed that dtypes would be preserved from GeoPandas through Fiona to the GPKG, but this appears not to be the case (happens even if there are no nulls set when writing to Fiona); all integer fields are converted to `int64` regardless of nulls.  And had also assumed that we needed `float64` to hold `np.nan`, but that isn't true either.

The `float32` field is indeed preserved correctly according to `ogrinfo`, but apparently we're not detecting the float32 subtype correctly in this case (it appears not to be set, according to logging I've added).  Investigating now...

brendan-ward: Oh - nevermind, `ogrinfo` is showing both as `float64` (`Real`) as well; was misled by our column names with the dtypes again...

I think the upshot is that we need to do something else to make sure dtypes are correctly written.  Updated to write the non-null GPKG ourselves then use Fiona to append in nulls; appears to preserve GDAL equivalents of our field types (none of unsigned integer types supported) and also preserves float32.

jorisvandenbossche (approved):
Looks good, thanks!

PR #97, opened 2022-05-10
Use certifi to set GDAL_CURL_CA_BUNDLE / PROJ_CURL_CA_BUNDLE defaults

### LINKED ISSUE #66: Linux wheels require CURL_CA_BUNDLE to be set
> The curl issue sounds like maybe a certificate issue. Perhaps you need to run

Looking for the error in GDAL context, I found some issues from rasterio/fiona: see the explanation at https://github.com/sgillies/frs-wheel-builds/issues/12 / https://github.com/rasterio/rasterio/issues/1737, and this is actually documented for rasterio at https://github.com/rasterio/rasterio/blob/master/README.rst#ssl-certs. 
And indeed, if I do `CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt pytest --pyargs pyogrio` all tests pass.

It seems that fiona fixed this by setting that environment variable for you: https://github.com/Toblerity/Fiona/commit/5743a9600e80ab033ecb9b9275c0c3f954349e21 
I think we can probably do something similar here. Although we can maybe just do it for wheels, and not in general. And in case of manylinux wheels (although can we detect this?) we could maybe even harcode the value, instead of adding `certifi` as an additional requirement.

_Originally posted by @jorisvandenbossche in https://github.com/geopandas/pyogrio/issues/55#issuecomment-1068224847_

brendan-ward:
Resolves #66

This makes `certifi` and installation requirement and follows approach from [Fiona #1095](https://github.com/Toblerity/Fiona/pull/1095) and [rasterio #2459](https://github.com/rasterio/rasterio/pull/2459)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -216,12 +216,6 @@ jobs:
           python -m pip install --find-links wheelhouse/artifact pyogrio
           python -m pip list
 
-      # TODO temporary set env variable for curl certificate for Linux
-      - name: Set CURL_CA_BUNDLE environment variable
-        if: runner.os == 'Linux'
-        run: |
-          echo "CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt" >> $GITHUB_ENV
-
       - name: Run tests
         shell: bash
         run: |
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -2,7 +2,6 @@ name: test
 channels:
   - conda-forge
 dependencies:
-  - certifi
   - numpy
   - libgdal
   - pytest
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -2,6 +2,7 @@ name: test
 channels:
   - conda-forge
 dependencies:
+  - certifi
   - numpy
   - libgdal
   - pytest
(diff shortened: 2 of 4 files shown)


jorisvandenbossche:
The conda macos failure is also happening on main, so unrelated:

```
 ImportError while loading conftest '/Users/runner/work/pyogrio/pyogrio/pyogrio/tests/conftest.py'.
pyogrio/__init__.py:1: in <module>
    from pyogrio.core import (
pyogrio/core.py:6: in <module>
    from pyogrio._ogr import (
E   ImportError: dlopen(/Users/runner/work/pyogrio/pyogrio/pyogrio/_ogr.cpython-39-darwin.so, 2): Symbol not found: _ZSTD_compressBound
E     Referenced from: /Users/runner/micromamba/envs/test/lib/libgdal.30.dylib
E     Expected in: /Users/runner/micromamba/envs/test/lib/libblosc.1.dylib
E    in /Users/runner/micromamba/envs/test/lib/libgdal.30.dylib
```

jorisvandenbossche:
Ah, I see you already opened https://github.com/geopandas/pyogrio/pull/99 to try to fix this

jorisvandenbossche (approved):
Looks good!

PR #99, opened 2022-05-10
CI: Try setting blosc version manually for conda test environment

brendan-ward:
We're running into CI errors with GDAL loading Blosc dynamically; looks possibly related to recent updates with blosc package on conda-forge.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -7,3 +7,6 @@ dependencies:
   - pytest
   - pygeos
   - geopandas-base
+
+  # temporary: to get around issues with GDAL using libblosc 1.21.1
+  - blosc==1.21.0


jorisvandenbossche:
I suppose a rerendering of GDAL should also fix it (https://github.com/conda-forge/gdal-feedstock/pull/624)

brendan-ward:
Not sure?  We can also wait for that PR to get merged and try again.  Since this affects only our conda tests on MacOS, we can also just let this sit for a bit - but ideally would be good to have that sorted before we push 0.4 to conda-forge, mostly just to know that upstream issues won't prevent users from being able to use pyogrio via conda.

brendan-ward:
Superseded by #100

Issue #94, opened 2022-05-02
Writing an empty geodataframe to .geojsons results in 0 byte file


theroggy:
Probably an issue in gdal...

Can be reproduced in a test that will be added in #75: test_write_empty_dataframe_unsupported()

brendan-ward:
I am not able to reproduce (MacOS 10.15, GDAL 3.4.1); pyogrio produces an empty GeoJSONSeq file (0 bytes).  This is identical to what I get from GeoPandas (via Fiona) using fixes in [geopandas #2240](https://github.com/geopandas/geopandas/pull/2240) to allow writing empty data frames and to what I get using `ogr2ogr` to convert an empty GeoJSON to GeoJSONSeq.

What version of GDAL are you using?

jorisvandenbossche:
> pyogrio produces an empty GeoJSONSeq file (0 bytes)

This is also what the title of the issue says?

Although it is annoying that it cannot read such an empty file (that is maybe something to report to gdal?), I suppose this is the expected behaviour? Given that a geojsonseq file consists of one line per record, if there are no records there are no lines in the file?

brendan-ward:
Ah yes, didn't read that, jumped from #101 where I was next going to comment along the lines of why are we attempting to read it when we should just check existence of it...  

I don't know what I would ask GDAL to do differently: not create a file at all?  That also seems to break the model with something like `ogr2ogr` where you would always expect an output file.

jorisvandenbossche:
Yes, I don't think writing it is a problem, it's just that ogr2ogr also is not able to roundtrip, it can't read the file it has written. If you start from an empty geojson:

```
$ ogr2ogr empty.geojsons empty.geojson
$ ogr2ogr empty_roundtrip.geojson empty.geojsons
FAILURE:
Unable to open datasource `empty.geojsons' with the following drivers.
...
```

theroggy:
Yes, indeed the way it is written is probably OK, but it seems more logical at first sight that gdal would be able to read an empty .geojsons file as a datasource with 0 rows...

theroggy:
I update the title and description based on the new insights...

PR #75, opened 2022-04-19
Add promote_to_multitype keyword in write_dataframe

theroggy:
Closes #56 

I also gave a shot at an automatic mode: if None is passed (the default):
- if the geodataframe is mixed multi+single for the same primitive type + the output format doesn't support mixed types
  -> promote_to_multitype=True  
- otherwise it is False 

I also made some changes to the write_dataframe test so it also tests the automatic detection of the driver + all supported file types

PS: in an ideal world, I think ?
- the `sort_geodataframe` function would be available as a parameter in `assert_geodataframe_is_equal`
- the `to_multipolygon` function would be available in GeoPandas (or rather a version that supports all geometry types)

PPS: when testing on 550.000 polygons, promote_to_multitype=True doesn't give any performance hit when writing a GeoDataFrame

PR #101, opened 2022-05-11
Add auto detection of driver for .geojson, .geojsonl and .geojsons files

theroggy:
\+ add to file formats that should be tested.

Issue found: writing empty file to .geojsons is appartently not supported: logged here: #94

### DIFF
--- a/pyogrio/tests/conftest.py
+++ b/pyogrio/tests/conftest.py
@@ -8,7 +8,7 @@ import pyogrio
 
 
 _data_dir = Path(__file__).parent.resolve() / "fixtures"
-ALL_EXTS = [".shp", ".gpkg", ".geojson", ".geojsons"]
+ALL_EXTS = [".shp", ".gpkg", ".geojson", ".geojsonl", ".geojsons"]
 
 
 def pytest_report_header(config):
--- a/pyogrio/tests/conftest.py
+++ b/pyogrio/tests/conftest.py
@@ -8,7 +8,7 @@ import pyogrio
 
 
 _data_dir = Path(__file__).parent.resolve() / "fixtures"
-ALL_EXTS = [".shp", ".gpkg", ".geojson", ".geojsonl", ".geojsons"]
+ALL_EXTS = [".shp", ".gpkg", ".geojson", ".geojsonl"]
 
 
 def pytest_report_header(config):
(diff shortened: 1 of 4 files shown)


pyogrio/tests/conftest.py
```diff
@@ -8,7 +8,7 @@
 
 
 _data_dir = Path(__file__).parent.resolve() / "fixtures"
-ALL_EXTS = [".shp", ".gpkg", ".json"]
+ALL_EXTS = [".shp", ".gpkg", ".geojson", ".geojsonl", ".geojsons"]
```
jorisvandenbossche, line 11: Is it needed to have both extensions in here? Because that means we duplicate tests for GeoJSONSeq for all tests that are parametrized.  
Just for testing that the extension is auto-recognized, I think we could also have one targeted test, and here only have one extension for this?

jorisvandenbossche (commented):
Thanks!

pyogrio/tests/conftest.py
theroggy: True. I removed the ".json" ext in ``ALL_EXTS`` for the same reason, but apparently wasn't quite awake when I added both GeoJsonSeq exts...

brendan-ward:
It looks like parts of this are superseded by changes in #75?  Except `test_autodetect_driver`?

theroggy:
> It looks like parts of this are superseded by changes in #75? Except `test_autodetect_driver`?

Yes, Joris suggested to put this in a seperate PR as it is unrelated to #75.

pyogrio/tests/test_raw_io.py
```diff
@@ -6,8 +6,38 @@
...
+
+
+@pytest.mark.parametrize("ext", DRIVERS)
+def test_autodetect_driver(tmp_path, naturalearth_lowres, ext):
+    # Test all supported autodetect drivers
+    testfile = prepare_testfile(naturalearth_lowres, dst_dir=tmp_path, ext=ext)
+
+    assert testfile.suffix == ext
+    assert testfile.exists()
+    meta, _, geometry, fields = read(naturalearth_lowres)
```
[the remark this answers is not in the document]
theroggy: Indeed! Fixed.

jorisvandenbossche (approved):
Thanks!

PR #108, opened 2022-05-16
Add --no-use-pep517 to get around pip oldest-supported-numpy issue

### LINKED ISSUE #107: CI: Docker-based tests failing due to oldest-supported-numpy
It appears that we are now running into issues trying to install `oldest-supported-numpy` in our GDAL docker test containers.

See: https://github.com/geopandas/pyogrio/runs/6455834609?check_suite_focus=true#step:5:27

brendan-ward:
Resolves #107 

As I understand it, a recent pip upgrade now runs into issues with `oldest-supported-numpy` declared in `pyproject.toml` when running with `--no-build-isolation` when `numpy` isn't yet installed.  The fix is to add `--no-use-pep517`.

This updates to do the same thing as pandas: https://github.com/pandas-dev/pandas/pull/47015

### DIFF
--- a/.github/workflows/docker-gdal.yml
+++ b/.github/workflows/docker-gdal.yml
@@ -33,8 +33,7 @@ jobs:
       - name: Install Python Dependencies
         run: |
           python3 -m pip install --no-cache-dir -U pip wheel
-          python3 -m pip install --no-cache-dir cython~=0.29
-          python3 -m pip install --no-cache-dir --no-build-isolation --no-use-pep517 -e .[dev,test,geopandas]
+          python3 -m pip install --no-cache-dir -e .[dev,test,geopandas]
 
       - name: Test with pytest
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-gdal.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: |
           python3 -m pip install --no-cache-dir -U pip wheel
           python3 -m pip install --no-cache-dir cython~=0.29
-          python3 -m pip install --no-cache-dir --no-build-isolation -e .[dev,test,geopandas]
+          python3 -m pip install --no-cache-dir --no-build-isolation --no-use-pep517 -e .[dev,test,geopandas]
```
jorisvandenbossche, line 37: Alternatively, we could also remove the `--no-build-isolation` altogether and test the default pip installation in this workflow (and then we can also remove the cython install step on the line above)

brendan-ward: Yeah, that's cleaner.

PR #116, opened 2022-06-02
TST: make sure that path tests run without geopandas

brendan-ward:
Resolves #116 

This splits out path tests to use the basic functions `raw.read`, `read_info`, `read_bounds` without GeoPandas being installed, and adds tests that use GeoPandas for each of those but marked to skip if GeoPandas is not present.  It is maybe a bit excessive to test all 3 basic functions every time, but not necessarily a bad thing.

### DIFF
--- a/pyogrio/tests/test_path.py
+++ b/pyogrio/tests/test_path.py
@@ -281,8 +281,7 @@ def test_url():
 def test_url_dataframe():
     url = "https://raw.githubusercontent.com/geopandas/pyogrio/main/pyogrio/tests/fixtures/naturalearth_lowres/naturalearth_lowres.shp"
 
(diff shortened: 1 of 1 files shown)


brendan-ward:
This was tested locally without GeoPandas; I'm not sure it is useful to add another CI test that does not include GeoPandas.  Should we?

jorisvandenbossche:
> This was tested locally without GeoPandas; I'm not sure it is useful to add another CI test that does not include GeoPandas. Should we?

It might be worth adding that (to avoid that we run into the same issue again), if it doesn't add too much complexity to the CI setup. Doing an attempt at https://github.com/geopandas/pyogrio/pull/132

PR #118, opened 2022-06-06
Support writing object columns with np.nan values

### LINKED ISSUE #60: Writing a dataframe with a string column with np.nan value gives an error
Writing a dataframe with string column with np.nan values in it gives an Exception: ValueError: "Could not encode value 'nan' in field 'PM' to string".

The following code triggers the Exception:
```
import geopandas as gpd
import numpy as np
import pyogrio
import shapely.geometry

output_path = r"C:\temp\test_write_nan.gpkg"
test_gdf = gpd.GeoDataFrame(
        geometry=[
                shapely.geometry.Polygon(shell=[(1, 1), (1, 2), (2, 2), (2, 1), (1, 1)]),
                shapely.geometry.Polygon(shell=[(1, 1), (1, 2), (2, 2), (2, 1), (1, 1)])],
        crs="epsg:31370")
test_gdf["PM"] = "foo"
test_gdf.at[1, "PM"] = np.nan
pyogrio.write_dataframe(test_gdf, output_path)
```

theroggy:
Closes #60

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 -   Use certifi to set `GDAL_CURL_CA_BUNDLE` / `PROJ_CURL_CA_BUNDLE` defaults (#97)
 -   automatically detect driver for `.geojson`, `.geojsonl` and `.geojsons` files (#101)
 -   read DateTime fields with millisecond accuracy (#111)
+-   support writing object columns with np.nan values (#60)
 
 ### Breaking changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,7 @@
 -   Use certifi to set `GDAL_CURL_CA_BUNDLE` / `PROJ_CURL_CA_BUNDLE` defaults (#97)
 -   automatically detect driver for `.geojson`, `.geojsonl` and `.geojsons` files (#101)
 -   read DateTime fields with millisecond accuracy (#111)
--   support writing object columns with np.nan values (#60)
+-   support writing object columns with np.nan values (#118)
 
 ### Breaking changes
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -14,6 +14,7 @@ import warnings
 from libc.stdint cimport uint8_t
 from libc.stdlib cimport malloc, free
 from libc.string cimport strlen
+from libc.math cimport isnan
 
 cimport cython
 import numpy as np
@@ -1364,17 +1365,19 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
 
                 elif field_type == OFTString:
                     # TODO: encode string using approach from _get_internal_encoding which checks layer capabilities
-                    try:
-                        # this will fail for strings mixed with nans
-                        value_b = field_value.encode("UTF-8")
+                    if isinstance(field_value, float) and isnan(field_value):
+                        OGR_F_SetFieldNull(ogr_feature, field_idx)
+                    else:
 
-                    except AttributeError:
-                        raise ValueError(f"Could not encode value '{field_value}' in field '{fields[field_idx]}' to string")
+                        try:
+                            value_b = field_value.encode("UTF-8")
+                            OGR_F_SetFieldString(ogr_feature, field_idx, value_b)
 
-                    except Exception:
-                        raise
+                        except AttributeError:
+                            raise ValueError(f"Could not encode value '{field_value}' in field '{fields[field_idx]}' to string")
 
-                    OGR_F_SetFieldString(ogr_feature, field_idx, value_b)
+                        except Exception:
+                            raise
 
                 elif field_type == OFTInteger:
                     OGR_F_SetFieldInteger(ogr_feature, field_idx, field_value)
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1360,12 +1360,12 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
                 field_value = field_data[field_idx][i]
                 field_type = field_types[field_idx][0]
 
-                if field_value is None:
-                    OGR_F_SetFieldNull(ogr_feature, field_idx)
-
-                elif field_type == OFTString:
+                if field_type == OFTString:
                     # TODO: encode string using approach from _get_internal_encoding which checks layer capabilities
-                    if isinstance(field_value, float) and isnan(field_value):
+                    if (
+                        field_value is None 
+                        or (isinstance(field_value, float) and isnan(field_value))
+                    ):
                         OGR_F_SetFieldNull(ogr_feature, field_idx)
 
                     else:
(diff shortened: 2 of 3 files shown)


pyogrio/_io.pyx
```diff
@@ -1364,17 +1365,19 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
 
                 elif field_type == OFTString:
                     # TODO: encode string using approach from _get_internal_encoding which checks layer capabilities
-                    try:
-                        # this will fail for strings mixed with nans
-                        value_b = field_value.encode("UTF-8")
+                    if isinstance(field_value, float) and isnan(field_value):
+                        OGR_F_SetFieldNull(ogr_feature, field_idx)
```
jorisvandenbossche, line 1369: Can we combine this with the `field_value is None` check above to set FieldNull in a single place? 
(or move that check here, as currently the only way to get a None is through object dtype)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -704,3 +704,26 @@ def test_custom_crs_io(tmpdir, naturalearth_lowres_all_ext):
...
+
+@pytest.mark.parametrize("ext", ALL_EXTS)
+def test_write_read_null(tmp_path, ext):
+    from shapely.geometry import Point
+
+    output_path = tmp_path / f"test_write_nan.{ext}"
+    geom = Point(0, 0)
+    test_data = {
+        "geometry": [geom, geom, geom],
+        "float64": [1.0, None, np.nan],
```
jorisvandenbossche, line 717: This doesn't have any effect in practice, as that will get converted to twice a NaN by pandas:

```
In [10]: pd.Series([1.0, None, np.nan])
Out[10]: 
0    1.0
1    NaN
2    NaN
dtype: float64
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -704,3 +704,26 @@ def test_custom_crs_io(tmpdir, naturalearth_lowres_all_ext):
     assert crs["lat_2"] == 51.5
     assert crs["lon_0"] == 4.3
     assert df.crs.equals(expected.crs)
+
+
+@pytest.mark.parametrize("ext", ALL_EXTS)
```
jorisvandenbossche, line 709: Would it be sufficient to test with a single driver in this case? 
The code affected is purely on the python (cython) side, determining when to indicate to GDAL to set a null or not, which shouldn't have driver specific behaviour? 

(I know it is not always easy in advance to predict when something actually runs into a driver specific issue, so we might miss things this way, but I also want to avoid the test suite becoming unnecessarily bloated; it's nice that the tests run fast)

jorisvandenbossche (commented):
Thanks for working on this!

CHANGES.md
```diff
@@ -22,6 +22,7 @@
 -   Use certifi to set `GDAL_CURL_CA_BUNDLE` / `PROJ_CURL_CA_BUNDLE` defaults (#97)
 -   automatically detect driver for `.geojson`, `.geojsonl` and `.geojsons` files (#101)
 -   read DateTime fields with millisecond accuracy (#111)
+-   support writing object columns with np.nan values (#60)
```
brendan-ward, line 25:
```suggestion
-   support writing object columns with np.nan values (#118)
```

(I've generally been putting the PRs instead of originating issues in the changelog, but then this requires first creating the PR before then adding an etry here...)

pyogrio/_io.pyx
brendan-ward: It seems like the check for `isnan` should be in the `elif field_type == OFTReal:` block?  The field type for the incoming column should be `float32` / `float64` rather than `object` (though I haven't verified this)

brendan-ward (commented):
Thanks for working on this @theroggy ; a few comments to add

pyogrio/_io.pyx
jorisvandenbossche: @brendan-ward note that the specific case that is being fixed here is if you have an `object` dtype column (but which contains NaN, pandas doesn't really distinguish None vs NaN in object columns)

pyogrio/tests/test_geopandas_io.py
theroggy: Yes, that's indeed the case. Nonetheless I think it is transparant that it is explicitly in the test?

But obviously if you think that's better I can put 2 times np.nan or whatever...

pyogrio/tests/test_geopandas_io.py
theroggy: I added the different file types to check that for all formats the roundtrip of reading + writing the null/nan/None values behave the same... but possibly we can trust gdal to show consistent behaviour for this, so I'll remove it.

pyogrio/_io.pyx
theroggy: @brendan-ward in addition to what @jorisvandenbossche wrote: np.nan values in an OFTReal (float) column are automatically treated correctly (as null) by GDAL, so no need to explicitly call OGR_F_SetFieldNull for OFTReal columns.

brendan-ward (approved):
Thanks @theroggy !

pyogrio/_io.pyx
jorisvandenbossche: > np.nan values in an OFTReal (float) column are automatically treated correctly (as null) by GDAL (OGR_F_SetFieldDouble), so no need to explicitly call OGR_F_SetFieldNull for OFTReal columns. Only for object columns this is needed.

Actually, that's not fully correct I think. They are written as NaN and not as Null. Since numpy/pandas only support NaN in float arrays, that's not an issue for correct rountrip for geopandas->gdal->geopandas, though.

Small illustrations:

```
# write file with GDAL + pyogrio
import geopandas
import pyogrio
gdf = geopandas.GeoDataFrame({'col': [0.1, np.nan]}, geometry=geopandas.points_from_xy([0, 1], [0, 1]))
pyogrio.write_dataframe(gdf, "test_nulls_pyogrio.arrow", driver="Arrow")

# write file with pyarrow that includes both NaN and Null (Arrow distinguishes both)
import pyarrow as pa
from pyarrow import feather
feather.write_feather(pa.table({"col": [0.1, np.nan, None]}), "test_nulls_pyarrow.arrow")
```

And check both using GDAL's `ogrinfo`:

```
(gdal-dev) $ ogrinfo test_nulls_pyogrio.arrow -al
INFO: Open of `test_nulls_pyogrio.arrow'
      using driver `Arrow' successful.

Layer name: test_nulls_pyogrio
Geometry: Point
Feature Count: 2
Extent: (0.000000, 0.000000) - (1.000000, 1.000000)
Layer SRS WKT:
(unknown)
Geometry Column = geometry
col: Real (0.0)
OGRFeature(test_nulls_pyogrio):0
  col (Real) = 0.1
  POINT (0 0)

OGRFeature(test_nulls_pyogrio):1
  col (Real) = nan
  POINT (1 1)

(gdal-dev) $ ogrinfo test_nulls_pyarrow.arrow -al
INFO: Open of `test_nulls_pyarrow.arrow'
      using driver `Arrow' successful.

Layer name: test_nulls_pyarrow
Geometry: None
Feature Count: 3
Layer SRS WKT:
(unknown)
col: Real (0.0)
OGRFeature(test_nulls_pyarrow):0
  col (Real) = 0.1

OGRFeature(test_nulls_pyarrow):1
  col (Real) = nan

OGRFeature(test_nulls_pyarrow):2
  col (Real) = (null)

jorisvandenbossche: Opened https://github.com/geopandas/pyogrio/issues/122 to further track this

PR #120, opened 2022-06-06
Support writing datetime columns

### LINKED ISSUE #58: ENH: support writing datetime columns in write_dataframe
Reading datetime columns seems to work, but when they are written again the following error is thrown: Exception has occurred: NotImplementedError field type is not supported datetime64[ns].

I just saw it is mentioned as a todo in the readme, but I suppose it is useful to have an issue for it as well?

theroggy:
Closes #58

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -76,10 +76,10 @@ DTYPE_OGR_FIELD_TYPES = {
 
     'float32': (OFTReal,OFSTFloat32),
     'float': (OFTReal, OFSTNone),
-    'float64': (OFTReal, OFSTNone)
+    'float64': (OFTReal, OFSTNone),
 
-    'datetime64[D]', (OFTDate, OFSTNone),
-    'datetime64[ms]',(OFTDateTime, OFSTNone),
+    'datetime64[D]': (OFTDate, OFSTNone),
+    'datetime64[ms]': (OFTDateTime, OFSTNone),
 }
 
 
@@ -1387,7 +1387,22 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
                 elif field_type == OFTReal:
                     OGR_F_SetFieldDouble(ogr_feature, field_idx, field_value)
 
-                elif field_type in (OFTDate, OFTDateTime):
+                elif field_type == OFTDate:
+                    datetime = field_value.astype(object)
+                    OGR_F_SetFieldDateTimeEx(
+                        ogr_feature,
+                        field_idx,
+                        datetime.year,
+                        datetime.month,
+                        datetime.day,
+                        0,
+                        0,
+                        0.0,
+                        0
+                    )
+
+                elif field_type == OFTDateTime:
+                    # TODO: what to do with tzinfo?
                     datetime = field_value.astype(object)
                     OGR_F_SetFieldDateTimeEx(
                         ogr_feature,
@@ -1398,9 +1413,9 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
                         datetime.hour,
                         datetime.minute,
                         float(f"{datetime.second}.{datetime.microsecond}"),
-                        datetime.tzinfo
+                        0
                     )
-                
+
                 else:
                     raise NotImplementedError(f"OGR field type is not supported for writing: {field_type}")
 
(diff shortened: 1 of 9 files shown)


pyogrio/_io.pyx
```diff
@@ -1385,6 +1392,35 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
...
+                        datetime.month,
+                        datetime.day,
+                        0,
+                        0,
+                        0.0,
+                        0
+                    )
+
+                elif field_type == OFTDateTime:
+                    # TODO: what to do with tzinfo?
```
jorisvandenbossche, line 1410: See https://github.com/Toblerity/Fiona/blob/791785b813fce97e5c3d60bd725fa2190645d2e3/fiona/ogrext.pyx#L401-L404 and https://github.com/Toblerity/Fiona/blob/791785b813fce97e5c3d60bd725fa2190645d2e3/fiona/ogrext.pyx#L431-L435 how fiona handles this (and https://gdal.org/development/rfc/rfc56_millisecond_precision.html?highlight=TZFlag#core-changes with very brief details)

I think it is also certainly fine to leave this for a separate PR, given the additional complexity for handling timezones

pyogrio/_io.pyx
```diff
@@ -1385,6 +1392,35 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
...
+                    datetime = field_value.astype("datetime64[ms]").astype(object)
+                    OGR_F_SetFieldDateTimeEx(
+                        ogr_feature,
+                        field_idx,
+                        datetime.year,
+                        datetime.month,
+                        datetime.day,
+                        datetime.hour,
+                        datetime.minute,
+                        float(f"{datetime.second}.{datetime.microsecond:06d}"),
```
jorisvandenbossche, line 1420: Can you do this with a calculation (something like datetime.second + datetime.microsecond / 10**6) instead of string formatting+parsing? (or is there a specific reason for the latter?)

pyogrio/_io.pyx
```diff
@@ -1385,6 +1392,35 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
                 elif field_type == OFTReal:
                     OGR_F_SetFieldDouble(ogr_feature, field_idx, field_value)
 
+                elif field_type == OFTDate:
+                    datetime = field_value.astype(object)
```
jorisvandenbossche, line 1396:
```suggestion
                    datetime = field_value.item()
```

Using the `item()` method to get a Python scalar seems slightly faster

pyogrio/tests/test_raw_io.py
```diff
@@ -422,6 +422,31 @@ def test_read_write_data_types_numeric(tmp_path, ext):
...
 
+def test_read_write_datetime(tmp_path):
+    field_data = [
+        np.array(["2005-02-01", "2005-02-02"], dtype="datetime64[D]"),
+        np.array(["2001-01-01T12:00", "2002-02-03T13:56:03"], dtype="datetime64[s]"),
+        np.array(
+            ["2001-01-01T12:00", "2002-02-03T13:56:03.072"], dtype="datetime64[ms]"
+        ),
+        np.array(
+            ["2001-01-01T12:00", "2002-02-03T13:56:03.072"], dtype="datetime64[ns]"
```
jorisvandenbossche, line 433: Maybe also include one additional case with datetime64[ns] data that actually has nanosecond resolution (the sub-millisecond details will get lost)?

jorisvandenbossche (commented):
Nice!

pyogrio/_io.pyx
brendan-ward: Agreed on leaving for a separate PR; just please add a note to `docs/known_issues.md` to note that timezones aren't yet supported.

brendan-ward (commented):
Thanks for working on this @theroggy !  

Please also update `docs/known_issues.md` to note that it is now possible to write datetime values, but only at `ms` precision.  Also probably a good idea to note that some drivers do not support full datetime and may instead store only date values (e.g., shapefile); we may further want to note that if needed to store the original values, they should encode them to a string (it seems the [recommendation](https://gis.stackexchange.com/questions/267413/storing-datetime-values-in-shapefiles-using-qgis) to store full datetime values is to put them in a string)

pyogrio/_io.pyx
theroggy: That's indeed cleaner...

pyogrio/_io.pyx
theroggy: I created #123...

jorisvandenbossche (approved):

PR #125, opened 2022-06-07
Support object columns containing non-string values (eg. float, datetime,...)

### LINKED ISSUE #119: Add support to write object columns that contain types different than string
Writing an object column containing different types than strings (eg. float, int,...) now gives an error.

theroggy:
Closes #119

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -183,7 +183,7 @@ def write_dataframe(
     ----------
     df : GeoDataFrame
         The data to write. For attribute columns of the "object" dtype, all values will
-        be converted to strings to be written to the output file.
+        be converted to strings to be written to the output file, except None and np.nan, which will be set to NULL in the output file.
     path : str
         path to file
     layer :str, optional (default: None)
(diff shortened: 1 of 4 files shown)


brendan-ward (approved):
Thanks @theroggy

jorisvandenbossche:
Should we document this somewhere that for mixed types in object dtype, the string representation is stored? (so no inference of the possible type) 
(haven't yet checked our docs to think about where this would fit in)

theroggy:
> Should we document this somewhere that for mixed types in object dtype, the string representation is stored? (so no inference of the possible type) (haven't yet checked our docs to think about where this would fit in)

Good idea. I gave it a try in the inline documentation of write_dataframe().

pyogrio/geopandas.py
```diff
@@ -181,6 +181,9 @@ def write_dataframe(
 
     Parameters
     ----------
+    df : GeoDataFrame
+        The data to write. For attribute columns of the "object" dtype, all values will
+        be converted to strings to be written to the output file.
```
brendan-ward, line 186:
```suggestion
        be converted to strings to be written to the output file, except None and np.nan, which will be set to NULL in the output file.
```

(not sure if the distinction is important, but clarifies that `None` is set to null rather than `"None"`)

theroggy: Ok, sounds good.

PR #126, opened 2022-06-07
CI: Enable linting on CI using GeoPandas configuration

brendan-ward:
This copies the linting CI step and pre-commit config from GeoPandas.  Includes auto-formatting using Black and manual updates to make Flake8 happy.

Note: all changes in `_version.py` are autoformatting and can be ignored.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,17 @@
+name: Linting
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  Linting:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+      - uses: pre-commit/action@v3
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,4 +14,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3
-      - uses: pre-commit/action@v3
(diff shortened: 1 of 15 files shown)


pyogrio/__init__.py
```diff
@@ -15,3 +15,18 @@
 
 __version__ = get_versions()["version"]
 del get_versions
+
+__all__ = [
+    list_drivers,
```
jorisvandenbossche, line 20: this need to be strings, I think

jorisvandenbossche:
Thanks! I have long been thinking to do the same ;) This will be useful to smooth development

jorisvandenbossche:
This can be un-drafted?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Linting
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
```
martinfleis, line 8: Can these be empty?

brendan-ward: Yes; these appear to be working as expected and I think I found similar things as examples in the GH action docs.  I may not be remembering correctly, but I think I got a syntax error when I tried to omit the `:`

jorisvandenbossche: Yes, I used that for the release workflow as well:

https://github.com/geopandas/pyogrio/blob/7fa0ded85f34f8821c8a57e6aee3faa8df133aa3/.github/workflows/release.yml#L16

This workflow_dispatch ensures you can trigger the action manually from github (without creating a new PR or so), which is useful for eg ensuring the release workflow is passing

jorisvandenbossche (approved):

PR #129, opened 2022-06-09
Speed up doc builds by using mambaforge

theroggy:
When specifically doing updates on the docs, I noticed the building takes quite some time.

This should speed this up (but I cannot verify exact timings).


martinfleis (approved):

brendan-ward (approved):
Thanks @theroggy !

PR #130, opened 2022-06-09
Remove noindex option under geopandas section so that permalinks work

weiji14:
Hi there, I was trying to add an intersphinx link using "py:func:\`pyogrio.read_dataframe\`" at https://github.com/weiji14/zen3geo/pull/19#discussion_r893027681 but it seems like there wasn't a permalink available. The closest link I could find was https://pyogrio.readthedocs.io/en/latest/api.html#geopandas-integration, but ideally it would be possible to link directly to `pyorgio.read_dataframe`.

This Pull Request removes the `:noindex:` option added in #18 (specifically at commit 679b81b880f744960efaa81b8f4e4b9ac92d2b91) so that intersphinx will work and the permalink shows up, like so:

Screenshot at 2022-06-09 11-34-53

And you can then click on https://pyogrio--130.org.readthedocs.build/en/130/api.html#pyogrio.read_dataframe!

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -10,6 +10,5 @@ Core
 GeoPandas integration
 ---------------------
 
-.. automodule:: pyogrio.geopandas
 .. autofunction:: pyogrio.read_dataframe
 .. autofunction:: pyogrio.write_dataframe


brendan-ward:
@weiji14 this passed all tests; is it ready for review?

weiji14:
Yep, just another one-line commit. Noticed that the "Geopandas integration" subsection had a permalink like https://pyogrio--130.org.readthedocs.build/en/130/api.html#module-0, but it should be https://pyogrio--130.org.readthedocs.build/en/130/api.html#module-pyogrio.geopandas instead.

weiji14:
Oh wait, now the docs say `pyogrio.geopandas.read_dataframe` which is probably not what we want. Hold on, give me a few more minutes...

jorisvandenbossche:
Yes, we would prefer to use the public "pyogrio.read_dataframe" path being shown

weiji14:
Ok, things look better now, just needed to use [`.. autofunction::`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autofunction). Links are:

- https://pyogrio--130.org.readthedocs.build/en/130/api.html#module-pyogrio.geopandas
- https://pyogrio--130.org.readthedocs.build/en/130/api.html#pyogrio.read_dataframe
- https://pyogrio--130.org.readthedocs.build/en/130/api.html#pyogrio.write_dataframe

docs/source/api.rst
```diff
@@ -10,6 +10,6 @@ Core
 GeoPandas integration
 ---------------------
 
-.. automodule:: pyogrio
-    :members: read_dataframe, write_dataframe
-    :noindex:
+.. automodule:: pyogrio.geopandas
```
martinfleis, line 13:
```suggestion
```

I believe that this line can be removed since you are listing both functions manually below.

weiji14: Ok, let's see

weiji14: So instead of https://pyogrio--130.org.readthedocs.build/en/130/api.html#module-pyogrio.geopandas, the link now becomes https://pyogrio--130.org.readthedocs.build/en/130/api.html#geopandas-integration. Is that ok?

martinfleis (approved):
Thanks!

PR #131, opened 2022-06-09
DOC: Add contributors section to changelog

brendan-ward:
This follows the model of pandas / numpy of including a contributors section to each release in the changelog, now that we have more contributors.  I only added this for 0.4.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,15 @@
 -   fix passing gdal creation parameters in `write_dataframe` (#62)
 -   fix passing kwargs to GDAL in `write_dataframe` (#67)
 
+### Contributors
+
+People with a “+” by their names contributed a patch for the first time.
+
+-   Brendan Ward
+-   Joris Van den Bossche
+-   Martin Fleischmann
+-   Pieter Roggemans +
+
 ## 0.3.0
 
 ### Major enhancements


CHANGES.md
```diff
@@ -51,6 +51,15 @@
...
 
+### Contributors
+
+People with a “+” by their names contributed a patch for the first time.
+
+-   Brendan Ward
+-   Joris Van den Bossche
+-   Martin Fleischmann
+-   Pieter Roggemans +
+
```
weiji14, line 62: Here's my full name :wink:
```suggestion
-   Wei Ji Leong +

```

CHANGES.md
```diff
@@ -51,6 +51,15 @@
 -   fix passing gdal creation parameters in `write_dataframe` (#62)
 -   fix passing kwargs to GDAL in `write_dataframe` (#67)
 
+### Contributors
+
+People with a “+” by their names contributed a patch for the first time.
```
weiji14, line 56: FYI, this is actually meant to be generated using https://github.com/pandas-dev/pandas/blob/v1.4.2/doc/sphinxext/contributors.py.

jorisvandenbossche: @brendan-ward I do this for geopandas as well (although I only seem to include it in the github release notes, and not in the changelog file (I assume for length purposes, but that's not (yet) an issue here): https://github.com/geopandas/geopandas/releases/tag/v0.10.0 
I have a script locally to create that (based on the pandas one), will share that.

PR #134, opened 2022-06-11
CI: Fix minimal test env and remove deps on pandas / geopandas

brendan-ward:
Supersedes #132 (thanks @jorisvandenbossche !)

Cleans up a dependency on pandas outside of our try / catch block for GeoPandas, and uses raw IO methods for preparing test files instead of geopandas

### DIFF
--- a/pyogrio/tests/conftest.py
+++ b/pyogrio/tests/conftest.py
@@ -4,7 +4,7 @@ from zipfile import ZipFile, ZIP_DEFLATED
 import pytest
 
 from pyogrio import __gdal_version_string__, __version__, list_drivers
-import pyogrio
+from pyogrio.raw import read, write
 
 
 _data_dir = Path(__file__).parent.resolve() / "fixtures"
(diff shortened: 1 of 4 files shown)


pyogrio/tests/conftest.py
```diff
@@ -27,12 +27,16 @@ def prepare_testfile(testfile_path, dst_dir, ext):
...
+    meta, _, geometry, field_data = read(testfile_path)
+
     if ext == ".fgb":
         # For .fgb, spatial_index=False to avoid the rows being reordered
-        pyogrio.write_dataframe(gdf, dst_path, spatial_index=False)
-    else:
-        pyogrio.write_dataframe(gdf, dst_path)
+        meta["spatial_index"] = False
+        # allow mixed Polygons/MultiPolygons type
+        meta["geometry_type"] = "Unknown"
```
jorisvandenbossche, line 37: This one line is a change compared to what was there before? Or did `write_dataframe` infer such a geometry type? (I would have expected that it would infer "MultiPolgyon" and set promote_to_multi to true)

brendan-ward: The detection and promotion happens in `write_geodataframe` and doesn't happen in the case of the raw `write`.

With a little bit of refactoring we could move that into `write` instead (I think this was even suggested when that was first added), but I didn't think we needed to do so prior to 0.4 (changes noted in the changelog are specific to GeoDataFrame)

PR #140, opened 2022-07-15
FIX: Handle nonlinear types that are not automatically converted

brendan-ward:
Even though we set `OGRSetNonLinearGeometriesEnabledFlag(0)` [ref](https://gdal.org/api/vector_c_api.html#_CPPv436OGRSetNonLinearGeometriesEnabledFlagi) before reading any geometries, it does not always seem to work for data sources that have mixed linear and nonlinear geometries (example failing [GDB](https://prd-tnm.s3.amazonaws.com/StagedProducts/Hydrography/NHDPlusHR/Beta/GDB/NHDPLUS_H_0307_HU4_GDB.zip), `layer=NHDWaterbody`).

This fix checks for and forces a conversion to linear types when reading geometries.

Since GEOS doesn't support nonlinear geometry types, it is a bit tricky to create a test case for this (open to ideas!), but this was verified using the dataset at the link above.


pyogrio/_ogr.pxd
```diff
@@ -240,22 +240,25 @@ cdef extern from "ogr_api.h":
...
-    OGRErr          OGR_G_ImportFromWkb(OGRGeometryH geometry, const void *bytes, int nbytes)
-    int             OGR_G_IsMeasured(OGRGeometryH geometry)
-    void            OGR_G_SetMeasured(OGRGeometryH geometry, int isMeasured)
-    int             OGR_G_Is3D(OGRGeometryH geometry)
-    void            OGR_G_Set3D(OGRGeometryH geometry, int is3D)
-    int             OGR_G_WkbSize(OGRGeometryH geometry)
-    OGRGeometryH    OGR_G_ForceToMultiPoint(OGRGeometryH geometry)
-    OGRGeometryH    OGR_G_ForceToMultiLineString(OGRGeometryH geometry)
-    OGRGeometryH    OGR_G_ForceToMultiPolygon(OGRGeometryH geometry)
+    OGRGeometryH        OGR_G_CreateGeometry(int wkbtypecode)
```
brendan-ward, line 243: Most of this diff is a white space change to handle consistent indents; additions are `OGR_G_GetGeometryType`, `OGR_G_GetLinearGeometry`, `OGR_GT_IsNonLinear`

pyogrio/_io.pyx
```diff
@@ -473,6 +473,9 @@ cdef validate_feature_range(OGRLayerH ogr_layer, int skip_features=0, int max_fe
 cdef process_geometry(OGRFeatureH ogr_feature, int i, geom_view, uint8_t force_2d):
 
     cdef OGRGeometryH ogr_geometry = NULL
+    cdef OGRwkbGeometryType ogr_geometry_type
+    cdef OGRwkbGeometryType linear_geometry_type
```
jorisvandenbossche, line 477: This second one isn't used?

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 0.4.1 (in progress)
+
+### Bug fixes
+
+-   always convert non-linear geometry types to linear geometry types (#140)
```
jorisvandenbossche, line 7: On first read, I found the "non-linear" confusing (as I interpreted "linear" as LineString/LinearRing geometry types), so maybe add a "(for example, Curve, Surface, etc)" ?

jorisvandenbossche (commented):
Looks good!

jorisvandenbossche:
> Since GEOS doesn't support nonlinear geometry types, it is a bit tricky to create a test case for this (open to ideas!), but this was verified using the dataset at the link above.

One option would be to include a tiny dataset that has such data (eg you could, using GDAL, extract a small subset of the failing GDB layer?)

jorisvandenbossche (approved):

PR #148, opened 2022-08-18
CI: Fix tests on GDAL Docker container

brendan-ward:
Editable installs are failing with latest version of pip due to presence of `pyproject.toml` file but without properly set version number.

This uses `--no-use-pep517` to sidestep some of the enforcement around recent python packaging standards.

Resolves build error noted in #146

### DIFF
--- a/.github/workflows/docker-gdal.yml
+++ b/.github/workflows/docker-gdal.yml
@@ -38,8 +38,8 @@ jobs:
 
       - name: Install Python Dependencies
         run: |
-          python3 -m pip install --no-cache-dir -U pip wheel setuptools
-          python3 -m pip install --no-cache-dir -e .[dev,test,geopandas]
+          python3 -m pip install --no-cache-dir -U pip wheel
+          python3 -m pip install --no-cache-dir --no-use-pep517 -e .[dev,test,geopandas]
 
       - name: Test with pytest
         run: |
--- a/.github/workflows/docker-gdal.yml
+++ b/.github/workflows/docker-gdal.yml
@@ -30,16 +30,16 @@ jobs:
       image: ${{ matrix.container }}
 
     steps:
-      - uses: actions/checkout@v3
-
       - name: Install packages
         run: |
-          apt-get update && apt-get install -y python3-pip
+          apt-get update && apt-get install -y git python3-pip
+
+      - uses: actions/checkout@v3
 
       - name: Install Python Dependencies
         run: |
           python3 -m pip install --no-cache-dir -U pip wheel
-          python3 -m pip install --no-cache-dir --no-use-pep517 -e .[dev,test,geopandas]
+          python3 -m pip install --no-cache-dir -e .[dev,test,geopandas]
 
       - name: Test with pytest
         run: |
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
My guess why it is failing: for some reason on the gdal docker image build, the checkout action is not downloading pyogrio as a git repo:

```
Run actions/checkout@v3
/usr/bin/docker exec  291871c385c1be548a23271e0df1e2376db935aaed7ceda0965438e80012972e sh -c "cat /etc/*release | grep ^ID"
Syncing repository: geopandas/pyogrio
Getting Git version info
Deleting the contents of '/__w/pyogrio/pyogrio'
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Downloading the archive
Writing archive to disk
Extracting the archive
/usr/bin/tar xz -C /__w/pyogrio/pyogrio/dae5f575-73d0-48e6-a8c5-dc8c21af9fd5 -f /__w/pyogrio/pyogrio/dae5f575-73d0-48e6-a8c5-dc8c21af9fd5.tar.gz
Resolved version geopandas-pyogrio-9360374
```

While on the conda tests, the checkout actions does clone:

```
Run actions/checkout@v3
Syncing repository: geopandas/pyogrio
Getting Git version info
Temporarily overriding HOME='/home/runner/work/_temp/c404576f-8499-4247-a66e-c776be41590d' before making global git config changes
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /home/runner/work/pyogrio/pyogrio
Deleting the contents of '/home/runner/work/pyogrio/pyogrio'
Initializing the repository
Disabling automatic garbage collection
Setting up auth
Fetching the repository
Determining the checkout info
Checking out the ref
/usr/bin/git log -1 --format='%H'
'dcf0dd69f181bc36f7b625c71a8634ecc219d5cd'
```

And so probably versioneer cannot create a proper version, and that now started to fail with some stricter check from pip or setuptools?

Now, fixing it by disabling the pep517 usage is also fine, it's green so :shrug:

brendan-ward:
@jorisvandenbossche are you more familiar with what we should be doing to have both a `pyproject.toml` present but also use `versioneer` (via our `setup.py`) to set dynamic versions, in a way that is compatible with PEP440?  Using `--no-use-pep517` is a legacy option and feels like a hack here, but I wasn't able to find clear guidance about what we should do instead.

brendan-ward:
Sorry, concurrent comments.  :)

Thanks for picking up on the different in how the repository is pulled in.  I wonder if we should install git on the container before checkout, then it would work like on Conda?

jorisvandenbossche:
In geopandas we have those lines:

https://github.com/geopandas/geopandas/blob/0ca19e160f1aad0f1538ecdb5c604166aa1adfdd/setup.py#L11-L14

That's something we should probably add as well (although it actually doesn't seem to give problems in the other builds that we don't have it ...), but so I don't think that will solve the issues here as it is the actual version number that's the problem (see above).

(we may also not always have to install in editable mode for testing?)

brendan-ward:
Yep, installing Git before checkout was a cleaner fix here; no hacks with legacy options.

jorisvandenbossche (approved):
That's indeed nicer!

Issue #149, opened 2022-08-19
BUG: (sometimes) error in read on .gpkg when using a bbox filter


theroggy:
When reading a .gpkg with a bbox filter, some `read` operations trigger the following error:
`pyogrio.errors.FeatureError: Could not read feature at index ???`

After investigating a bit it seems that for .gpkg files OGR_L_GetFeatureCount returns the number of results doing a "coarse" bbox comparison (probably it just uses the spatial index), but when fetching the results with OGR_L_GetNextFeature it does a more thorough check (probably checks for an intersection).
Because the result of OGR_L_GetFeatureCount is used to determine the range of the loop in `_io.pyx::get_features()` this results in the error above.

For some other file types (.geojson, shapefile) this doesn't seem to give an issue at first sight. 

The gdal documentation for [OGR_L_GetFeatureCount](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_L_GetFeatureCount9OGRLayerHi) also gives a hint that the result of it is not 100% reliable: `For dynamic databases the count may not be exact.`

brendan-ward:
Is this consistently reproducible for the same GPKG?  If this is reproducible, perhaps characteristics of the failing record may help us here.  What version of GDAL is failing?

I had interpreted "dynamic" to mean that the actual set of records within the database may be changing when running the feature count or iterating over features; I assume that is not the case here?

It is possible that bounding boxes of a given geometry intersect that of the spatial filter and yet the actual geometry does not intersect the filter box.  But I'm surprised we haven't seen errors on this before, though maybe spatial filter has not yet been used extensively in practice for pyogrio (I rarely use it) and not sufficiently covered in our test cases for this sort of case.

We do need an accurate count in advance in order to set up the arrays to receive the data.  We currently set the `bForce` flag in 'OGR_L_GetFeatureCount` to get this even if expensive, so I'm unsure of what else we should do to get an accurate count.

theroggy:
> Is this consistently reproducible for the same GPKG? If this is reproducible, perhaps characteristics of the failing record may help us here.

Yes. As I mentioned above, the problem for the case I encountered is 
>>After investigating a bit it seems that for .gpkg files OGR_L_GetFeatureCount returns the number of results doing a "coarse" bbox comparison (probably it just uses the spatial index), but when fetching the results with OGR_L_GetNextFeature it does a more thorough check (probably checks for an intersection).

[...]

PR #151, opened 2022-08-20
FIX: Avoid duplicate count of features

brendan-ward:
As noted in #149 [comment](https://github.com/geopandas/pyogrio/issues/149#issuecomment-1221104056), we were calculating the count of features twice, which incurs extra and unnecessary overhead.  This removes that duplication.

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -607,6 +607,9 @@ cdef get_features(
     cdef int i
     cdef int field_index
 
+    # make sure layer is read from beginning
+    OGR_L_ResetReading(ogr_layer)
+
     if skip_features > 0:
         OGR_L_SetNextByIndex(ogr_layer, skip_features)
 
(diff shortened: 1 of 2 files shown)


pyogrio/_io.pyx
```diff
@@ -447,20 +447,20 @@ cdef validate_feature_range(OGRLayerH ogr_layer, int skip_features=0, int max_fe
...
-            raise ValueError("'max_features' must be >= 0")
+    elif max_features == 0:
+        max_features = feature_count - skip_features
 
-        if max_features > feature_count:
-            max_features = feature_count
+    elif max_features > feature_count:
+        max_features = feature_count
 
     return skip_features, max_features
```
jorisvandenbossche, line 465: One small comment: should we maybe rename this processed `max_features` (not the original keyword) to `num_features`, to make the actual meaning clearer?

brendan-ward: Yes, good idea.  Coming up... (sorry, should have requested review on this)

jorisvandenbossche: > sorry, should have requested review on this

no, it has been here for a long time, no problem to merge (it's easy to do follow-ups if needed)

PR #158, opened 2022-09-22
BLD: fix library detection on Windows / CI: Upgrade macos runner from deprecated 10.15

brendan-ward:
GitHub deprecated macOS 10.15 runners; this updates to 11 but retains the build target of 10.15, so everything should still work as before.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,8 +62,8 @@ def get_gdal_config():
         gdal_libs = ["gdal"]
 
         if platform.system() == "Windows":
-            # NOTE: libgdal on conda-forge now builds the library as "gdal" instead
-            # of "gdal_i", but older Windows builds still use "gdal_i"
+            # NOTE: if libgdal is built for Windows using CMake, it is now "gdal",
+            # but older Windows builds still use "gdal_i"
             if (Path(library_dir) / "gdal_i.lib").exists():
                 gdal_libs = ["gdal_i"]
 
--- a/setup.py
+++ b/setup.py
@@ -59,10 +59,16 @@ def get_gdal_config():
     gdal_version_str = os.environ.get("GDAL_VERSION")
 
     if include_dir and library_dir and gdal_version_str:
+        gdal_libs = ["gdal"]
+        if platform.system() == "Windows":
+            # NOTE: libgdal on conda-forge now builds the library as "gdal" instead
+            # of "gdal_i", but older Windows builds still use "gdal_i"
+            gdal_libs.append("gdal_i")
+
         return {
             "include_dirs": [include_dir],
             "library_dirs": [library_dir],
-            "libraries": ["gdal_i" if platform.system() == "Windows" else "gdal"],
+            "libraries": gdal_libs,
         }, gdal_version_str
 
     if include_dir or library_dir or gdal_version_str:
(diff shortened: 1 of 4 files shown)


brendan-ward:
CI madness: Windows Conda tests started failing when the only thing we changed was a macOS specific setting:

https://github.com/geopandas/pyogrio/actions/runs/3107179475/jobs/5035196379#step:5:91

Looks like the linker is not finding GDAL on Windows, even though it is exactly the same version of GDAL as for a preceding successful version of these tests: https://github.com/geopandas/pyogrio/actions/runs/3100662293/jobs/5021184008

brendan-ward:
Possible suspect: CMake builds just landed for GDAL in conda-forge: https://github.com/conda-forge/gdal-feedstock/pull/629
Perhaps the library is in a different location than expected or different filename.

jorisvandenbossche:
Yes, just wanted to mention the same about GDAL cmake landed on conda-forge.

Looking at this comment: https://github.com/conda-forge/gdal-feedstock/pull/629#issuecomment-1251165438, it lists the differences in files being installed before/after, and indeed for Windows, I see this part:

```
-Library/lib/gdal_i.lib
+Library/lib/cmake/gdal/GDAL-targets-release.cmake
+Library/lib/cmake/gdal/GDAL-targets.cmake
+Library/lib/cmake/gdal/GDALConfig.cmake
+Library/lib/cmake/gdal/GDALConfigVersion.cmake
+Library/lib/gdal.lib
+Library/lib/gdalplugins/drivers.ini
+Library/lib/pkgconfig/gdal.pc
```

so `gdal_i.lib` now being `gdal.lib`. 

That will require an update in our setup.py, since we currently hardcode that:

https://github.com/geopandas/pyogrio/blob/b60dbd5269bf560c9e650831372194d583496cd4/setup.py#L65

(not sure if we can just list both, or if there is some way to detect which of the two it should be)

brendan-ward:
Apparently we can't just list both, it fails on whichever one is not present.  I added a search for it, which appears to work for both the older and newer libgdal.

Should we comment out / remove the specific test for libgdal3.5.1 (which uses `gdal_i.lib`) from our conda-forge CI tests?  Now that we know that detection of `gdal_i.lib` appears to be working properly, I don't know that there is much value in retaining it - or at least running it all the time.

brendan-ward:
(release failures are unrelated, will work on those in a bit in a different PR)

jorisvandenbossche:
> Should we comment out / remove the specific test for libgdal3.5.1 (which uses `gdal_i.lib`) from our conda-forge CI tests? Now that we know that detection of `gdal_i.lib` appears to be working properly, I don't know that there is much value in retaining it - or at least running it all the time.

We could also pin gdal to an older version in the environment-minimal.yml, to not increase the number of builds?

jorisvandenbossche:
> (release failures are unrelated, will work on those in a bit in a different PR)

Are those related to the wheels not having GEOS enabled?

brendan-ward:
> We could also pin gdal to an older version in the environment-minimal.yml

I think we'd want this environment to be the same versions as `environment.yml`.

It's not a problem to keep the CI test for libgdal 3.5.1 on Windows in place, these aren't taking a long time.

brendan-ward:
> Are those related to the wheels not having GEOS enabled?

Yes, that appears to be it

jorisvandenbossche:
> I think we'd want this environment to be the same versions as `environment.yml`.

Is there a reason to want to have the same gdal version in the different env files?

---

In the meantime, those new conda-forge packages have been removed. But I think it would be good to already merge this fix anyway, so that if it changes again in the future (probably the migration will be done with the next gdal release).

setup.py
```diff
@@ -59,10 +59,18 @@ def get_gdal_config():
     gdal_version_str = os.environ.get("GDAL_VERSION")
 
     if include_dir and library_dir and gdal_version_str:
+        gdal_libs = ["gdal"]
+
+        if platform.system() == "Windows":
+            # NOTE: libgdal on conda-forge now builds the library as "gdal" instead
+            # of "gdal_i", but older Windows builds still use "gdal_i"
```
jorisvandenbossche, line 66: The difference is not necessarily specific to conda-forge, but the it's whether `cmake` is used to build GDAL versus `nmake` (the old build system, which support will be removed in the next version of GDAL). For example I suppose that also our wheels using vcpkg will start using "gdal" if we update the version there.

brendan-ward:
> Is there a reason to want to have the same gdal version in the different env files?

Just to control the things that vary between the tests; we don't _have_ to, it just makes it easier to determine that the reason something broke was related to minimal dependencies rather than also the version of GDAL.

PR #163, opened 2022-09-28
CI/TST: Fix bbox tests to avoid breaking when GEOS is not available

### LINKED ISSUE #162: Missing GEOS on MacOS / Linux release test instances breaks bbox tests
Side effect of #150 ; GDAL returns different results for a bbox filter when GEOS is available on the system.

All our main CI test instances have GEOS available, but we don't install additional binary dependencies on the wheel test instances (GDAL is packaged in the wheel but not GEOS).

I think we have 3 options:
- ensure that GEOS is available on the wheel test instance
- parameterize (or skip) the bbox tests to expect different behavior when GEOS is not available (`pyogrio.__gdal_geos_version__ is None`)
- adjust the bounds used for testing to avoid cases where bounding boxes overlap but geometries to not (leaning toward this option)

brendan-ward:
Resolves #162 

GDAL uses GEOS for intersection tests used by `bbox` filter when it is available, and falls back to a simple bounding box filter otherwise.

This updates the bbox tests to use a bbox that returns the same results regardless of whether GEOS is available or not.

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -177,50 +177,57 @@ with the bbox.
 
 Note: the `bbox` values must be in the same CRS as the dataset.
 
+Note: if GEOS is present and used by GDAL, only geometries that intersect `bbox`
+will be returned; if GEOS is not available or not used by GDAL, all geometries
+with bounding boxes that intersect this bbox will be returned.
+`pyogrio.__gdal_geos_version__` will be `None` if GEOS is not detected.
+
 ## Execute a sql query
 
-You can use the `sql` parameter to execute a sql query on a dataset. 
+You can use the `sql` parameter to execute a sql query on a dataset.
 
-Depending on the dataset, you can use different sql dialects. By default, if 
-the dataset natively supports sql, the sql statement will be passed through 
+Depending on the dataset, you can use different sql dialects. By default, if
+the dataset natively supports sql, the sql statement will be passed through
 as such. Hence, the sql query should be written in the relevant native sql
 dialect (e.g. [GeoPackage](https://gdal.org/drivers/vector/gpkg.html)/
-[Sqlite](https://gdal.org/drivers/vector/sqlite.html), 
-[PostgreSQL](https://gdal.org/drivers/vector/pg.html)). If the datasource 
-doesn't natively support sql (e.g. 
-[ESRI Shapefile](https://gdal.org/drivers/vector/shapefile.html), 
-[FlatGeobuf](https://gdal.org/drivers/vector/flatgeobuf.html)), you can choose 
-between '[OGRSQL](https://gdal.org/user/ogr_sql_dialect.html#ogr-sql-dialect)' 
-(the default) and  
-'[SQLITE](https://gdal.org/user/sql_sqlite_dialect.html#sql-sqlite-dialect)'. 
-For SELECT statements the 'SQLITE' dialect tends to provide more spatial 
-features as all 
-[spatialite](https://www.gaia-gis.it/gaia-sins/spatialite-sql-latest.html) 
+[Sqlite](https://gdal.org/drivers/vector/sqlite.html),
+[PostgreSQL](https://gdal.org/drivers/vector/pg.html)). If the datasource
+doesn't natively support sql (e.g.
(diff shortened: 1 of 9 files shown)


pyogrio/tests/test_core.py
```diff
@@ -138,22 +138,27 @@ def test_read_bounds_bbox(naturalearth_lowres_all_ext):
     assert fids.shape == (0,)
     assert bounds.shape == (4, 0)
 
-    fids, bounds = read_bounds(naturalearth_lowres_all_ext, bbox=(-140, 20, -100, 45))
+    fids, bounds = read_bounds(naturalearth_lowres_all_ext, bbox=(-85, 8, -80, 10))
 
     assert fids.shape == (2,)
     if naturalearth_lowres_all_ext.suffix == ".gpkg":
+        # GPKG results are in opposite order
```
jorisvandenbossche, line 145: For fgb we use `spatial_index=False` to avoid this ordering issue. The same might help for geopackage as well?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -317,12 +318,13 @@ def test_read_sql_columns_where_bbox(naturalearth_lowres_all_ext):
...
         columns=["iso_a3_renamed", "name"],
-        where="iso_a3_renamed IN ('CAN', 'USA', 'MEX')",
-        bbox=(-140, 20, -100, 45),
+        where="iso_a3_renamed IN ('CRI', 'PAN')",
+        bbox=(-85, 8, -80, 10),
     )
     assert len(df.columns) == 3
     assert len(df) == 2
-    assert df.iso_a3_renamed.tolist() == ["USA", "MEX"]
+    # records may be in different order
```
jorisvandenbossche, line 326: This comment can be removed?

jorisvandenbossche:
This looks good to me!

I was thinking that for one of the tests, we could keep the original bbox query, and then assert a different expected result depending on whether GDAL has GEOS available or not. But in the end, that's more testing something from GDAL, and that's maybe not worth asserting on our side that this is happening?

brendan-ward:
> But in the end, that's more testing something from GDAL, and that's maybe not worth asserting on our side that this is happening

That's what I was originally thinking, but looking further at this, we don't document this on our end (we should), so a test like you describe would be a good way of backing that statement up.

Also noticed that [docs for OGR_L_SetSpatialFilterRect](https://gdal.org/api/vector_c_api.html#_CPPv426OGR_L_SetSpatialFilterRect9OGRLayerHdddd) don't note variation if GEOS is available or not, but [OGR_G_Intersects(](https://gdal.org/api/vector_c_api.html#_CPPv416OGR_G_Intersects12OGRGeometryH12OGRGeometryH) does (presumably uses similar methods internally within GDAL, I didn't check).

docs/source/introduction.md
```diff
@@ -177,50 +177,57 @@ with the bbox.
 
 Note: the `bbox` values must be in the same CRS as the dataset.
 
+Note: if GEOS is present and used by GDAL, only geometries that intersect `bbox`
```
brendan-ward, line 180: This is the only substantive change in this file, everything else is auto-formatter stripping of trailing whitespace at end of lines.  Sorry for the noise here.

pyogrio/tests/test_core.py
```diff
@@ -138,26 +143,52 @@ def test_read_bounds_bbox(naturalearth_lowres_all_ext):
...
         [
-            [-171.791111, 18.916190, -66.964660, 71.357764],
-            [-117.127760, 14.538829, -86.811982, 32.720830],
+            [-82.96578305, 7.22054149, -77.24256649, 9.61161001],
+            [-85.94172543, 8.22502798, -82.54619626, 11.21711925],
         ],
     )
 
 
+def test_read_bounds_bbox_intersects_vs_envelope_overlaps(naturalearth_lowres_all_ext):
```
brendan-ward, line 166: Right now we are relying on the Release runners for macOS / Linux to test this; I did not add a specific environment to do so (and doing so may prove a bit challenging).

jorisvandenbossche: Relying on the release runners for now sounds good

jorisvandenbossche (approved):

PR #167, opened 2022-10-29
Benchmark fixes

TheBiggerGuy:
Small fixes to benchmarks:
* _read_ returns fids (as None) even when return_fids=False
* Duplicate dataset fixture in conftest
* Dataset fixture without file extension when using VSI


brendan-ward (approved):
Thanks @TheBiggerGuy !

PR #169, opened 2022-11-02
release CI: Always use latest minor version of cibuildwheel

EwoutH:
By removing the exact minor and patch version of the cibuildwheel action, it will always use the latest minor/patch version.

Major version updates will be handled by #168.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,7 +110,7 @@ jobs:
           BUILDKIT_PROGRESS: plain
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.10.2
+        uses: pypa/cibuildwheel@v2
 
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks for the PR! 
I am not sure why this didn't run any CI?

brendan-ward:
Not sure why this didn't run, changes to the `release.yml` file are specifically whitelisted for running CI:

```
  pull_request: # also build on PRs touching files that affect building sdist / wheels
    paths:
      - ".github/workflows/release.yml"
```

which I thought was working properly based on prior PRs...

jorisvandenbossche:
But also the normal tests should in principle _always_ run for any PR.

EwoutH:
My mistake, cibuildwheel (currently) doesn't allow this and requires pinned `major.minor.patch` versions specified. That's also probably why the CI isn't running. So closing this PR.

#168 still does it job, and #170 can thus be merged.

PR #172, opened 2022-11-02
windows CI: Always use latest patch version of setup-msbuild

EwoutH:
By removing the exact patch version of the microsoft/setup-msbuild action, it will always use the latest patch version. This is also how it's suggested in the [setup-msbuild](https://github.com/microsoft/setup-msbuild) docs.

This PR replaces #171.

Major and minor version updates will be handled by https://github.com/geopandas/pyogrio/pull/168.

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -52,7 +52,7 @@ jobs:
           key: ${{ runner.os }}-gdal-${{ matrix.GDAL_VERSION }}
 
       - name: Add msbuild to PATH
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
This file isn't actually being used (windows tests are covered by conda, this file has a manual `on: workflow_dispatch` so is in practice never run). @brendan-ward ok with removing this file?

brendan-ward:
Yes, OK with removing the file; this predates the vcpkg windows builds.

PR #173, opened 2022-11-02
CI: Remove unused windows-specific CI test action

brendan-ward:
This removes the unused Windows-specific CI action; Windows builds are now handled using vcpkg in the `release.yml` and Windows tests are performed using conda.

This makes a minor update to the installation documentation for Windows to make reference to vcpkg and remove the previous cross-reference to the CI action removed here.


martinfleis (approved):

docs/source/install.md
```diff
@@ -87,8 +87,18 @@ pytest pyogrio/tests
...
+GDAL and other dependencies using vcpkg, and then build Pyogrio from source.
+
+See `.github/workflows/release.yml` for details about how vcpkg is used as part
+of the wheel-building process.
+
+We do not yet have instructions on building Pyogrio from source using vcpkg for
+local development; please feel free to contribute additional documentation!
+
+Alternatively, you can install GDAL from an appropriate provider of Windows
+binaries. We've heard that the [OSGeo4W](https://trac.osgeo.org/osgeo4w/) works.
```
jorisvandenbossche, line 101: This was like this before as well, but my first go-to for having GDAL on windows (and build pyogrio from source) would be to use conda .. We mention conda above in the general installation section, but should we mention that here again for being explicit? When you just read the "Development" section for Windows, it might sound you need either vcpkg or something like OSGeo4W, while conda is still the easiest solution I would say.

brendan-ward: Good suggestion.  I assume you still need to set the environment variables when using conda, right?

jorisvandenbossche: Yes, with our current setuptools-based building, using an environment variable to point to GDAL is still needed on Windows

PR #194, opened 2023-01-05
DOC: Add examples of dataset & layer creation options

### LINKED ISSUE #193: DOC: add example of dataset / layer creation options to Introduction
Follow-on to #189 , we should demonstrate an example of setting dataset and layer creation options.

brendan-ward:
Resolves #193

This should only be merged after #189

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -392,6 +392,48 @@ You can also read from a URL with this syntax:
 >>> read_dataframe("zip+https://s3.amazonaws.com/bucket/shapefile.zip")
 ```
 
+## Dataset and layer creation options
+
+It is possible to use dataset and layer creation options available for a given
+driver in GDAL (see the relevant
+[GDAL driver page](https://gdal.org/drivers/vector/index.html)). These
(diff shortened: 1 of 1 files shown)


docs/source/introduction.md
```diff
@@ -392,6 +392,48 @@ You can also read from a URL with this syntax:
...
+
+```python
+>>> write_dataframe(df, "/tmp/test.shp", spatial_index=True)
+```
+
+You can use upper case to match the GDAL options exactly (creation options are
+converted to uppercase by default):
+
+```python
+>>> write_dataframe(df, '/tmp/test.shp', SPATIAL_INDEX=True)
```
jorisvandenbossche, line 423: Since you are showing the upper case to match GDAL exactly, also use "YES" since that is the exact value documented by GDAL?
```suggestion
>>> write_dataframe(df, '/tmp/test.shp', SPATIAL_INDEX="YES")
```

jorisvandenbossche (approved):
Looks good, thanks!

jorisvandenbossche (approved):

Issue #105, opened 2022-05-15
ENH: Support writing dataframes (without geometry)


theroggy:
Eg. a geopackage supports attribute tables: tables/layers without geometries.

OGR supports writing this type of tables by specifying geometry_type=None and the geometries being None.

Probably will be easier to implement once #59 is fixed.

PR #197, opened 2023-01-05
ENH: Add append support to write functions

### LINKED ISSUE #137: ENH: support appending to existing files while writing
Currently for writing, we only support the `mode="w"` pure write option (overwriting a file if it already exists). We should also add an append mode.

I think the main difference is that in the append case, the existing dataset gets opened (`GDALOpenEx`, used in `ogr_open`) instead of creating a new dataset (`GDALCreate`, used in `ogr_create`) to obtain a OGR Dataset pointer. 
The rest of the code might more or less work as is?

Fiona exposes this is as a `mode="w"|"a"` option (but also has `mode="r"`) in the same API, because it mimics the Python `open()` API. For pyogrio we don't necessarily have to follow that (eg we could also do a `append=True|False` which can be more accessible), although I see that also for example pandas' to_csv uses `mode="a"` to append to an existing file.

brendan-ward:
Resolves #137 

This adds basic support to append additional records to existing layers when supported by the driver.

This uses driver metadata plus a known list of drivers to exclude in order to determine which drivers support write capability (and updates listing from `list_drivers()`), and raises errors when attempting to write to unsupported drivers.

This adds a specific GDAL version test when appending to FlatGeobuf to avoid segfaults due to https://github.com/OSGeo/gdal/issues/5739; anything prior to GDAL 3.5.1 appears to segfault.

I could not find a driver metadata key that indicates if a driver supports append capability.  Rather than try to maintain a list of known drivers that support append, I think we probably just need to let GDAL raise the errors when a given driver / GDAL version does not support append.

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -340,8 +340,8 @@ overrule the default driver for an extension, you can specify the driver with th
 
 ## Appending to an existing data source
 
-Certain drivers may support the ability to append records to an existing data
-source. See the
+Certain drivers may support the ability to append records to an existing
+data layer in an existing data source. See the
 [GDAL driver listing](https://gdal.org/drivers/vector/index.html)
 for details about the capabilities of a driver for your version of GDAL.
 
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -24,10 +24,15 @@ Not all geometry or field types may be supported for all drivers.
 {...'GeoJSON': 'rw', 'GeoJSONSeq': 'rw',...}
 ```
 
-Drivers that are not known to be supported are listed with `"?"` for capabilities.
-Drivers that are known to support write capability end in `"w"`.
+Drivers that support write capability in your version of GDAL end in `"w"`.
+Certain drivers that are known to be unsupported in Pyogrio are disabled for
+write capabilities.
 
-To find subsets of drivers that have known support:
+NOTE: not all drivers support writing the contents of a GeoDataFrame; you may
+encounter errors due to unsupported data types, unsupported geometry types,
+or other driver-related errors when writing to a data source.
+
+To find subsets of drivers that support read or write capabilities:
 
 ```python
 >>> list_drivers(read=True)
@@ -38,8 +43,13 @@ See the full list of [drivers](https://gdal.org/drivers/vector/index.html) for
 more information about specific drivers, including their write support and
 configuration options.
 
-You can certainly try to read or write using unsupported drivers that are
-available in your installation, but you may encounter errors.
+The following drivers are known to be well-supported and tested in Pyogrio:
+
+-   `ESRI Shapefile`
+-   `FlatGeobuf`
+-   `GeoJSON`
+-   `GeoJSONSeq`
+-   `GPKG`
 
 ## List available layers
 
@@ -327,6 +337,23 @@ By default, the appropriate driver is inferred from the extension of the filenam
 If you want to write another file format supported by GDAL or if you want to
 overrule the default driver for an extension, you can specify the driver with the
 `driver` keyword, e.g. `driver="GPKG"`.
+
+## Appending to an existing data source
+
+Certain drivers may support the ability to append records to an existing data
+source. See the
+[GDAL driver listing](https://gdal.org/drivers/vector/index.html)
+for details about the capabilities of a driver for your version of GDAL.
+
+```
+>>> write_dataframe(df, "/tmp/existing_file.gpkg", append=True)
+```
+
+NOTE: the data structure of the data frame you are appending to the existing
+data source must exactly match the structure of the existing data source.
+
+NOTE: not all drivers that support write capabilities support append
+capabilities for a given GDAL version.
 
 ## Reading from compressed files / archives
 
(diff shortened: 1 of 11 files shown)


jorisvandenbossche:
With the other PRs merged, this has some conflicts now

pyogrio/_io.pyx
```diff
@@ -1361,42 +1360,50 @@ def ogr_write(str path, str layer, str driver, geometry, field_data, fields,
...
     if ogr_dataset == NULL:
         ogr_dataset = ogr_create(path_c, driver_c)
 
+    # if we are not appending to an existing layer, we need to create
+    # the layer and all associated properties (CRS, field defs, etc)
+    create_layer = not (append and layer_exists)
+
     ### Create the CRS
-    if crs is not None:
+    if create_layer and crs is not None:
```
jorisvandenbossche, line 1406: This can be inside the next `if create_layer:` block?

pyogrio/geopandas.py
```diff
@@ -254,6 +255,11 @@ def write_dataframe(
     layer_options : dict, optional
         Layer creation option (format specific) passed to OGR. Specify as
         a key-value dictionary.
+    append : bool, optional (default: False)
```
jorisvandenbossche, line 258: Something else: I like `append=True/False`, but note that geopandas / fiona uses `mode="w"/"a"` (but fiona has this because it mimics the `open(..)` API, and in geopandas we can easily map the keyword for compat; we currently even explicitly raise an error for `mode="a"` on the geopandas side in case of pyogrio engine)

pyogrio/raw.py
```diff
@@ -2,11 +2,18 @@
...
 from pyogrio._env import GDALEnv
+from pyogrio.errors import DataSourceError
 from pyogrio.util import get_vsi_path
 
 with GDALEnv():
     from pyogrio._io import ogr_read, ogr_read_arrow, ogr_write
-    from pyogrio._ogr import remove_virtual_file, _get_driver_metadata_item
+    from pyogrio._ogr import (
+        get_gdal_version,
+        get_gdal_version_string,
```
jorisvandenbossche, line 12:
```suggestion
        get_gdal_version,
```

(the latter is not used here?)

pyogrio/_ogr.pyx
```diff
@@ -100,24 +100,18 @@ def get_gdal_config_option(str name):
...
-    "GPKG": "rw",
-    "OAPIF": "r",
-    "OpenFileGDB": "r",
-    "TopoJSON": "r",
-    # "XLSX": "rw",  # TODO: needs geometry conversion method
-}
+def ogr_driver_supports_write(driver):
+    # exclude drivers known to be unsupported by pyogrio even though they are
+    # supported for write by GDAL
+    if driver in {"CSV", "XLSX"}:
```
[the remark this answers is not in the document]
jorisvandenbossche: Also wondering (but for a different issue/PR), do we want to support writing only the attributes without geometry (like we support reading with ignoring the geometry)

docs/source/introduction.md
```diff
@@ -328,6 +338,23 @@ If you want to write another file format supported by GDAL or if you want to
 overrule the default driver for an extension, you can specify the driver with the
 `driver` keyword, e.g. `driver="GPKG"`.
 
+## Appending to an existing data source
+
+Certain drivers may support the ability to append records to an existing data
+source. See the
```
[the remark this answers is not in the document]
brendan-ward: Yes, I probably should have said "append records to an existing layer in an existing data source".  Will update.

pyogrio/_ogr.pyx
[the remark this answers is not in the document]
brendan-ward: Ah - didn't know that; probably means that XLSX would work with that as well.  In which case, maybe we should just pass through those drivers that are supported for write, and let the user figure out what additional options may be required if GDAL returns an error.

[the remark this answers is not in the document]
brendan-ward: Yes: #105 addresses writing without geometry; was not trying to fast-track that for the upcoming release though.

pyogrio/geopandas.py
brendan-ward: I see that pandas supports the same modes as the `open()` API for things like `to_csv()`.  So we may want to be consistent but only support a subset: `write_dataframe` would never support `mode="r"`, for instance.

I'm not sure what the meaning of `mode="r+"` would be here, since we don't expose a stateful API, and we don't (yet) let you update existing records an existing data layer.

In which case, `read_dataframe` would only ever use `mode="r"` (so parameter is not needed), and `write_dataframe` could only ever use `mode="w"` or `mode="a"` - which ends up being an awful lot like our boolean `append`.

Would be nice not to change this API later, so I'm certainly open to using `mode` in order to make this more compatible with GeoPandas if that is what you'd like here.

jorisvandenbossche: Personally, for a direct user of pyogrio, I find `append=True` the nicer and easier API.

pyogrio/_ogr.pyx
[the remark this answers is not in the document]
brendan-ward: XLSX does not support geometry, so we need to keep this exclusion for XLSX until we do #105 and allow writing to it without geometry (though probably cleaner for user to do this through Pandas `to_excel(...)`)

pyogrio/raw.py
brendan-ward: Both are used here; using the string version on line 289 for the error message

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #203, opened 2023-01-14
FIX: Only use transactions when writing features if layer supports them

brendan-ward:
OpenFileGDB now supports write from GDAL 3.6.0 onwards, but does not support transactions.  We were always assuming transaction support was present because it was available for all drivers we specifically tested previously (though GDAL makes clear not all datasets / layers support transactions).

This adds a check for that capability and creates features without a transaction if layer doesn't support them, and adds a test to ensure that we can write OpenFileGDB correctly now.

### DIFF
--- a/pyogrio/tests/test_raw_io.py
+++ b/pyogrio/tests/test_raw_io.py
@@ -11,6 +11,9 @@ from pyogrio.raw import DRIVERS, read, write
 from pyogrio.errors import DataSourceError, DataLayerError, FeatureError
 from pyogrio.tests.conftest import prepare_testfile
 
+# mapping of driver name to extension
+DRIVER_EXT = {driver: ext for ext, driver in DRIVERS.items()}
+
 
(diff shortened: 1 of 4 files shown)


pyogrio/tests/test_raw_io.py
```diff
@@ -394,7 +410,7 @@ def test_write_supported(tmpdir, naturalearth_lowres, driver):
     # we take the first record only.
     meta["geometry_type"] = "MultiPolygon"
 
-    filename = tmpdir / "test"
+    filename = tmpdir / f"test{DRIVER_EXT[driver]}"
```
brendan-ward, line 413: This was added because GDAL raises errors for some drivers if the extension is not what is required; i.e., even though we don't test OpenFileGDB in this block, it must have a `.gdb` extension.  This is an attempt to future-proof adding additional drivers to `DRIVERS`.

jorisvandenbossche (approved):

Issue #205, opened 2023-01-18
Yield batches from ogr_read_arrow


kylebarron:
I'm playing around more with reading arrow tables from pyogrio and it's really exciting. It does feel like having some API to yield batches would be helpful to work with larger datasets. @jorisvandenbossche wrote in https://github.com/geopandas/pyogrio/pull/155:

> Returning a materialized Table (from `read_all()`) is fine for our current use case, but I can imagine that in the future, we might want to expose an iterative RecordBatchReader as well (eg as batch-wise input to query engine).
> When we want to do that, I assume that we somehow need to keep the GDAL Dataset alive (putting it in a Python object (wrapping in a small class, or putting in a PyCapsule with destructor), and keeping a reference to that object from the RecordBatchReader).

I've never touched ogr bindings before, but naively it seems the easiest way to do this is by using a context manager:
```py
with open_arrow("file.shp") as reader:
    for record_batch in reader:
		table
```
would that work? just putting a yield [here](https://github.com/geopandas/pyogrio/pull/155/files#diff-772414414b26a9df2e74063fd9e20d64cbc189d0e7e5671613c3bcc39faf0e47R1071-R1081)?

brendan-ward:
Do you envision that there would be a counterpart function to write data in batches, via the Arrow I/O API (once available)?

kylebarron:
That seems entirely dependent on GDAL? https://gdal.org/development/rfc/rfc86_column_oriented_api.html says

> Potential future work, not in the scope of this RFC, could be the addition of a column-oriented method to write new features, a WriteRecordBatch() method.

I would of course love for that to be added to GDAL, and getting greater adoption of RFC 86 seems very helpful for that.

https://github.com/geopandas/pyogrio/pull/206 appeared to work on my local machine 🤷‍♂️. If a maintainer could enable CI on that PR that would be helpful!

kylebarron:
If GDAL were to add it, I think a similar API like

```py
with write_ogr_batches("file.gpkg", arrow_schema) as writer:
    writer.write_batch(batch)
```

could make sense. But for my own needs I think I'm more likely to only _write_ to geoparquet and thus not use OGR as much for writing

jorisvandenbossche:
@kylebarron thanks for looking into this! That seems like a really nice idea, I didn't think about a contextmanager to keep the dataset alive.

jorisvandenbossche:
This was closed by #206

PR #206, opened 2023-01-18
Streaming Arrow reader

kylebarron:
A quick 10-minute hack to test out my idea in https://github.com/geopandas/pyogrio/issues/205, but it appears to be working!

```py
from pyogrio.raw import open_arrow
import shapely

# dataset of buildings in utah from
# https://github.com/microsoft/USBuildingFootprints
path = '/Users/kyle/Downloads/Utah.geojson'

centroids = []
with open_arrow(path) as file:
    meta, reader = file
    i = 0
    for table in reader:
        if i == 10:
            break

        geometries = shapely.from_wkb(table['wkb_geometry'])
        centroids.append(shapely.centroid(geometries))
        i += 1

centroids[0][:10]
```
```
array([<POINT (-109.04 37.633)>, <POINT (-109.044 38.108)>,
       <POINT (-109.051 37.376)>, <POINT (-109.057 37.494)>,
       <POINT (-109.058 40.588)>, <POINT (-109.062 37.311)>,
       <POINT (-109.065 37.861)>, <POINT (-109.066 37.891)>,
       <POINT (-109.07 37.482)>, <POINT (-109.073 37.343)>], dtype=object)
```

This is just a POC to test that it's working. I think it would be a slightly cleaner API to wrap the `RecordBatchStreamReader` and metadata into one class so you could do
```py
with open_arrow(path) as reader:
    geometry_column = reader.geometry_column
    for table in reader:
    geometries = shapely.from_wkb(table[geometry_column])
```

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -4,6 +4,7 @@
 """
 
 
+import contextlib
 import datetime
 import locale
 import logging
@@ -1111,6 +1112,142 @@ def ogr_read_arrow(
 
     return meta, table
 
+
+@contextlib.contextmanager
+def ogr_open_arrow(
+    str path,
+    object layer=None,
+    object encoding=None,
+    int read_geometry=True,
+    int force_2d=False,
+    object columns=None,
+    int skip_features=0,
+    int max_features=0,
+    object where=None,
+    tuple bbox=None,
+    object fids=None,
+    str sql=None,
+    str sql_dialect=None,
+    int return_fids=False,
+    **kwargs):
+
+    cdef int err = 0
+    cdef const char *path_c = NULL
+    cdef const char *where_c = NULL
+    cdef OGRDataSourceH ogr_dataset = NULL
+    cdef OGRLayerH ogr_layer = NULL
+    cdef char **fields_c = NULL
+    cdef const char *field_c = NULL
+    cdef char **options = NULL
+    cdef ArrowArrayStream stream
+    cdef ArrowSchema schema
+
+    path_b = path.encode('utf-8')
+    path_c = path_b
+
+    if force_2d:
+        raise ValueError("forcing 2D is not supported for Arrow")
+
+    if fids is not None:
+        raise ValueError("reading by FID is not supported for Arrow")
+
+    if skip_features or max_features:
+        raise ValueError(
+            "specifying 'skip_features' or 'max_features' is not supported for Arrow"
+        )
+
+    if sql is not None and layer is not None:
+        raise ValueError("'sql' paramater cannot be combined with 'layer'")
+
+    ogr_dataset = ogr_open(path_c, 0, kwargs)
+    try:
+        if sql is None:
+            # layer defaults to index 0
+            if layer is None:
+                layer = 0
+            ogr_layer = get_ogr_layer(ogr_dataset, layer)
+        else:
+            ogr_layer = execute_sql(ogr_dataset, sql, sql_dialect)
+
+        crs = get_crs(ogr_layer)
+
+        # Encoding is derived from the user, from the dataset capabilities / type,
+        # or from the system locale
+        encoding = (
+            encoding
+            or detect_encoding(ogr_dataset, ogr_layer)
+            or locale.getpreferredencoding()
+        )
+
+        fields = get_fields(ogr_layer, encoding, use_arrow=True)
+
+        ignored_fields = []
+        if columns is not None:
+            # Fields are matched exactly by name, duplicates are dropped.
+            ignored_fields = list(set(fields[:,2]) - set(columns))
+        if not read_geometry:
+            ignored_fields.append("OGR_GEOMETRY")
+
+        geometry_type = get_geometry_type(ogr_layer)
+
+        geometry_name = get_string(OGR_L_GetGeometryColumn(ogr_layer))
+
+        # Apply the attribute filter
+        if where is not None and where != "":
+            apply_where_filter(ogr_layer, where)
+
+        # Apply the spatial filter
+        if bbox is not None:
+            apply_spatial_filter(ogr_layer, bbox)
+
+        # Limit to specified columns
+        if ignored_fields:
+            for field in ignored_fields:
+                field_b = field.encode("utf-8")
+                field_c = field_b
+                fields_c = CSLAddString(fields_c, field_c)
+
+            OGR_L_SetIgnoredFields(ogr_layer, <const char**>fields_c)
+
+        if not return_fids:
+            options = CSLSetNameValue(options, "INCLUDE_FID", "NO")
+
+        # make sure layer is read from beginning
+        OGR_L_ResetReading(ogr_layer)
+
+        IF CTE_GDAL_VERSION < (3, 6, 0):
+            raise RuntimeError("Need GDAL>=3.6 for Arrow support")
+
+        if not OGR_L_GetArrowStream(ogr_layer, &stream, options):
+            raise RuntimeError("Failed to open ArrowArrayStream from Layer")
+
+        stream_ptr = <uintptr_t> &stream
+
+        # stream has to be consumed before the Dataset is closed
+        import pyarrow as pa
+        reader = pa.RecordBatchStreamReader._import_from_c(stream_ptr)
+
+        meta = {
+            'crs': crs,
+            'encoding': encoding,
+            'fields': fields[:,2], # return only names
+            'geometry_type': geometry_type,
+            'geometry_name': geometry_name,
+        }
+
+        yield meta, reader
+
+    finally:
+        CSLDestroy(options)
+        if fields_c != NULL:
+            CSLDestroy(fields_c)
+            fields_c = NULL
+        if ogr_dataset != NULL:
+            if sql is not None:
+                GDALDatasetReleaseResultSet(ogr_dataset, ogr_layer)
+
+            GDALClose(ogr_dataset)
+            ogr_dataset = NULL
 
 def ogr_read_bounds(
     str path,
--- a/pyogrio/raw.py
+++ b/pyogrio/raw.py
@@ -1,3 +1,4 @@
+import contextlib
 import warnings
 import os
 
@@ -6,7 +7,7 @@ from pyogrio.errors import DataSourceError
 from pyogrio.util import get_vsi_path
 
 with GDALEnv():
-    from pyogrio._io import ogr_read, ogr_read_arrow, ogr_write
+    from pyogrio._io import ogr_open_arrow, ogr_read, ogr_read_arrow, ogr_write
     from pyogrio._ogr import (
         get_gdal_version,
         get_gdal_version_string,
@@ -198,6 +199,59 @@ def read_arrow(
             remove_virtual_file(path)
 
     return result
+
+
+@contextlib.contextmanager
+def open_arrow(
+    path_or_buffer,
+    /,
+    layer=None,
+    encoding=None,
+    columns=None,
+    read_geometry=True,
+    force_2d=False,
+    skip_features=0,
+    max_features=None,
+    where=None,
+    bbox=None,
+    fids=None,
+    sql=None,
+    sql_dialect=None,
+    return_fids=False,
+):
+    """
+    Read OGR data source into a pyarrow Table.
+
+    See docstring of `read` for details.
+    """
+    try:
+        import pyarrow  # noqa
+    except ImportError:
+        raise RuntimeError("the 'pyarrow' package is required to read using arrow")
+
+    path, buffer = get_vsi_path(path_or_buffer)
+
+    try:
+        with ogr_open_arrow(
+            path,
+            layer=layer,
+            encoding=encoding,
+            columns=columns,
+            read_geometry=read_geometry,
+            force_2d=force_2d,
+            skip_features=skip_features,
+            max_features=max_features or 0,
+            where=where,
+            bbox=bbox,
+            fids=fids,
+            sql=sql,
+            sql_dialect=sql_dialect,
+            return_fids=return_fids,
+        ) as result:
+            yield result
+    finally:
+        if buffer is not None:
+            remove_virtual_file(path)
 
 
 def detect_driver(path):


jorisvandenbossche:
I suppose one problem with this is that a user is not forced to use the reader within the context. Assume someone would do:

```
with open_arrow(path) as file:
    meta, reader = file

# further use reader after dataset is closed
reader.read_all()
```

Or we should make the `reader` object "unusable" outside the context in `finally` (maybe just `del` it? Or would that give unexpected python semantics)

jorisvandenbossche:
> I think it would be a slightly cleaner API to wrap the RecordBatchStreamReader and metadata into one class so you could do

That would also solve the issue I mentioned above, because then we could link the lifetime of the GDAL dataset to this custom object

kylebarron:
> I suppose one problem with this is that a user is not forced to use the reader within the context.

Yes indeed, as of now I just wrote a strongly worded statement in the docstring 😄 

>     The RecordBatchStreamReader is reading from a stream provided by OGR and must not be
>     accessed after the OGR dataset has been closed, i.e. after the context manager has
>     been closed.

..

> That would also solve the issue I mentioned above, because then we could link the lifetime of the GDAL dataset to this custom object

I agree that makes sense.

kylebarron:
With the last commit I think the lifetime issue is solved, just by closing the record batch reader in the `finally` block... I got a segmentation fault trying to use the tab completion on the `reader` object in `ipython` after the context 😄 

```
In [4]: with open_arrow(path) as file:
   ...:     meta, reader = file
   ...:

In [5]: reader
Out[5]: <pyarrow.lib.RecordBatchReader at 0x123619b00>

In [6]: reader.[1]    52643 segmentation fault  ./env/bin/ipython
```

pyogrio/arrow.py
```diff
@@ -0,0 +1,13 @@
...
+class OGRRecordBatchStreamReader(pa.RecordBatchStreamReader):
+    _geo_metadata: dict
+
+    @property
+    def geo_metadata(self):
+        return self._geo_metadata
+
+    @geo_metadata.setter
+    def geo_metadata(self, value):
+        self._geo_metadata = value
```
kylebarron, line 13: Any comments on this general structure? Should this work?

kylebarron: @jorisvandenbossche do you know if there's an issue with subclassing `pa.RecordBatchStreamReader` like this?

jorisvandenbossche: Sorry for the slow response here! Will take a closer look tomorrow. Is there a specific issue you are seeing with subclassing, or just wondering in general if this is OK to do?

jorisvandenbossche:
> I got a segmentation fault trying to use the tab completion on the `reader` object in `ipython` after the context smile

That's probably an issue in pyarrow that we should look into

jorisvandenbossche:
I think I would personally also be fine with the pattern of 

```
with open_arrow(path) as source:
    meta, reader = source
    ...
```

that you had initially, which would avoid the subclass. Of course, if you don't need to metadata, the one that directly returns a reader object is less verbose to use.

pyogrio/arrow.py
kylebarron: I reverted the wrapper for subclassing

kylebarron:
> Of course, if you don't need to metadata, the one that directly returns a reader object is less verbose to use.

Either way. I tend to prefer dataclasses over returning a tuple as a response, because I like the named access, but no strong preference (and the pyogrio api is already returning tuples elsewhere)

jorisvandenbossche:
> > I got a segmentation fault trying to use the tab completion on the `reader` object in `ipython` after the context smile
> 
> That's probably an issue in pyarrow that we should look into

I could reproduce some segfault with just pyarrow: https://github.com/apache/arrow/issues/34906. Not sure it's exactly the same, but it might be (tab completion in my case didn't segfault, but in general that might trigger evaluating some of the attributes)

pyogrio/raw.py
```diff
@@ -191,7 +206,93 @@ def read_arrow(
...
+    read_geometry=True,
+    force_2d=False,
+    skip_features=0,
+    max_features=None,
+    where=None,
+    bbox=None,
+    fids=None,
+    sql=None,
+    sql_dialect=None,
+    return_fids=False,
```
jorisvandenbossche, line 249:
```suggestion
    return_fids=False,
    **kwargs,
```

Some fixes are needed to update for a recently merged PR: we need to accept kwargs here, and then convert that to `dataset_kwargs` variable

pyogrio/raw.py
```diff
@@ -191,7 +206,93 @@ def read_arrow(
     dataset_kwargs = _preprocess_options_key_value(kwargs) if kwargs else {}
 
     try:
-        result = ogr_read_arrow(
+        with ogr_open_arrow(
```
jorisvandenbossche, line 209: Could we use the `open_arrow` (instead of the lower level `ogr_open_arrow`)  here in `read_arrow` ? Then the generic boilerplate (checking for pyarrow being installed, processing the path, the dataset kwargs etc) only needs to be written once.

jorisvandenbossche:
> I tend to prefer dataclasses over returning a tuple as a response, because I like the named access, but no strong preference (and the pyogrio api is already returning tuples elsewhere)

Or a namedtuple? Then both attribute access as tuple unpacking should work?

pyogrio/raw.py
kylebarron: I think I fixed this correctly

kylebarron:
Do you know the right way to fix this cython compilation?
```
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      
              yield meta, reader
      
          finally:
              # Mark reader as closed to prevent reading batches
              reader.close()
             ^
      ------------------------------------------------------------
```

I thought this worked locally for me before. I'm guessing we need a top-level `cdef`?

jorisvandenbossche:
> Edit: why is it only failing on GDAL pre 3.6?

Ah, this object is only defined then, because of the compile-time `IF CTE_GDAL_VERSION < (3, 6, 0):`. We probably need to do a similar if check in the finally clause

jorisvandenbossche:
@kylebarron thanks for the update, looking good!

Are you happy with the API? (I am going back and forth a bit on the question of just yielding a reader with metadata attached vs yielding the tuple. Just the reader feels indeed nicer, but given that this is a more advanced API and it has a good example in the docstring, it's probably just fine as it is now)

One last thing: can you add a small note in the changelog? (CHANGES.md) 
And maybe an explicit test for `open_arrow` (although it is used by `read_arrow` under the hood) would be good as well (just a dummy one like `test_read_arrow_raw` asserting the API is fine)

kylebarron:
> Are you happy with the API?

I don't have a strong preference, given that it's a more advanced API

pyogrio/_io.pyx
```diff
@@ -1142,7 +1143,7 @@ def ogr_read_arrow(
 
         # stream has to be consumed before the Dataset is closed
         import pyarrow as pa
-        table = pa.RecordBatchStreamReader._import_from_c(stream_ptr).read_all()
+        reader = pa.RecordBatchStreamReader._import_from_c(stream_ptr)
```
brendan-ward, line 1146: I think you need to define `reader = None` before the `try` block, and then check against that before you close it:

```
if reader is not None:
    reader.close()
```


Otherwise, we get nested errors like this:

```
>>> t = read_arrow('foo')
  File "pyogrio/_io.pyx", line 150, in pyogrio._io.ogr_open
    ogr_dataset = exc_wrap_pointer(
  File "pyogrio/_err.pyx", line 178, in pyogrio._err.exc_wrap_pointer
    raise exc
pyogrio._err.CPLE_OpenFailedError: foo: No such file or directory

During handling of the above exception, another exception occurred:

  File "pyogrio/_io.pyx", line 160, in pyogrio._io.ogr_open
    raise DataSourceError(str(exc))
pyogrio.errors.DataSourceError: foo: No such file or directory

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/bcward/projects/pyogrio/pyogrio/raw.py", line 199, in read_arrow
    with open_arrow(
  File "/Users/bcward/.pyenv/versions/3.10.5/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "pyogrio/_io.pyx", line 1162, in ogr_open_arrow
    reader.close()
UnboundLocalError: local variable 'reader' referenced before assignment
```

brendan-ward (commented):
Thanks for working on this @kylebarron and apologies for the long-delayed review!

In terms of API, my preference is to stick to using a tuple for now instead of a class; it is more consistent with what we're doing for other lower-level operations elsewhere.

Longer term, we can move these to `namedtuple` for easier access / description of the return values.

Can you please add tests, including one to verify that the reader is closed / inoperable after the end of a `with` block?

kylebarron:
> verify that the reader is closed / inoperable after the end of a `with` block?

@jorisvandenbossche might have feedback here, but it's not clear to me if there's any way to tell from pyarrow if a `RecordBatchReader` object has already been closed. Often on file-like objects there's a `closed` attribute, but the `RecordBatchReader` [doesn't appear to have that](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatchReader.html), but only a `close()` method.

kylebarron:
I also added a `batch_size` parameter that sets `MAX_FEATURES_IN_BATCH` (as described by the doc [here](https://gdal.org/api/vector_c_api.html#_CPPv420OGR_L_GetArrowStream9OGRLayerHP16ArrowArrayStreamPPc))

The current default in OGR is 65,536. I learned you can't have `None` as a default value to a Cython parameter that's defined as `int`. So here I defined the default as `0` and only set the option when `> 0`. Or I can set the default as `65536` if you prefer

brendan-ward:
> verify that the reader is closed / inoperable after the end of a with block?

I was more referring to testing that `reader.close()` was called but I didn't give thought as to how to verify it; the point was to make sure we _don't_ segfault.  If we can't detect that the reader is already closed without segfaulting, I guess there isn't much we can test other than by hand.

kylebarron:
> I think in this case it makes sense to be transparent about the default value, assuming it is unlikely to change on the GDAL side.

👍 I set the default value to 65536

kylebarron:
> If we can't detect that the reader is already closed without segfaulting, I guess there isn't much we can test other than by hand

Given the current pyarrow API, I don't see an obvious way to test this, but I might be missing something. Maybe it would make sense to add to Pyarrow a `closed` boolean attribute here https://github.com/apache/arrow/blob/f0c8229f5a09fe53186df171d518430243ddf112/python/pyarrow/ipc.pxi#L755-L761

pyogrio/_io.pyx
```diff
@@ -1156,8 +1165,11 @@ def ogr_open_arrow(
         yield meta, reader
 
     finally:
-        # Mark reader as closed to prevent reading batches
-        reader.close()
+        # We need to use this compile-time clause again to close the reader
+        IF CTE_GDAL_VERSION >= (3, 6, 0):
```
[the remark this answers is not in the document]
kylebarron: Removed

jorisvandenbossche:
Yeah, I am also not familiar with a way to check that it was properly closed. You could do another call to "read" and that should give an error. But that's only fixed in upcoming pyarrow release, so that's not something we can already test right now.

> Maybe it would make sense to add to Pyarrow a `closed` boolean attribute here https://github.com/apache/arrow/blob/f0c8229f5a09fe53186df171d518430243ddf112/python/pyarrow/ipc.pxi#L755-L761

I think the problem might be that not all RecordbatchReader implementations really have this concept. For example, the simplest version is just backed by an in-memory vector of record batches, and for that one the `close()` method is just a dummy no-op. Of course, those could keep track of a "closed" status as well (although they don't really need it), but so that would require some changes either on the C++ side, or do this only in the cython class.

jorisvandenbossche:
One more general comment: should we maybe expose those `read_arrow`/`open_arrow` methods in the top-level namespace instead of only from `pyogrio.raw`? Compared to the `pyogrio.raw.read`, they are quite a bit more high-level, I would say.

kylebarron:
> One more general comment: should we maybe expose those `read_arrow`/`open_arrow` methods in the top-level namespace instead of only from `pyogrio.raw`?

I don't have a strong preference there. I don't know how many people will be using them; at least for now the Arrow geo ecosystem is small. But right now `read_arrow` isn't in the public docs at all; at least adding them to the docs page would add visibility

brendan-ward:
> should we maybe expose those read_arrow/open_arrow methods in the top-level namespace instead of only from pyogrio.raw?

I also don't have a strong opinion, but since we don't decode the geometries from WKB, it seems a bit more _raw_ than the others.  I'm also not opposed to promoting `read` to the top-level namespace.

pyogrio/tests/test_arrow.py
```diff
@@ -74,3 +76,24 @@ def test_read_arrow_raw(naturalearth_lowres):
     meta, table = read_arrow(naturalearth_lowres)
     assert isinstance(meta, dict)
     assert isinstance(table, pyarrow.Table)
+
+
+def test_open_arrow_raw(naturalearth_lowres):
```
brendan-ward, line 81:
```suggestion
def test_open_arrow(naturalearth_lowres):
```

pyogrio/tests/test_arrow.py
```diff
@@ -74,3 +76,24 @@ def test_read_arrow_raw(naturalearth_lowres):
...
+    with open_arrow(naturalearth_lowres) as (meta, reader):
+        assert isinstance(meta, dict)
+        assert isinstance(reader, pyarrow.RecordBatchReader)
+        assert isinstance(reader.read_all(), pyarrow.Table)
+
+
+def test_open_arrow_raw_batch_size(naturalearth_lowres):
```
brendan-ward, line 88:
```suggestion
def test_open_arrow_batch_size(naturalearth_lowres):
```

pyogrio/tests/test_arrow.py
```diff
@@ -74,3 +76,24 @@ def test_read_arrow_raw(naturalearth_lowres):
...
+    meta, table = read_arrow(naturalearth_lowres)
+    batch_size = math.ceil(len(table) / 2)
+
+    with open_arrow(naturalearth_lowres, batch_size=batch_size) as (meta, reader):
+        assert isinstance(meta, dict)
+        assert isinstance(reader, pyarrow.RecordBatchReader)
+        count = 0
+        for table in reader:
+            count += 1
```
brendan-ward, line 97: You could also verify that the size of the first table is equal to `batch_size`

brendan-ward (commented):
Thanks @kylebarron !  A couple minor suggestions re: test names, and a merge needed from `main`, but otherwise this looks good to go!

PR #223, opened 2023-03-09
ENH: add automatic detection of 2.5D in write_dataframe

### LINKED ISSUE #121: Bug: write_dataframe does not properly handle Z geometry types
For example:

```Python
df = geopandas.GeoDataFrame(geometry=[pygeos.points([0,0,0])])
write_dataframe(df, '/tmp/test.fgb')
```

fails because FGB is picky about mismatched geometry types and our geometry type inference resolves this to `Point` instead of `Point Z`, but the WKB data passed to GDAL contain Z values.  This is because the GeoSeries reports this as `Point`:

```Python
>>> df.geometry.type.unique()
array(['Point'], dtype=object)
```

Which happens because pygeos reports this as Point type (Z values are not given specific type codes):

```Python
>>> pygeos.get_type_id(df.geometry.values.data)
array([0], dtype=int32)
```
([`POINT is 0`](https://pygeos.readthedocs.io/en/latest/geometry.html#pygeos.geometry.get_type_id))

Shapely does the same:

```Python
>>> df.geometry.values[0].type
'Point'
```

I think the upshot is that we also have to intersect the geometry types we obtain from the GeoSeries with `" Z"` if the geometry has Z values (@jorisvandenbossche is this something that should be handled in GeoPandas instead, so that the `df.geometry.type` reports `Point Z` instead of `Point`?)

Specifying the geometry type directly:

```Python
write_dataframe(geopandas.GeoDataFrame(geometry=[pygeos.points([0,0,0])]), '/tmp/test.fgb', layer_geometry_type="Point Z")
```

fails, because we don't have `Point Z` in `_geometry.pyx::GEOMETRY_TYPE_CODES`; we only have the 2.5D variants (which is technically how GDAL would store them, but we should consider using "<Geometry> Z" aliases to them because I think that all GEOS geometry Z types are equivalent to GDAL 2.5D geometry types).

### LINKED ISSUE #222: ENH: add automatic detection of 2.5D in write_dataframe
At the moment write_dataframe doesn't take in account whether the input is 3D or not while determining the geometry_type.

The has_z property should be used to determine this...

Noticed in the context of this PR: [geopandas #2829](https://github.com/geopandas/geopandas/pull/2829)

theroggy:
closes #222

apparently closes #121 as well

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -309,9 +309,11 @@ def write_dataframe(
     # Determine geometry_type and/or promote_to_multi
     if geometry_type is None or promote_to_multi is None:
         tmp_geometry_type = "Unknown"
+        has_z = False
 
         # If there is data, infer layer geometry type + promote_to_multi
         if not df.empty:
+            has_z = geometry.has_z.any()
             geometry_types = pd.Series(geometry.type.unique()).dropna().values
             if len(geometry_types) == 1:
                 tmp_geometry_type = geometry_types[0]
@@ -347,6 +349,8 @@ def write_dataframe(
 
         if geometry_type is None:
             geometry_type = tmp_geometry_type
+            if has_z:
+                geometry_type = f"2.5D {geometry_type}"
 
     crs = None
     if geometry.crs:
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -313,7 +313,8 @@ def write_dataframe(
 
         # If there is data, infer layer geometry type + promote_to_multi
         if not df.empty:
-            has_z = geometry.has_z.any()
+            # Empty geometries sometimes report as Z incorrectly, so ignore them
+            has_z = geometry[~geometry.is_empty].has_z.any()
             geometry_types = pd.Series(geometry.type.unique()).dropna().values
             if len(geometry_types) == 1:
                 tmp_geometry_type = geometry_types[0]
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -1,5 +1,7 @@
-from pyogrio.raw import DRIVERS_NO_MIXED_SINGLE_MULTI
+import numpy as np
+from pyogrio.raw import DRIVERS_NO_MIXED_SINGLE_MULTI, DRIVERS_NO_MIXED_DIMENSIONS
 from pyogrio.raw import detect_driver, read, read_arrow, write
+from pyogrio.errors import DataSourceError
 
 
 def _stringify_path(path):
@@ -313,8 +315,18 @@ def write_dataframe(
 
         # If there is data, infer layer geometry type + promote_to_multi
         if not df.empty:
-            # Empty geometries sometimes report as Z incorrectly, so ignore them
-            has_z = geometry[~geometry.is_empty].has_z.any()
+            # None/Empty geometries sometimes report as Z incorrectly, so ignore them
+            has_z_arr = geometry[
+                (geometry != np.array(None)) & (~geometry.is_empty)
+            ].has_z
+            has_z = has_z_arr.any()
+            all_z = has_z_arr.all()
+
+            if driver in DRIVERS_NO_MIXED_DIMENSIONS and has_z and not all_z:
+                raise DataSourceError(
+                    f"Mixed 2D and 3D coordinates are not supported by {driver}"
+                )
+
             geometry_types = pd.Series(geometry.type.unique()).dropna().values
             if len(geometry_types) == 1:
                 tmp_geometry_type = geometry_types[0]
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -15,6 +15,7 @@ try:
     from pandas.testing import assert_frame_equal, assert_index_equal
 
     import geopandas as gp
+    from geopandas.array import from_wkt
     from geopandas.testing import assert_geodataframe_equal
 
     from shapely.geometry import Point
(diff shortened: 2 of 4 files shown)


pyogrio/geopandas.py
```diff
@@ -309,9 +309,11 @@ def write_dataframe(
     # Determine geometry_type and/or promote_to_multi
     if geometry_type is None or promote_to_multi is None:
         tmp_geometry_type = "Unknown"
+        has_z = False
 
         # If there is data, infer layer geometry type + promote_to_multi
         if not df.empty:
+            has_z = geometry.has_z.any()
```
jorisvandenbossche, line 316: We might want to filter out empties before calling `has_z` (something like `geometry[~geometry.is_empty].has_z.any()`). Depending on how the empty geometry is created, they tend to incorrectly think they are 3D (also depending on the GEOS version)

(I did something similar in shapely where we needed to determine to take a 2D vs 3D code path: https://github.com/shapely/shapely/blob/1ee6e96a4c84d1b99152b16ac96b71d2b8012c97/shapely/_ragged_array.py#L249-L252)

pyogrio/geopandas.py
```diff
@@ -347,6 +349,8 @@ def write_dataframe(
 
         if geometry_type is None:
             geometry_type = tmp_geometry_type
+            if has_z:
+                geometry_type = f"2.5D {geometry_type}"
```
jorisvandenbossche, line 353: https://github.com/geopandas/pyogrio/commit/8679f3e2a07b62940b8bd9f55d1b4c690acc76ef added the versions with " Z" suffix as well? Do we want to use those ourselves as well? (I personally find that easier to read than this strange "2.5D", but of course that's what GDAL itself uses it)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -843,17 +844,56 @@ def test_write_read_null(tmp_path):
...
+)
+def test_write_geometry_z_types_auto(tmp_path, ext, wkt, exp_geometry_type):
+    # Shapefile has some different behaviour that other file types
+    if ext == ".shp":
+        if exp_geometry_type == "2.5D GeometryCollection":
+            pytest.skip(f"ext {ext} doesn't support {exp_geometry_type}")
+        elif exp_geometry_type == "2.5D MultiLineString":
+            exp_geometry_type = "2.5D LineString"
+        elif exp_geometry_type == "2.5D MultiPolygon":
+            exp_geometry_type = "2.5D Polygon"
```
jorisvandenbossche, line 882: But so it still writes correctly? (and reads as 3D multi geometries?) 
If so, this might also be a bug in GDAL?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -843,17 +844,56 @@ def test_write_read_null(tmp_path):
...
+
+    filename = tmp_path / f"test{ext}"
+    gdf = gp.GeoDataFrame(geometry=from_wkt([wkt]), crs="EPSG:4326")
+    write_dataframe(gdf, filename)
+    info = read_info(filename)
+    assert info["geometry_type"] == exp_geometry_type
+    df = read_dataframe(filename)
+    if ext == ".shp":
+        df = df.drop(columns="FID")
```
jorisvandenbossche, line 891: Do you know why this is needed here? (I see other tests using `ALL_EXTS` with roundtripts and `assert_geodataframe_equal` that don't have this)

jorisvandenbossche (commented):
Thanks for looking into this! Couple of quick questions

pyogrio/tests/test_geopandas_io.py
theroggy: Yes, the data is saved correctly. Shapefile doesn't make a distinction between multi and single at file/layer metadata level... so gdal always report single. This is why it is also not needed to promote to multi to save mixed single and multi geometries in shapefile.

pyogrio/geopandas.py
theroggy: I used this notation because this is also the way `pyogrio.read_info` will report it... so at first sight it seems less confusing to me to stick to it... but...

pyogrio/tests/test_geopandas_io.py
theroggy: If I recall correctly it is because the dataframe only has a geometry column, and if there are no columns the shapefile driver creates the fid column.

pyogrio/geopandas.py
brendan-ward: We could also change what is presented on read to make them consistent, and stick to GEOS-style "Z" suffixes throughout.  Perhaps to keep this moving ahead, stick with `2.5D` for consistency, then we revisit this comprehensively later?

pyogrio/geopandas.py
theroggy: Done

pyogrio/geopandas.py
theroggy: Making the "Z" syntax consistent is in my opinion also a good option. It's a breaking change, but few people will notice. 

In fiona they apparently use a third variant for the same thing: e.g. "3D POINT": https://fiona.readthedocs.io/en/latest/manual.html#geometry-types

But, it's only internal here, so indeed not "urgent".

pyogrio/geopandas.py
```diff
@@ -313,7 +313,8 @@ def write_dataframe(
 
         # If there is data, infer layer geometry type + promote_to_multi
         if not df.empty:
-            has_z = geometry.has_z.any()
+            # Empty geometries sometimes report as Z incorrectly, so ignore them
+            has_z = geometry[~geometry.is_empty].has_z.any()
```
brendan-ward, line 317: Since the error from GDAL for FlatGeobuf on mixed dimensions is not very easy to interpret, I wonder if we can't catch it sooner - at least for the GeoDataFrame case?

Perhaps something like:

```Python
ix = geometry[~geometry.is_empty].has_z
has_z = ix.any()

if driver == "FlatGeobuf" and has_z and not ix.all():
    raise DataSourceError("FlatGeobuf does not support mixed 2D and 3D coordinates")
```

This may also be a place to generalize via a `DRIVERS_NO_MIXED_DIMENSIONS` similar to the way `DRIVERS_NO_MIXED_SINGLE_MULTI` is used.

brendan-ward (commented):
Thanks for working on this @theroggy .  Per the comment, I wonder if we can more deliberately catch the GDAL error for FlatGeobuf to make a nicer error message.

pyogrio/geopandas.py
theroggy: I added a cleaner error... but fgb seems to be a can of worms: a new problem arose: #226

pyogrio/tests/test_geopandas_io.py
jorisvandenbossche: Ah, yes, now I remember running into something similar. I see you already added a column so the special case is no longer needed then

pyogrio/tests/test_geopandas_io.py
```diff
@@ -843,17 +848,89 @@ def test_write_read_null(tmp_path):
...
+            exp_geometry_type = "2.5D LineString"
+        elif exp_geometry_type == "2.5D MultiPolygon":
+            exp_geometry_type = "2.5D Polygon"
+
+    column_data = {}
+    column_data["test_descr"] = [test_descr] * len(wkt)
+    column_data["idx"] = [str(idx) for idx in range(len(wkt))]
+    gdf = gp.GeoDataFrame(column_data, geometry=from_wkt(wkt), crs="EPSG:4326")
+    filename = tmp_path / f"test{ext}"
+    if mixed_dimensions and DRIVERS[ext] in DRIVERS_NO_MIXED_DIMENSIONS:
```
brendan-ward, line 913: minor nit: please add a newline before / after most of the conditional blocks; they're a bit dense here.  E.g., above lines 913, 922, 925

pyogrio/tests/test_geopandas_io.py
```diff
@@ -843,17 +848,89 @@ def test_write_read_null(tmp_path):
...
+        return
+    else:
+        write_dataframe(gdf, filename)
+    info = read_info(filename)
+    assert info["geometry_type"] == exp_geometry_type
+    result_gdf = read_dataframe(filename)
+    if ext == ".geojsonl":
+        result_gdf.crs = "EPSG:4326"
+    if ext == ".fgb":
+        empty = gdf.geometry[(gdf.geometry == np.array(None)) | (gdf.geometry.is_empty)]
```
brendan-ward, line 928: Another option would be to create an `expected` gdf to compare to, something like (untested)

```Python
if ext == ".fgb":
    # FGDB ignores empty / None geometries on write
    expected_gdf = gdf.loc[~((gdf.geometry == None) | gdf.geometry.is_empty)]
else:
    expected_gdf = gdf


assert_geodataframe_equal(result_gdf, expected_gdf)
```

theroggy: I changed it, but now the test will start failing once the gdal bug causing this will be solved. I also added a reference to the bug involved: https://github.com/OSGeo/gdal/issues/7401

brendan-ward (approved):
Thanks @theroggy ; the failing test will be a helpful signal that GDAL fixed it.  Thanks for reporting the bug there.

jorisvandenbossche:
The geopandas test suite discovered one issue with this (https://github.com/geopandas/geopandas/actions/runs/4454470754/jobs/7823684402?pr=2834): if you have mixed 3D geometry types, the geometry_type gets inferred as "2.5D Unknown", which then gives "Geometry type is not supported: 2.5D Unknown"

I suppose for the case of mixed, we just need to use "Unknown" without prepending "2.5D"

theroggy:
> The geopandas test suite discovered one issue with this (https://github.com/geopandas/geopandas/actions/runs/4454470754/jobs/7823684402?pr=2834): if you have mixed 3D geometry types, the geometry_type gets inferred as "2.5D Unknown", which then gives "Geometry type is not supported: 2.5D Unknown"
> 
> I suppose for the case of mixed, we just need to use "Unknown" without prepending "2.5D"

I'll have a look...

PR #229, opened 2023-03-18
Fix geometry type "2.5D Unknown" is not supported

theroggy:
Fix issue raised in #223: https://github.com/geopandas/pyogrio/pull/223#issuecomment-1474795893


brendan-ward (approved):
Thanks for the quick fix @theroggy !

PR #234, opened 2023-04-02
ENH: Standardize 2.5D => Z in geometry type labels

### LINKED ISSUE #225: Choose a consistent way of dealing with +2D geometries
**Remark**; content here is being edited based on comments below to keep a consistent overview

Out in the wild there are 3 flavours how +2D geometry types are defined. With a "Point" layer as example: "2.5D POINT", "3D POINT" or "POINT Z".
- GDAL uses the following flavours:
    - in low-level API's: "2.5D POINT"
    - in this [ogrinfo example](https://live.osgeo.org/archive/10.0/en/quickstart/gdal_quickstart.html#:~:text=Get%20information%20about%20the%20vector%20data%20with%20ogrinfo) "3D Point" is shown for this type of layers
    - in [ogr2ogr -nlt](https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-nlt) (=force geometry_type): "POINT Z" (or POINT M, POINT ZM for elevation, measure, or elevation and measure). 
- [fiona](https://fiona.readthedocs.io/en/latest/manual.html#geometry-types) uses "3D POINT"
- in WKT you use "POINT Z"
- in the QGIS UI you don't choose a type like this explicily, but you choose "Point" and then have radiobuttons for "Z", and/or "M", which leans towards the "POINT Z" flavour: [new layer](https://docs.qgis.org/3.22/en/docs/training_manual/create_vector_data/create_new_vector.html#:~:text=Open%20QGIS%20and%20create%20a,to%20define%20a%20new%20layer).

In the pyogrio codebase and documentation some flavours are already in use:
- `pyogrio.read_info` returns geometry type "the GDAL-low-level way": "2.5D POINT"
- the geometry_type parameter in `write_dataframe` supports 2 flavours: "2.5D POINT" and "POINT Z"
- ...

Probably explicitly choosing a flavour to be returned in the pyogrio API (eg. `pyogrio.read_info`) would be a good idea. That would probably be the logical one to be used consinstently in the code base and documentation as well.

For input parameters it is also an option to support multiple flavours.

The issue became apparent in #223

brendan-ward:
Resolves #225

This standardizes from `2.5D <type>` to `<type> Z` for geometry type labels.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,8 +4,10 @@
 
 ### Improvements
 
-- Add automatic detection of 2.5D geometries in write_dataframe (#223, #229)
-- Add "driver" property to read_info result (#224)
+-   Add automatic detection of 3D geometries in write_dataframe (#223, #229)
+-   Add "driver" property to read_info result (#224)
+-   Standardized 3-dimensional geometry type labels from "2.5D <type>" to
+    "<type> Z" for consistency with well-known text (WKT) formats
 
 ## 0.5.1 (2023-01-26)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@
 -   Add automatic detection of 3D geometries in write_dataframe (#223, #229)
 -   Add "driver" property to read_info result (#224)
 -   Standardized 3-dimensional geometry type labels from "2.5D <type>" to
-    "<type> Z" for consistency with well-known text (WKT) formats
+    "<type> Z" for consistency with well-known text (WKT) formats (#234)
 
 ## 0.5.1 (2023-01-26)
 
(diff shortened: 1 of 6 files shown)


jorisvandenbossche (approved):
Looks good!

Do we want to keep some test with the 2.5D version to ensure this keeps working as well? (it's difficult to see from the diff if that's the case)

brendan-ward:
We have a test `test_geopandas_io.py::test_write_geometry_z_types` that tests setting the geometry type when writing a dataset that uses both `2.5D` and `Z` variants (didn't need to be updated here).  I think that tests the only case where geometry type is an input.

PR #237, opened 2023-04-02
ENH: Add basic support for data layer metadata IO

### LINKED ISSUE #218: ENH: add support for metadata tags if supported by driver
Allow passing in tags as a dict to `write()` / `write_dataframe()`, similar to [Fiona #821](https://github.com/Toblerity/Fiona/pull/821)

To support [GeoPandas #2777](https://github.com/geopandas/geopandas/issues/2777)

brendan-ward:
Resolves #218

Helps support GeoPandas [#2777](https://github.com/geopandas/geopandas/issues/2777) and [#2850](https://github.com/geopandas/geopandas/pull/2850)

GDAL silently ignores attempts to write metadata when not supported by the underlying driver.

The output of `read_info(...)` now includes a `metadata` key with the dict of metadata from the data layer (or None); note that depending on the driver, there may be metadata read even if not written.  E.g., the metadata of a shapefile can be `{'DBF_DATE_LAST_UPDATE': '2023-04-02'}`.

I did not add `metadata` to `read` because it didn't seem useful there.

/cc @DahnJ


pyogrio/geopandas.py
```diff
@@ -261,6 +262,10 @@ def write_dataframe(
         driver supports appending to an existing data source, will cause the
         data to be appended to the existing records in the data source.
         NOTE: append support is limited to specific drivers and GDAL versions.
+    metadata : dict, optional (defaul: None)
+        Metadata to be stored in the output file; limited to drivers that
+        support writing metadata (GPKG) and silently ignored otherwise.  Keys
```
jorisvandenbossche, line 267: Only GPKG? Or should this be "for example GPKG"?

jorisvandenbossche (commented):
Wondering: does GDAL only support setting it on the layer object? Or could it also be set on the dataset object?

brendan-ward:
GDAL appears to support reading / writing at the dataset or layer level, and seems to keep track of which was used (so they roundtrip).  This means we probably should pass this in as `dataset_metadata` or `layer_metadata`.

jorisvandenbossche:
Needs a merge of main.

I was initially wondering if we should given one of both (probably the layer metadata) the "preference" and use plain `metadata` for that one, for simplicity in case users just need that (I also think Fiona essentially only supports layer metadata, but maybe more accidentally, as the docstring indicates they support both, but the code seems to prefer the layer: https://github.com/Toblerity/Fiona/blob/bd2aabb950206da14a450bd0d863e0b9b92bf5bd/fiona/ogrext.pyx#L1455-L1477). 
Often, when writing a GeoDataFrame to a file, you just have a single layer, and this distinction is not that important. On the other hand, it's probably more explicit to just use `layer_metadata` and `dataset_metadata`.

brendan-ward:
Adding `metadata` as an alias of `layer_metadata` seems reasonable.

brendan-ward:
@jorisvandenbossche @martinfleis looks like we might be running into an auth issue with CircleCI that is coming from the GeoPandas organization level?  Perhaps an expired personal access token?  I don't recall that we specifically set up CircleCI for pyogrio.

jorisvandenbossche:
Sorry, you can ignore the Circle CI failure. That's indeed from https://github.com/geopandas/pyogrio/pull/243, I enabled Circle CI to try building linux arm wheels there (we use CircleCI for that in shapely). But since that's not yet working, will disable it again.

jorisvandenbossche (approved):

Issue #123, opened 2022-06-07
Support timezones when writing (and reading?) timezones


theroggy:
See https://github.com/Toblerity/Fiona/blob/791785b813fce97e5c3d60bd725fa2190645d2e3/fiona/ogrext.pyx#L401-L404 and https://github.com/Toblerity/Fiona/blob/791785b813fce97e5c3d60bd725fa2190645d2e3/fiona/ogrext.pyx#L431-L435 how fiona handles this (and https://gdal.org/development/rfc/rfc56_millisecond_precision.html?highlight=TZFlag#core-changes with very brief details)

m-richards:
Hey @theroggy have you done any work on this? If not I might have a look if that's okay.

theroggy:
> Hey @theroggy have you done any work on this? If not I might have a look if that's okay.

@m-richards No I haven't... feel free!

m-richards:
So having a look at this in a bit more detail there's more complexity than I originally realised (unless I'm missing something), numpy datetime64 doesn't support utc offsets- which seems to make numpy arrays as the intermediate between dataframes and ogr a little tricky to make work.

Some thoughts:
* Create a custom representation of a utc offset field as a datetime field and a second field encoding offset - bad for interop with other tools so not ideal
* Convert timezone aware fields to UTC on write, read them back as utc - simpler, but not a fully faithful round-trip, still need to handle naive datetimes and distinguishing them from utc ones.
* The last thought I have seems rather complex, but would it make sense to use the pandas `datetime64[ns, tz]` extension array dtype and feed that down (or the underlying representation) to `ogr_write`

(apologies if I'm missing something obvious, I thought this might be alright point to get familiar with some the pyogrio internals, but still verying much learning at this point)

theroggy:
Hmm... indeed... when writing it seems the datetimes at the moment will be implicitly converted to UTC already during the conversion of pandas series to numpy arrays:
https://pandas.pydata.org/docs/reference/api/pandas.Series.values.html

theroggy:
At the time of adding write support I noticed that the conversion to python `datetime` took quite some time... and so I wondered if it would be faster to convert the datetime columns to string and using (OGRParseDate)[https://gdal.org/api/vector_c_api.html#_CPPv412OGRParseDatePKcP8OGRFieldi] to let gdal parse them back to datetimes.

Not a super elegant solution, but it might be an option here as well...

jorisvandenbossche:
[...]

PR #253, opened 2023-06-03
ENH: support for reading and writing datetimes with timezones

m-richards:
This is an initial proof of concept into reading and writing datetimes with timezones by reading and writing strings from gdal, and parsing the strings on the pandas side, very directly following discussion and suggestions in #123 123.

I'm trying not to affect the numpy `read` api, as a this functionality doesn't make sense from the numpy layer, as it doesn't support timezones.

There is something missing from this, the numpy api should still be able to read datetimes with timezones and convert them correctly to UTC (currently timezones are ignored complete), but I think that can be a separate PR (perhaps it should even come before this).

There's also probably a fair bit of work to handle driver support for this, I noticed issues locally trying to write geopackages (using WSL with an old version of gdal). I had a quick look at the tests in fiona, I was hoping for a simple list of supported formats, but it's quite complex there). I figure though that's a secondary issue after figuring out the implementation.


pyogrio/geopandas.py
```diff
@@ -19,6 +20,18 @@ def _stringify_path(path):
     return path
 
 
+def _try_parse_datetime(ser):
+    import pandas as pd  # only called in a block where pandas is known to be installed
+
+    if Version(pd.__version__) >= Version("2.0.0"):
+        res = pd.to_datetime(ser, format="ISO8601")
+        if res.dtype != "object":
+            res = res.dt.as_unit("ms")
```
jorisvandenbossche, line 29: We might need to try to detect if there are timezones or not (eg check the first (few) strings to see if they contain a "+" or "-"), because if so we have to pass `utc=True` to not get object dtype:

```
In [20]: pd.to_datetime(["2021/01/02 01:02:03+01:00", "2021/01/02 01:02:03+01:00"], format="ISO8601")
Out[20]: DatetimeIndex(['2021-01-02 01:02:03+01:00', '2021-01-02 01:02:03+01:00'], dtype='datetime64[ns, UTC+01:00]', freq=None)

In [21]: pd.to_datetime(["2021/01/02 01:02:03+01:00", "2021/01/02 01:02:03+02:00"], format="ISO8601")
Out[21]: Index([2021-01-02 01:02:03+01:00, 2021-01-02 01:02:03+02:00], dtype='object')
```

(with longer timeseries, it is very normal to have mixed offsets for the same timezone, i.e. values with and without DST)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -139,6 +143,49 @@ def test_read_datetime(test_fgdb_vsi):
...
+    ser_localised = ser_naive.dt.tz_localize(tz)
+    df = gp.GeoDataFrame(
+        {"dates": ser_localised, "geometry": [Point(1, 1), Point(1, 1)]}
+    )
+    fpath = tmp_path / "test.geojson"
+    write_dataframe(df, fpath)
+    df_no_tz = read_dataframe(
+        fpath, datetime_as_string=False
+    )  # TODO this shouldn't be called datetime as string in the pandas layer,
+    #     should it even be accessible?
```
jorisvandenbossche, line 179: Unless we have some users asking for it, I wouldn't expose it in read_dataframe (so only in raw.read), just to keep it simpler in read_dataframe

jorisvandenbossche (commented):
Thanks a lot for looking into this! Added some comments / suggestions

pyogrio/geopandas.py
```diff
@@ -327,7 +344,12 @@ def write_dataframe(
     field_data = []
     field_mask = []
     for name in fields:
-        col = df[name].values
+        ser = df[name]
+        col = ser.values
+        if isinstance(ser.dtype, pd.DatetimeTZDtype):
+            # Deal with datetimes with timezones as strings
+            col = ser.astype(str)
```
jorisvandenbossche, line 351: How is this sufficient? (then GDAL sees string values and will write it as strings instead of datetimes in the file format?) I think we still need to pass down to _io.pyx somehow that those fields are datetime, so we create the OGRField with the correct type.

pyogrio/geopandas.py
m-richards: that's a good point, I checked with what we did in geopandas for the fiona IO and did the same thing as there to be consistent -> read without utc=True, if still object dtype then read with utc=True and see if that makes a difference.

I could also look at check for "+" and "-" as you suggest, I'd imagine that's faster, but maybe there are some edge cases?

pyogrio/geopandas.py
m-richards: It's not - I have no idea why I thought this would work. 

I've come back to this now, and as far as I can see, I don't think we can parse strings down and ask GDAL to convert them - at least not with what we have exposed in pyogrio. I'm certainly not well across GDAL so happy to be corrected, but it seems we would need to call `OGRParseDate` do to this (I can't see any other way). But this operates on the `OGRField *psField`, whereas the pyogrio code operates on `OGRFeatureH` using the setters and getters to update values for fields without them being exposed directly.

So in the updated code I've pushed, it's using a numpy object array of datetimes to pass the timezone information down for each row. I did have a decent look in the pandas code to see if there's a direct way of extracting utc offsets, but as far as I can see it's not provided as a vectorised method.

m-richards (commented):
@jorisvandenbossche Apologies for taking so long to get back to your comments. I've just replied to them / resolved some, and also push a change to how the writing is done - it should actually work this time.

pyogrio/geopandas.py
```diff
@@ -326,8 +353,23 @@ def write_dataframe(
...
+        col = ser.values
+        if isinstance(ser.dtype, pd.DatetimeTZDtype):
+            # Deal with datetimes with timezones by passing down timezone separately
+            # pass down naive datetime
+            col = ser.dt.tz_localize(None).values
+            # pandas only supports a single offset per column
+            # access via array since we want a numpy array not a series
+            # (only care about the utc offset, not actually the date)
+            # but ser.array.timetz won't have valid utc offset for pytz time zones
+            # (per https://docs.python.org/3/library/datetime.html#datetime.time.utcoffset) # noqa
```
m-richards, line 369: this is probably a bit verbose right now.
Basic idea of this is that I can't seem to find a way to produce a series/ numpy array of timezone offsets in a vectorised way through pandas.

Best I could seem to do is that `ser.array.to_pydatetime()` uses the cython function `ints_to_pydatetime`. Note the DateTimeArray is marked as experimental and doesn't actually document any methods on the website which isn't great. The equivalent Series / DateTimeIndex method is another option, but it throws unavoidable userwarnings for changing in behaviour.

This seems wasteful in terms of duplication (basically the same datetime information twice), but seemed better to pass this down and compute the offsets in cython (but I haven't profiled it either).

pyogrio/geopandas.py
```diff
@@ -19,6 +20,28 @@ def _stringify_path(path):
     return path
 
 
+def _try_parse_datetime(ser):
+    import pandas as pd  # only called in a block where pandas is known to be installed
+
+    pandas_ge_20 = Version(pd.__version__) >= Version("2.0.0")
```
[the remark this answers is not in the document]
brendan-ward: a global variable toward the top of `geopandas.py` seems reasonable; since this is pretty isolated, I don't think we need another file to manage that.

pyogrio/geopandas.py
```diff
@@ -19,6 +29,26 @@ def _stringify_path(path):
...
 
 
+def _try_parse_datetime(ser):
+    import pandas as pd  # only called when pandas is known to be installed
+
+    if PANDAS_GE_20:
+        datetime_kwargs = dict(format="ISO8601", errors="ignore")
+    else:
+        datetime_kwargs = dict(yearfirst=True)
+    res = pd.to_datetime(ser, **datetime_kwargs)
```
jorisvandenbossche, line 39: This first attempt will already raise a warning with the latest pandas, so we will need to catch that warning:

> FutureWarning: In a future version of pandas, parsing datetimes with mixed time zones will raise a warning unless `utc=True`. Please specify `utc=True` to opt in to the new behaviour and silence this warning. To create a `Series` with mixed offsets and `object` dtype, please use `apply` and `datetime.datetime.strptime`

(I have to check if the warning is actually wrong about the future behaviour of raising a warning. I would assume the deprecation would warn for an error in the future. In which case we should also already start catching that, I think)

pyogrio/geopandas.py
jorisvandenbossche: A potentially faster alternative:

```
In [20]: arr = pd.date_range("2012-10-25 09:00", periods=5, freq="D", tz="Europe/Brussels")

In [21]: arr.tz_localize(None) - arr.tz_convert("UTC").tz_localize(None)
Out[21]: 
TimedeltaIndex(['0 days 02:00:00', '0 days 02:00:00', '0 days 02:00:00',
                '0 days 01:00:00', '0 days 01:00:00'],
               dtype='timedelta64[ns]', freq=None)

```
Although it has some more steps, it's vectorized and avoids creating python datetime.datetime objects, and based on a quick test is much faster. And I _think_ it should give the same result?

pyogrio/raw.py
```diff
@@ -385,8 +391,10 @@ def write(
     metadata=None,
     dataset_options=None,
     layer_options=None,
+    timezone_cols_metadata=None,
     **kwargs,
 ):
+    kwargs.pop("dtypes", None)
```
jorisvandenbossche, line 397: This is needed for our roundtrip tests because `read` now returns dtypes as well?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -132,13 +135,52 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext):
...
+    assert_series_equal(df_read.col, expected)
+
+
+def test_write_datetime_mixed_offset(tmp_path):
+    # Summer Time (GMT+11), standard time (GMT+10)
+    dates = ["2023-01-01 11:00:01.111", "2023-06-01 10:00:01.111"]
+    tz = pytz.timezone("Australia/Sydney")
+    ser_naive = pd.Series(pd.to_datetime(dates), name="dates")
+    ser_localised = ser_naive.dt.tz_localize(tz)
+    ser_utc = ser_localised.dt.tz_convert(pytz.UTC)
```
jorisvandenbossche, line 169:
```suggestion
    ser_localised = ser_naive.dt.tz_localize("Australia/Sydney")
    ser_utc = ser_localised.dt.tz_convert("UTC")
```

(that avoids using pytz, and the string version will just work as well (and what most people do I think), and which is more robust for when pandas will switch from pytz to stdlib zoneinfo)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -132,13 +135,52 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext):
...
+
+    df = gp.GeoDataFrame(
+        {"dates": ser_localised, "geometry": [Point(1, 1), Point(1, 1)]}
+    )
+    fpath = tmp_path / "test.gpkg"
+    write_dataframe(df, fpath)
+    df_local = read_dataframe(fpath)
+    # GDAL tz only encodes offsets, not timezones, for multiple offsets
+    # read as utc datetime as otherwise would be read as string
+    assert_series_equal(ser_utc, df_local["dates"])
```
jorisvandenbossche, line 181:
```suggestion
    result = read_dataframe(fpath)
    # GDAL tz only encodes offsets, not timezones, for multiple offsets
    # read as utc datetime as otherwise would be read as string
    assert_series_equal(result["date"], ser_utc)
```

I find `df_local` confusing as name (it's not local time?)

jorisvandenbossche (commented):
And now from my part my apologies for taking so long to get back to this.

jorisvandenbossche:
I checked with `use_arrow=True`, and this is actually also not supported for reading: we always get naive wall-time timestamps as results (i.e. the same as we have now before this PR). Opened an issue on the GDAL side about this: https://github.com/OSGeo/gdal/issues/8460

pyogrio/geopandas.py
m-richards: Thanks, this looks much nicer! We can also do the conversion into the gdal representation directly rather than in cython. I've pushed a commit which should implement this. I will do some performance comparisons but haven't got to that yet.

pyogrio/raw.py
m-richards: Yep that's right- this was a quick hack to get some feedback on the tests without changing to much. When this is finalised the intent was to clean this up by doing the equivalent to this in individual tests, rather than pollute this function, I think that's cleaner?

pyogrio/tests/test_geopandas_io.py
m-richards: Makes sense - it's much nicer - I didn't realise this worked!

pyogrio/tests/test_geopandas_io.py
m-richards: Yep agree, I think this is leftover from an earlier version of the test

pyogrio/geopandas.py
```diff
@@ -368,13 +368,14 @@ def write_dataframe(
...
-            # (per https://docs.python.org/3/library/datetime.html#datetime.time.utcoffset) # noqa
-            timezone_cols_metadata[name] = ser.array.to_pydatetime()
+            naive = ser.dt.tz_localize(None)
+            col = naive.values
+            # compute offset relative to UTC explicitly
+            tz_offset = naive - ser.dt.tz_convert("UTC").dt.tz_localize(None)
+            # Convert to GDAL timezone offset representation.
+            # https://gdal.org/development/rfc/rfc56_millisecond_precision.html#core-changes
+            gdal_offset_representation = tz_offset // pd.Timedelta("15m") + 100
+            timezone_cols_metadata[name] = gdal_offset_representation
```
jorisvandenbossche, line 378:
```suggestion
            timezone_cols_metadata[name] = gdal_offset_representation.values
```

For the cython code, passing an array is fine (and will have a bit less overhead)

We should maybe also test with missing values (if you don't do that yet), because this might then be a float instead of int array, to ensure that doesn't give any problem in the cython code

m-richards: you just mean a null datetime right? (There's not some way to have a datetime with a null timezone that I don't know about?) I added an explicit test for this.

pyogrio/geopandas.py
```diff
@@ -27,6 +38,32 @@ def _stringify_path(path):
...
+        datetime_kwargs = dict(format="ISO8601", errors="ignore")
+    else:
+        datetime_kwargs = dict(yearfirst=True)
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            ".*parsing datetimes with mixed time zones will raise.*",
+            FutureWarning,
+        )
+        res = pd.to_datetime(ser, **datetime_kwargs)
```
jorisvandenbossche, line 54: I think we still might want to add a try/except around this, for when this would start raising an error in the future (also in that case we should fall back to the `utc=True` path below

pyogrio/geopandas.py
jorisvandenbossche: Yes, indeed a test where the datetime column has a missing value (NaT in pandas)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -649,7 +706,7 @@ def test_write_dataframe_gdal_options_unknown(tmp_path, naturalearth_lowres):
     df = read_dataframe(naturalearth_lowres)
 
     # geojson has no spatial index, so passing keyword should raise
-    outfilename = tmp_path / "test.geojson"
+    outfilename = tmp_path / "test.geosjon"
```
jorisvandenbossche, line 709: This seems a wrong change ;)
```suggestion
    outfilename = tmp_path / "test.geojson"
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -807,7 +864,7 @@ def test_write_dataframe_promote_to_multi_layer_geom_type_invalid(
 def test_write_dataframe_layer_geom_type_invalid(tmp_path, naturalearth_lowres):
     df = read_dataframe(naturalearth_lowres)
 
-    filename = tmp_path / "test.geojson"
+    filename = tmp_path / "test.geosjon"
```
jorisvandenbossche, line 867:
```suggestion
    filename = tmp_path / "test.geojson"
```

pyogrio/tests/test_geopandas_io.py
m-richards: Yeah thanks. sorry for the notification spam if you're getting it.

pyogrio/raw.py
m-richards: I had another look at this, we can move this pop to conftest.py::prepare_testfile for some cases, but there are still 22 tests that would need to pop the dtypes key. Another (still messy) option might be to add a kwarg to `read` along the lines of "include_dtypes_in_meta", defaulting to False, and is set to true in the datetime as string path (it could also be tied to the value of `datetime_as_string` but that feels confusing)

jorisvandenbossche: cc @brendan-ward (in case you can think of a better solution)  

Context is that this PR adds a new entry to the `meta` dictionary returned by `pyogrio.raw.read`, i.e. a "dtypes" key to know what the actual dtype is according to the OGR fields (which we need to know that the field was datetime64). 
But in `test_raw_io.py`, we have lots of places where we do a read and then write, and pass the `meta` from read to the write unpacked as keyword arguments with `**meta`. 

However, this "dtypes" key in `meta` is not an argument of `pyogrio.raw.write`, in contrast with the other keys (crs/encoding/fields/geometry_type), and so that would error if we don't remove "dtypes" from `meta`.

So in the current PR, to not have the current "hack" above of removing `dtypes` from the kwargs _inside_ `write`, we would have to constantly do `meta.pop("dtypes")` before passing them as `**meta` in the tests.

brendan-ward:
Apologies for the very slow review here; I am slowly trying to get up to speed and looking for other alternatives to adding dtype info to the `meta`.

pyogrio/raw.py
brendan-ward: I quite like having `dtypes` returned in `meta` in general.  I think popping it here makes sense, just add a comment indicating that we are doing it explicitly to allow passing `meta` produced by `read` directly into `write`.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -169,13 +172,67 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext):
...
 
+def test_read_datetime_tz(test_datetime_tz, tmp_path):
+    df = read_dataframe(test_datetime_tz)
+    raw_expected = ["2020-01-01T09:00:00.123-05:00", "2020-01-01T10:00:00-05:00"]
+
+    if PANDAS_GE_20:
+        expected = pd.to_datetime(raw_expected, format="ISO8601").as_unit("ms")
+    else:
+        expected = pd.to_datetime(raw_expected)
+    expected = pd.Series(expected, name="col")
```
brendan-ward, line 190:
```suggestion
    expected = pd.Series(expected, name="datetime_col")
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -169,13 +172,67 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext):
...
+    assert_series_equal(df.col, expected)
+    # test write and read round trips
+    fpath = tmp_path / "test.gpkg"
+    write_dataframe(df, fpath)
+    df_read = read_dataframe(fpath)
+    assert_series_equal(df_read.col, expected)
+
+
+def test_write_datetime_mixed_offset(tmp_path):
+    # Summer Time (GMT+11), standard time (GMT+10)
```
brendan-ward, line 200:
```suggestion
    # Australia Summer Time (GMT+11), Standard Time (GMT+10)
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -169,13 +172,67 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext):
...
+    dates = ["2023-01-01 11:00:01.111", "2023-06-01 10:00:01.111"]
+    ser_naive = pd.Series(pd.to_datetime(dates), name="dates")
```
brendan-ward, line 202:
```suggestion
    naive_col = pd.Series(pd.to_datetime(dates), name="dates")
```

(suggest `ser` => `col` and reverse order for others too)

pyogrio/tests/test_raw_io.py
```diff
@@ -775,6 +775,19 @@ def test_read_datetime_millisecond(test_datetime):
     assert field[1] == np.datetime64("2020-01-01 10:00:00.000")
 
 
+def test_read_datetime_tz(test_datetime_tz):
```
brendan-ward, line 778:
```suggestion
def test_read_datetime_as_string(test_datetime_tz):
```

This is testing both the `datetime_as_string` parameter as well as ensuring that the string values include timezone; probably best to name the test case after the parameter.

brendan-ward (commented):
Thanks @m-richards for all your work on this!  

I'll readily admit I don't work with datetimes very much, so the details around that are a bit beyond my experience.  However, your approach here seems like a reasonable way of getting around the limitations of datetime handling in numpy vs pandas.  Adding `dtypes` to `meta` seems reasonable, as does ignoring it on in `raw.write()`.

It might be a good idea to add some notes about this to `introduction.md` to guide users about when to use `datetime_as_string` and otherwise how the returned values will be dependent on the version of Pandas used (if I follow correctly).

m-richards:
> Thanks @m-richards for all your work on this!

Thanks for the detailed review, I hope I have addressed most of the suggestions now.

> It might be a good idea to add some notes about this to `introduction.md` to guide users about when to use `datetime_as_string` and otherwise how the returned values will be dependent on the version of Pandas used (if I follow correctly).

Happy to do this, but just want to double check the structuring. I am not super familiar with the pyogrio documentation, but `introduction.md` seems to all describe using `read_dataframe`. At present this pull request doesn't expose `datetime_as_string` to `read_dataframe`, only in `raw.py::read` so I don't know if it makes sense to document / whether it should be documented elsewhere?

On return types, happy to add this, but I saw there is already somewhat of a discussion of this at `known_issues.md` - which I now realise I need to update as part of this pull request too. Happy to have an example in the introduction as well though if you think that makes sense.

pyogrio/geopandas.py
```diff
@@ -389,19 +431,38 @@ def write_dataframe(
     # TODO: may need to fill in pd.NA, etc
     field_data = []
     field_mask = []
+    # dict[str, np.array(datetime.datetime)] special case for dt-tz fields
```
jorisvandenbossche, line 434:
```suggestion
    # dict[str, np.array(int)] special case for dt-tz fields
```

pyogrio/geopandas.py
```diff
@@ -389,19 +431,38 @@ def write_dataframe(
...
+    gdal_tz_offsets = {}
     for name in fields:
-        col = df[name].values
-        if isinstance(col, pd.api.extensions.ExtensionArray):
+        col = df[name]
+        values = col.values
```
jorisvandenbossche, line 438:
```suggestion
```

(duplicated with the `else` block below)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -167,17 +170,69 @@ def test_read_layer_invalid(naturalearth_lowres_all_ext, use_arrow):
 
 
 @pytest.mark.filterwarnings("ignore: Measured")
-def test_read_datetime(test_fgdb_vsi, use_arrow):
-    df = read_dataframe(
-        test_fgdb_vsi, layer="test_lines", use_arrow=use_arrow, max_features=1
-    )
-    if Version(pd.__version__) >= Version("2.0.0"):
+def test_read_datetime(test_fgdb_vsi):
+    df = read_dataframe(test_fgdb_vsi, layer="test_lines", max_features=1)
```
jorisvandenbossche, line 174:
```suggestion
def test_read_datetime(test_fgdb_vsi, use_arrow):
    df = read_dataframe(
        test_fgdb_vsi, layer="test_lines", use_arrow=use_arrow, max_features=1
    )
```

(essentially reverting the change on the lines above, this might be from a merge conflict)

jorisvandenbossche (approved):
Looking good!

jorisvandenbossche:
> At present this pull request doesn't expose `datetime_as_string` to `read_dataframe`, only in `raw.py::read` so I don't know if it makes sense to document / whether it should be documented elsewhere?

I would personally say it's good to keep this `raw` only, and then having the keyword explained in the docstrings seems good enough to me.

And the section in the "known_issues" about the datetime gotchas is already a nice and quite complete section I think. Maybe we could just move it "introduction", as now it's actually no longer a known "issue" with pyogrio, but general limitations of reading/writing datetimes because of GDAL's data model.

brendan-ward:
Apologies for the poor direction.  Leaving `datetime_as_string` at `raw` level only is fine.  I'd suggest moving the datimetime section from known issues to the introduction.

pyogrio/_io.pyx
```diff
@@ -1647,7 +1664,8 @@ def ogr_write(
     str path, str layer, str driver, geometry, fields, field_data, field_mask,
     str crs, str geometry_type, str encoding, object dataset_kwargs,
     object layer_kwargs, bint promote_to_multi=False, bint nan_as_null=True,
-    bint append=False, dataset_metadata=None, layer_metadata=None
+    bint append=False, dataset_metadata=None, layer_metadata=None,
+    timezone_cols_metadata=None
```
brendan-ward, line 1668:
```suggestion
    gdal_tz_offsets=None
```

brendan-ward (commented):
Thanks for the updates @m-richards !  Getting really close.  Looks like there are a few uncommitted suggestions, but after those are in this looks good to me.

m-richards:
> Thanks for the updates @m-richards ! Getting really close. Looks like there are a few uncommitted suggestions, but after those are in this looks good to me.

I think they're all done now, don't know how I missed them before.

brendan-ward:
Thanks @m-richards ! 🚀

PR #260, opened 2023-08-09
BUG: Fix integer overflow when reading int64 column

### LINKED ISSUE #259: BUG: integer overflow when reading .gpkg
When reading a .gpkg file with an integer column with values > the max 32 bit int value (2.147.483.647), the values overflows to negative values when read.

The use_arrow=True code path doesn't have this issue.

Code to reproduce behaviour
```
from pathlib import Path

import geopandas as gpd
import pyogrio
import shapely


script_dir = Path(__file__).resolve().parent
path = script_dir / "int_overflow_tmp.gpkg"
gdf = gpd.GeoDataFrame(
    data={"overflowing_int": [2192502720]}, geometry=[shapely.Point(0,1)], crs="epsg:31370"
)
pyogrio.write_dataframe(gdf, path)

gdf = pyogrio.read_dataframe(path, use_arrow=False)
print(f"with use_arrow=False:\n{gdf}")
gdf = pyogrio.read_dataframe(path, use_arrow=True)
print(f"with use_arrow=True:\n{gdf}")
```

theroggy:
Closes #259

### DIFF
--- a/pyogrio/tests/test_raw_io.py
+++ b/pyogrio/tests/test_raw_io.py
@@ -598,8 +598,8 @@ def test_read_write_int64_large(tmp_path, ext):
     filename = tmp_path / f"test.{ext}"
     write(filename, geometry, field_data, fields, **meta)
     result = read(filename)[3]
-    assert all([np.array_equal(f1, f2) for f1, f2 in zip(result, field_data)])
-    assert all([f1.dtype == f2.dtype for f1, f2 in zip(result, field_data)])
+    assert np.array_equal(result, field_data)
+    assert result[0].dtype == field_data[0].dtype
 
 
 def test_read_data_types_numeric_with_null(test_gpkg_nulls):


pyogrio/_ogr.pxd
```diff
@@ -214,7 +214,7 @@ cdef extern from "ogr_api.h":
     int             OGR_F_GetFieldAsDateTimeEx(OGRFeatureH feature, int n, int *y, int *m, int *d, int *h, int *m, float *s, int *z)
     double          OGR_F_GetFieldAsDouble(OGRFeatureH feature, int n)
     int             OGR_F_GetFieldAsInteger(OGRFeatureH feature, int n)
-    long            OGR_F_GetFieldAsInteger64(OGRFeatureH feature, int n)
+    int64_t         OGR_F_GetFieldAsInteger64(OGRFeatureH feature, int n)
```
brendan-ward, line 217: Good catch, looks like this should have originally been a `long long`, which would then have matched [Fiona](https://github.com/Toblerity/Fiona/blob/master/fiona/ogrext3.pxd#L338)

`int64_t` seems preferable because it is more explicit / less prone to error.

Would you mind updating `OGR_F_SetFieldInteger64` to `OGR_F_SetFieldInteger64(OGRFeatureH feature, int n, int64_t value)` to match, for symmetry?

pyogrio/tests/test_raw_io.py
```diff
@@ -583,6 +583,25 @@ def test_read_write_datetime(tmp_path):
...
+        [bytes.fromhex("010100000000000000000000000000000000000000")] * 3, dtype=object
+    )
+    field_data = [np.array([1, 2192502720, -5], dtype="int64")]
+    fields = ["overflow_int64"]
+    meta = dict(geometry_type="Point", crs="EPSG:4326", spatial_index=False)
+
+    filename = tmp_path / f"test.{ext}"
+    write(filename, geometry, field_data, fields, **meta)
+    result = read(filename)[3]
+    assert all([np.array_equal(f1, f2) for f1, f2 in zip(result, field_data)])
```
brendan-ward, line 601:
```suggestion
    assert np.array_equal(result, field_data)
```

pyogrio/tests/test_raw_io.py
```diff
@@ -583,6 +583,25 @@ def test_read_write_datetime(tmp_path):
...
+    assert all([f1.dtype == f2.dtype for f1, f2 in zip(result, field_data)])
```
brendan-ward, line 602: Since there is only the one field, this could be simplified to

```
assert result[0].dtype == field_data[0].dtype
```

though it is also fine as is.

brendan-ward (commented):
Thanks @theroggy !

pyogrio/_ogr.pxd
theroggy: Sure!

pyogrio/tests/test_raw_io.py
theroggy: ok!

pyogrio/tests/test_raw_io.py
theroggy: True, indeed simpler and more readable...

brendan-ward (approved):

OwenLu88:
I'm using the 0.6.0 version now. I don't know how to fix it in the current version. Could you help me?

theroggy:
It depends a bit on your specific needs, but using the current version you could try using `read_datafram` with `use_arrow=True`. This is a bit experimental, so you could run in other issues, but it is something you can try without having to wait for a new version to land.

OwenLu88:
OK. Thanks. I will try it.
BTW, do you have a plan to release a new version(0.7.0)?

OwenLu88:
Wow, it's a good solution using read_dataframe with use_arrow=True can fix the issue. Very very thanks.

theroggy:
> OK. Thanks. I will try it. BTW, do you have a plan to release a new version(0.7.0)?

Yes, but there is no date fixed yet.

OwenLu88:
I run an executable file after running pyinstaller. It reported, "Failed to execute script 'main' due to unhandled exception: No module named 'pyogrio._geometry'". Why is it?

theroggy:
> I run an executable file after running pyinstaller. It reported, "Failed to execute script 'main' due to unhandled exception: No module named 'pyogrio._geometry'". Why is it?

I don't think this is related to this PR? Anyway, I'm not familiar with pyinstaller, so I'm afraid I cannot help you.

OwenLu88:
I got it.  Thanks.

PR #261, opened 2023-08-09
DOC: instructions to setup development env on windows using conda

theroggy:
This is the way I develop on windows...

### DIFF
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -150,7 +150,7 @@ Before being able to build on Windows, you need to set some additional environme
 variables because the location of the GDAL binaries and headers cannot be
 automatically determined.
 
-After activating the pyogrio-dev environment the CONDA_PREFIX environment variable will
+After activating the `pyogrio-dev` environment the `CONDA_PREFIX` environment variable will
 be set and you will be alble to set the necessary environment variables as follows:
 
 ```bash
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -91,6 +91,11 @@ pytest pyogrio/tests
 
 ### Windows
 
+There are different ways to install the necessary dependencies and setup your local
+development environment on windows.
+
+#### vcpkg
+
 [vcpkg](https://vcpkg.io/en/index.html) is used to build pyogrio from source
 as part of creating the Pyogrio Python wheels for Windows. You can install
 GDAL and other dependencies using vcpkg, and then build Pyogrio from source.
@@ -101,8 +106,10 @@ of the wheel-building process.
 We do not yet have instructions on building Pyogrio from source using vcpkg for
 local development; please feel free to contribute additional documentation!
 
-Alternatively, you can install GDAL from an appropriate provider of Windows
-binaries. We've heard that the [OSGeo4W](https://trac.osgeo.org/osgeo4w/) works.
+#### OSGeo4W
+
+You can also install GDAL from an appropriate provider of Windows binaries. We've heard
+that the [OSGeo4W](https://trac.osgeo.org/osgeo4w/) works.
 
 To build on Windows, you need to provide additional environment variables or
 command-line parameters because the location of the GDAL binaries and headers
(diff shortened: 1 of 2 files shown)


docs/source/install.md
```diff
@@ -129,6 +136,32 @@ The location of the GDAL DLLs must be on your system `PATH`.
...
+It is also possible to install the necessary dependencies using conda.
+
+After cloning the environment, you can create a conda environment with the necessary
+dependencies like this:
+```
+conda env create -f environment-dev.yml
+```
+
+Before being able to build on Windows, you need to set some additional environment
+variables or because the location of the GDAL binaries and headers cannot be
```
brendan-ward, line 150:
```suggestion
variables because the location of the GDAL binaries and headers cannot be
```

docs/source/install.md
```diff
@@ -129,6 +136,32 @@ The location of the GDAL DLLs must be on your system `PATH`.
...
+automatically determined.
+
+After activating the pyogrio-dev environment the CONDA_PREFIX environment variable will
```
brendan-ward, line 153:
```suggestion
After activating the `pyogrio-dev` environment the `CONDA_PREFIX` environment variable will
```

docs/source/install.md
```diff
@@ -129,6 +136,32 @@ The location of the GDAL DLLs must be on your system `PATH`.
...
+be set and you will be alble to set the necessary environment variables as follows:
```
brendan-ward, line 154:
```suggestion
be set and you will be able to set the necessary environment variables as follows:
```

environment-dev.yml
```diff
@@ -0,0 +1,13 @@
+name: pyogrio-dev
+channels:
+  - conda-forge
+dependencies:
+  - cython
+  - numpy
+  - gdal ==3.7.0
```
brendan-ward, line 7: Can`gdal` be dropped and just use `libgdal==3.5.1`?  (only picked this version to match `ci/environment-libgdal3.5.1.yml`)

Other than our CI tests for `osgeo/gdal:ubuntu-small-latest`, we're not yet specifically testing GDAL > 3.6.2.  Though when we bump the versions we're testing, we should also do so in this environment...

docs/source/install.md
```diff
@@ -129,6 +136,32 @@ The location of the GDAL DLLs must be on your system `PATH`.
...
+
+```bash
+set GDAL_INCLUDE_PATH=%CONDA_PREFIX%\Library\include
+set GDAL_LIBRARY_PATH=%CONDA_PREFIX%\Library\lib
+set GDAL_VERSION=3.7.0
```
brendan-ward, line 159: Is it possible to set this dynamically?  For our windows tests, we're deriving this dynamically `GDAL_VERSION=$(gdalinfo --version | cut -c 6-10)`

See `.github/workflows/tests-conda.yml`.

Would be good to not have this value hard-coded, if possible.

brendan-ward (commented):
Thanks @theroggy !  I don't have a windows machine to test this against, but enough of this is similar to our Windows CI tests that it seems reasonable, and it works for you.

environment-dev.yml
theroggy: gdal==3.5.1 seemed to give some weird issues in combination with pyarrow, but gdal 3.6.2 seems to work fine.

I also added pre-commit and mentioned in the general develop that it is recommended to install pre-commit. Feedback welcome.

docs/source/install.md
theroggy: No. That command only works in a bash shell window, which isn't very common on windows. In the standard shells, cmd or powershell, this doesn't work. The tooling available in e.g. cmd is also very limited, so not trivial (at least to me) to write something similar and even worse to have it work both in powershell and cmd.

I haven't looked into it and am not familiar with it, but isn't it possible to determine the version in the build script? If I recall correctly the GDAL_VERSION environment variable wasn't needed in the beginning.

brendan-ward: We use it early on to check that it is a supported version of GDAL.  Thanks for considering more general approaches; your updates to this section are fine.

brendan-ward (approved):
Thanks @theroggy !

PR #267, opened 2023-08-14
ENH: support writing dataframes without geometry column

### LINKED ISSUE #105: ENH: Support writing dataframes (without geometry)

theroggy:
closes #105

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
--   Support writing dataframes without geometry (#267)
+-   Support writing dataframes without geometry column (#267)
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
--   Support writing dataframes without geometry (#)
+-   Support writing dataframes without geometry (#267)
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## 0.6.1 (???)
 
+### Improvements
+
+-   Support writing dataframes without geometry (#)
+
 ### Bug fixes
 
 -   Fix int32 overflow when reading int64 columns (#260)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## 0.6.1 (???)
+## 0.7.0 (???)
 
 ### Improvements
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1504,16 +1504,26 @@ def ogr_write(
     cdef OGRwkbGeometryType geometry_code
     cdef int err = 0
     cdef int i = 0
-    cdef int num_records = len(geometry)
+    cdef int num_records = -1
     cdef int num_fields = len(field_data) if field_data else 0
 
     if len(field_data) != len(fields):
         raise ValueError("field_data and fields must be same length")
 
-    if num_fields:
-        for i in range(1, len(field_data)):
-            if len(field_data[i]) != num_records:
-                raise ValueError("field_data arrays must be same length as geometry array")
+    if geometry is not None:
+        num_records = len(geometry)
+        if num_fields:
+            for i in range(1, len(field_data)):
+                if len(field_data[i]) != num_records:
+                    raise ValueError(
+                        "field_data arrays must be same length as geometry array"
+                    )
+    elif num_fields == 0:
+        raise ValueError(
+            "geometry is None without any fields is not supported."
+        )
+    else:
+        num_records = len(field_data[0])
 
     if field_mask is not None:
         if len(field_data) != len(field_mask):
@@ -1706,7 +1716,9 @@ def ogr_write(
 
             # create the geometry based on specific WKB type (there might be mixed types in geometries)
             # TODO: geometry must not be null or errors
-            wkb = geometry[i]
+            wkb = None
+            if geometry is not None:
+                wkb = geometry[i]
             if wkb is not None:
                 wkbtype = <int>bytearray(wkb)[1]
                 # may need to consider all 4 bytes: int.from_bytes(wkb[0][1:4], byteorder="little")
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1505,9 +1505,10 @@ def ogr_write(
     cdef int err = 0
     cdef int i = 0
     cdef int num_records = -1
-    cdef int num_fields = len(field_data) if field_data else 0
+    cdef int num_field_data = len(field_data) if field_data is not None else 0
+    cdef int num_fields = len(fields) if fields is not None else 0
 
-    if len(field_data) != len(fields):
+    if num_fields != num_field_data:
         raise ValueError("field_data and fields must be same length")
 
     if geometry is not None:
@@ -1520,7 +1521,7 @@ def ogr_write(
                     )
     elif num_fields == 0:
         raise ValueError(
-            "geometry is None without any fields is not supported."
+            "having no geometry column as well as no fields is not supported"
         )
     else:
         num_records = len(field_data[0])
@@ -1532,7 +1533,7 @@ def ogr_write(
             if field_mask[i] is not None and len(field_mask[i]) != num_records:
                 raise ValueError("field_mask arrays must be same length as geometry array")
     else:
-        field_mask = [None] * len(field_data)
+        field_mask = [None] * num_fields
 
     path_b = path.encode('UTF-8')
     path_c = path_b
@@ -1658,7 +1659,9 @@ def ogr_write(
             layer_options = NULL
 
     ### Create the fields
-    field_types = infer_field_types([field.dtype for field in field_data])
+    field_types = None
+    if num_fields > 0:
+        field_types = infer_field_types([field.dtype for field in field_data])
 
     ### Create the fields
     if create_layer:
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1509,22 +1509,25 @@ def ogr_write(
     cdef int num_fields = len(fields) if fields is not None else 0
 
     if num_fields != num_field_data:
-        raise ValueError("field_data and fields must be same length")
+        raise ValueError("field_data array needs to be same length as fields array")
+
+    if num_fields == 0 and geometry is None:
+        raise ValueError("You must provide at least a geometry column or a field")
+    
+    if num_fields > 0:
+        num_records = len(field_data[0])
+        for i in range(1, len(field_data)):
+            if len(field_data[i]) != num_records:
+                raise ValueError("field_data arrays must be same length")
 
     if geometry is not None:
-        num_records = len(geometry)
-        if num_fields:
-            for i in range(1, len(field_data)):
-                if len(field_data[i]) != num_records:
-                    raise ValueError(
-                        "field_data arrays must be same length as geometry array"
-                    )
-    elif num_fields == 0:
-        raise ValueError(
-            "having no geometry column as well as no fields is not supported"
-        )
-    else:
-        num_records = len(field_data[0])
+        if num_fields > 0:
+            if len(geometry) != num_records:
+                raise ValueError(
+                    "field_data arrays must be same length as geometry array"
+                )
+        else:
+            num_records = len(geometry)
 
     if field_mask is not None:
         if len(field_data) != len(field_mask):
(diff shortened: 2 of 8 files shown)


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## 0.6.1 (???)
```
brendan-ward, line 3: Since this is a new feature, we probably should bump the version to 0.7

pyogrio/geopandas.py
```diff
@@ -411,11 +415,17 @@ def write_dataframe(
         else:
             crs = geometry.crs.to_wkt(WktVersion.WKT1_GDAL)
 
+    # Prepare geometry array to write
+    if geometry_column is not None:
+        geometry = to_wkb(geometry.values)
+    else:
+        geometry = np.repeat(None, len(df))
```
brendan-ward, line 422: I could be wrong (didn't test the idea), but it seems like with some relatively minor changes in `_io.pyx::ogr_write(...)` that it would be possible to pass in `geometry=None` and avoid passing in an array of `None` values.

We may also want to raise an error if `geometry` is `None` and there is no field data.

For instance, writing a data frame that has no columns produces different behavior depending on the driver: GDAL automatically creates a `FID` column if there are no columns passed in, whereas geopackage will be read back as an empty data frame (by default; it _does_ create an FID column as well)

brendan-ward (commented):
Thanks for working on this @theroggy !  I needed this exact functionality recently and hadn't had a chance to work on it, so I'm very grateful you got this going.

I think the primary open-ended question here is whether or not to allow passing `geometry=None` instead of an array of `None` values to `ogr_write(...)`.

pyogrio/geopandas.py
```diff
@@ -306,25 +306,27 @@ def write_dataframe(
 
     path = str(path)
 
-    if not isinstance(df, gp.GeoDataFrame):
-        raise ValueError("'df' must be a GeoDataFrame")
+    if not isinstance(df, (gp.GeoDataFrame, pd.DataFrame)):
```
jorisvandenbossche, line 309:
```suggestion
    if not isinstance(df, pd.DataFrame):
```

Since GeoDataFrame is a subclass, isinstance works with DataFrame as well

pyogrio/tests/test_geopandas_io.py
```diff
@@ -460,6 +460,51 @@ def test_write_dataframe(tmp_path, naturalearth_lowres, ext):
     )
 
 
+@pytest.mark.parametrize("ext", [ext for ext in ALL_EXTS if ext != ".fgb"])
```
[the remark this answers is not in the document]
theroggy: I don't seem to get this warning, but added the filter. Possibly dependent on gdal version.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -460,6 +460,51 @@ def test_write_dataframe(tmp_path, naturalearth_lowres, ext):
...
+    if ext == ".shp":
+        # A shapefile without geometry column results in only a .dbf file.
+        output_path = output_path.with_suffix(".dbf")
+    assert output_path.exists()
+    result_df = read_dataframe(output_path)
+
+    assert isinstance(result_df, pd.DataFrame)
+
+    # some dtypes do not round-trip precisely through these file types
+    check_dtype = ext not in [".json", ".geojson", ".geojsonl", ".xlsx"]
```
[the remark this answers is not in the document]
theroggy: Not really to ALL_EXT but specifically to this test because xlsx is now removed from the unsupported drivers.

But it kind of depends on the extended write driver detection so I'll remove it... xlsx isn't that critical anyway I suppose.

pyogrio/geopandas.py
theroggy: ok

CHANGES.md
theroggy: True

pyogrio/tests/test_geopandas_io.py
```diff
@@ -460,6 +460,51 @@ def test_write_dataframe(tmp_path, naturalearth_lowres, ext):
...
+    input_df = input_gdf.drop(columns=["geometry"])
+    assert isinstance(input_df, pd.DataFrame)
+    assert not isinstance(input_df, gp.GeoDataFrame)
+
+    output_path = tmp_path / f"test{ext}"
+    write_dataframe(input_df, output_path)
+
+    if ext == ".shp":
+        # A shapefile without geometry column results in only a .dbf file.
+        output_path = output_path.with_suffix(".dbf")
```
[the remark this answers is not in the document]
theroggy: OK

pyogrio/geopandas.py
theroggy: I had given it a brief thought, but because the number of rows was based on the geometry field and it is possible not to have any columns I didn't pursue it.

However, having our discussion in #263 about reading no geometry nor any columns in mind, as well as your following suggestion it should indeed not be a problem to do so.
    `We may also want to raise an error if geometry is None and there is no field data.`

So, I just committed a version that supports this (and fixes support for field_data and fields being None as well).

pyogrio/_io.pyx
```diff
@@ -1504,16 +1504,27 @@ def ogr_write(
...
 
-    if len(field_data) != len(fields):
+    if num_fields != num_field_data:
         raise ValueError("field_data and fields must be same length")
 
-    if num_fields:
-        for i in range(1, len(field_data)):
-            if len(field_data[i]) != num_records:
-                raise ValueError("field_data arrays must be same length as geometry array")
+    if geometry is not None:
```
brendan-ward, line 1514: It might be cleaner to validate that all fields have the same length first (if applicable) and then compare the length of the first field (if present) to the length of the geometry column (if present).  

```
if num_fields == 0 and geometry is None:
    raise ValueError("You must provide at least a geometry column or field data")

if num_fields > 0:
    num_records = len(field_data[0])
    for i in range (1, len(field_data)):
        if len(field_data[i) != num_records
            raise ValueError("all field_data arrays must be the same length")

if geometry is not None:
    if num_fields > 0:
        if len(geometry) != num_records:
            raise ValueError("field_data arrays must be same length as geometry array")
    else:
        num_records = len(geometry)
```

pyogrio/_io.pyx
```diff
@@ -1706,7 +1719,9 @@ def ogr_write(
 
             # create the geometry based on specific WKB type (there might be mixed types in geometries)
             # TODO: geometry must not be null or errors
-            wkb = geometry[i]
+            wkb = None
```
brendan-ward, line 1722:
```suggestion
            wkb = None if geometry is None else geometry[i]
```

pyogrio/raw.py
```diff
@@ -421,13 +418,13 @@ def write(
                 if not isinstance(v, str):
                     raise ValueError(f"metadata value {v} must be a string")
 
-    if promote_to_multi is None:
+    if geometry_type is not None and promote_to_multi is None:
```
brendan-ward, line 421:
```suggestion
    if geometry is not None and promote_to_multi is None:
```

pyogrio/raw.py
```diff
@@ -421,13 +418,13 @@ def write(
...
         promote_to_multi = (
             geometry_type.startswith("Multi")
             and driver in DRIVERS_NO_MIXED_SINGLE_MULTI
         )
 
-    if crs is None:
+    if geometry_type is not None and crs is None:
```
brendan-ward, line 427:
```suggestion
    if geometry is not None and crs is None:
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -460,6 +460,56 @@ def test_write_dataframe(tmp_path, naturalearth_lowres, ext):
     )
 
 
+@pytest.mark.filterwarnings("ignore:.*No SRS set on layer.*")
+@pytest.mark.parametrize("ext", [ext for ext in ALL_EXTS + [".xlsx"] if ext != ".fgb"])
+def test_write_dataframe_nogeom(tmp_path, naturalearth_lowres, ext):
+    """Test writing a dataframe, so without a geometry column.
```
brendan-ward, line 466:
```suggestion
    """Test writing a dataframe without a geometry column.
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -460,6 +460,56 @@ def test_write_dataframe(tmp_path, naturalearth_lowres, ext):
...
+
+    FlatGeobuf (.fgb) doesn't seem to support this, and just writes an empty file.
+    """
+    # Prepare test data
+    input_gdf = read_dataframe(naturalearth_lowres)
```
brendan-ward, line 471:
```suggestion
    input_df = read_dataframe(naturalearth_lowres).drop(columns=["geometry"])
```

pyogrio/tests/test_raw_io.py
```diff
@@ -316,6 +316,79 @@ def test_write_geojson(tmpdir, naturalearth_lowres):
...
+    field_data = None
+    meta["fields"] = None
+
+    # Test
+    filename = tmp_path / "test.gpkg"
+    with pytest.raises(
+        ValueError,
+        match="having no geometry column as well as no fields is not supported",
+    ):
+        write(filename, geometry, field_data, driver="GPKG", **meta)
```
brendan-ward, line 389:
```suggestion
        write(filename, geometry=None, field_data=None, driver="GPKG", **meta)
```

pyogrio/tests/test_raw_io.py
```diff
@@ -316,6 +316,79 @@ def test_write_geojson(tmpdir, naturalearth_lowres):
...
+        "gdp_md_est",
+    ]
+
+    assert len(fields) == 5
+    assert len(fields[0]) == 177
+
+
+def test_write_nogeom_nofields(tmp_path, naturalearth_lowres):
+    # Prepare test data
+    meta, _, geometry, field_data = read(naturalearth_lowres)
```
brendan-ward, line 377:
```suggestion
    meta = read(naturalearth_lowres)[0]
```
(other return values not used)

brendan-ward (commented):
Thanks for updating this @theroggy , this is getting close; a few minor suggestions.  Apologies for the slow review.

pyogrio/_io.pyx
theroggy: This indeed increases readability!

pyogrio/_io.pyx
theroggy: ok!

pyogrio/raw.py
theroggy: I think it is a bit weird here to check on geometry because the code is about the geometry_type, but it won't influence anything so it doesn't matter so as you seem to have a preference I'll change it.

pyogrio/tests/test_geopandas_io.py
theroggy: ok

pyogrio/tests/test_geopandas_io.py
theroggy: I added parameter read_geometry=False, than it doesn't even need to be dropped...

pyogrio/tests/test_raw_io.py
theroggy: I simplified the test some further, and don't read any input data at all anymore...

pyogrio/raw.py
theroggy: If `geometry_type` is specified, but `geometry=None` at the moment the geometry column is created but with all geometry values None. In this case I think the crs should still be specified, so I don't think this should be changed?

brendan-ward: If geometry is `None` there is no geometry column, and thus not filled with `None` values.  I was intending the change to be based on whether or not there is a `geometry` column; if there isn't then updating geometry type or needing a CRS are not relevant.  I'm not sure it is useful to allow the user to specify `geometry_type` but not `geometry`; that seems more like an error.  Or put differently, the presence of a geometry column signals that there should also be a CRS.

theroggy: As you wish, if someone wants a geometry column with all None values they can also still use the `np.repeat(None,...)` trick.

So, I'll add an error if geometry_type is not None and geometry is None?

brendan-ward: I don't know that we need to specifically raise an error in that case; it should be uncommon.  I think it is more that we'll simply ignore `geometry_type` if there is no geometry, which seems OK.

theroggy: Ok, it doesn't matter to me, they are indeed all uncommon cases that aren't really interesting, so as long as it they don't give weird errors or side effects it's all fine for me.

So I'll just overrule `geometry_type` and set it to `None` if `geometry` is `None`.

brendan-ward (approved):
Thanks @theroggy !

Issue #220, opened 2023-02-13
ENH: Auto-detect driver from filename extension when writing


jorisvandenbossche:
We currently already have some limited, hardcoded version of this using:

https://github.com/geopandas/pyogrio/blob/75e8f13940fea6e30554115760275c7da978058c/pyogrio/raw.py#L19-L27

But now we have initial support for internally using the GDAL driver metadata (https://github.com/geopandas/pyogrio/issues/103, first used in https://github.com/geopandas/pyogrio/pull/189), we could generalize this based on the extension information stored in the driver metadata. 
Equivalent fiona PR: https://github.com/Toblerity/Fiona/pull/948

PR #268, opened 2023-08-15
Use gdal driver detection based on output path for writing files

theroggy:
Because I wanted to test .xlsx for #267, I was reminded that the driver detection for writing is really minimal at the moment.

For reading the driver detection is implicitly done by gdal resulting in automatic support of +- all gdal drivers. This PR is a proposal to follow the same principle: rely as much as possible on existing gdal logic to infer the driver from the output file.

Because it introduces a dependency on the gdal python bindings, I started with this first version as a POC. If the idea is supported I'll finish the implementation.

PS: I started this PR based on #267 to have the xlsx test case handy... but only the last commit is relevant for this PR.

### DIFF
--- a/pyogrio/raw.py
+++ b/pyogrio/raw.py
@@ -1,6 +1,8 @@
 import warnings
 import os
 
+from osgeo_utils.auxiliary.util import GetOutputDriversFor
+
 from pyogrio._env import GDALEnv
 from pyogrio.errors import DataSourceError
 from pyogrio.util import get_vsi_path
@@ -308,23 +310,20 @@ def open_arrow(
 
 
 def detect_driver(path):
-    # try to infer driver from path
+    # Default gdal detection doesn't handle .geojson as we want, so special treatment
     parts = os.path.splitext(path)
-    if len(parts) != 2:
+    if len(parts) == 2 and parts[1] in [".geojson", ".json"]:
+        return "GeoJSON"
+
+    # Infer driver using default gdal implementation
+    drivers = GetOutputDriversFor(path, is_raster=False)
+    if len(drivers) == 1:
+        return drivers[0]
+    else:
         raise ValueError(
             f"Could not infer driver from path: {path}; please specify driver "
             "explicitly"
         )
-
-    ext = parts[1].lower()
-    driver = DRIVERS.get(ext, None)
-    if driver is None:
-        raise ValueError(
-            f"Could not infer driver from path: {path}; please specify driver "
-            "explicitly"
-        )
-
-    return driver
 
 
 def _parse_options_names(xml):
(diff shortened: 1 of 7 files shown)


pyogrio/raw.py
```diff
@@ -1,6 +1,8 @@
 import warnings
 import os
 
+from osgeo_utils.auxiliary.util import GetOutputDriversFor
```
brendan-ward, line 4: I'd rather avoid this dependency and only rely on the parts of GDAL that we are binding against in Cython.

Looking at the [implementation](https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py#L62) in GDAL, it looks like it is mostly a matter of iterating over drivers that support write and checking the driver's `GDAL_DMD_EXTENSION` [metadata item](https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py#L57).

From there, it seems like we could handle this in a couple different ways:
- build a map of extensions to drivers when pyogrio is first initialized (roughly similar to how we use `DRIVERS` now)
- do the driver lookup at runtime based on the extension of a specific path

brendan-ward: In fact, it might be possible to make some minor modifications to `_ogr.pyx::ogr_list_drivers()` to return both the read / write capability of a driver as well as associated extensions, which we could then use to create an inverse map of extension to driver.

jorisvandenbossche: Agreed that we don't want to depend on the GDAL python bindings (that would be an additional dependency, which also has no wheels, and thus would defeat the purpose of our wheels as a way to get GDAL).

As Brendan mentioned, I think we can indeed use the metadata from GDAL, now that we have bindings to getting that information (`_get_driver_metadata_item`). We have an open issue for this feature that mentions this as well (https://github.com/geopandas/pyogrio/issues/220), and that issue also links to the equivalent PR in fiona to add this feature (https://github.com/Toblerity/Fiona/pull/948), that might be closer to our code to look at compared to the GDAL implementation.

theroggy: Yes, I also already had a look at the implementation in gdal and it is indeed mainly based on the drivers `DMD_EXTENSIONS` metadata, with some extra logic.

I'm wondering, I also encountered a similar C++ function with even some more logic in it, would it be an option that is gets exposed in the C API. This could deduplicate the logic in gdal itself as wel as make it reusable in fiona, rasterio, pyogrio,... [communutils.cpp](https://github.com/OSGeo/gdal/blob/949ac934a6518f0e4fe60ca55827ec5e9d29d755/apps/commonutils.cpp#L69)?

theroggy:
Superseded by #270

pyogrio/raw.py
theroggy: When reviewing #270 I also noticed there is a bug in the python version of `GetOutputDriversFor` so I opened a ticket in GDAL to report this: https://github.com/OSGeo/gdal/issues/8277

PR #270, opened 2023-08-15
ENH: Add support for detecting write drivers using GDAL

### LINKED ISSUE #220: ENH: Auto-detect driver from filename extension when writing

brendan-ward:
Resolves #220
Supersedes #268 

This adds support to use GDAL directly to detect supported _write_ drivers based on a path, instead of using the hard-coded list we had previously.

This moves the `detect_driver` function to core; we may want to expose this as part of the public API for use in GeoPandas but I did not yet do so (let me know if this is a good idea).

As part of this, I had to rearrange a few things that are not part of the public API:
- `_preprocess_options_key_value` moved to `util.py` to avoid a circular import issue; it is more utility-oriented anyway
- `DRIVERS` was moved to the testing side, which is the only other place we were using it directly besides extension to driver detection.  This wasn't part of the official public API, so hopefully this is not a breaking change, but we can mention it in the changelog if you think appropriate.


pyogrio/_ogr.pyx
```diff
@@ -323,3 +323,34 @@ def _get_driver_metadata_item(driver, metadata_item):
             metadata = None
 
     return metadata
+
+
+def _get_drivers_for_path(path):
+    cdef OGRSFDriverH driver = NULL
+    cdef int i
+    cdef char *name_c
+
```
theroggy, line 332: Maybe nice to also add support for ".shp.zip" and ".gpkg.zip" as it is also supported in [gdal commonutils](https://github.com/OSGeo/gdal/blob/949ac934a6518f0e4fe60ca55827ec5e9d29d755/apps/commonutils.cpp#L69)?

theroggy:
Looks nice!

I'm doing some driver-specific stuff when writing in [geofileops](https://github.com/geofileops/geofileops), so being able to reuse the write-driver-detection would be useful to me at least. The `detect_driver` API is also pretty straightforward, so I also think it won't be difficult to keep it stable. Just not sure if it wouldn't be better to make it explicit in the function name that it will only return writable drivers?
Conclusion: exposing it as a public function sounds like a good idea to me, possibly with a slightly different name.

Probably it will be a good idea to revesit `DRIVERS` vs `DRIVERS_EXT` vs `ALL_EXTS` once upon a time now they are all only supporting the tests... but that's definitely not something to do in this PR.

pyogrio/_ogr.pyx
```diff
@@ -324,27 +324,39 @@ def _get_drivers_for_path(path):
...
+            continue
+
+        # extensions is a space-delimited list of supported extensions
+        # for driver
+        extensions = _get_driver_metadata_item(name, "DMD_EXTENSIONS")
+        if ext is not None and extensions is not None and ext in extensions.lower().split(' '):
+            drivers.append(name)
+        else:
+            prefix = _get_driver_metadata_item(name, "DMD_CONNECTION_PREFIX")
+            if prefix is not None and str(path).lower().startswith(prefix.lower()):
```
theroggy, line 359:
```suggestion
            if prefix is not None and path.startswith(prefix.lower()):
```
Path has already been cast to string + lowered at the beginning of function due to your latest change.

theroggy (approved):

Issue #262, opened 2023-08-10
Difference in way date columns are returned by pyarrow=True or False 


theroggy:
If pyarrow=False, the default, a date column is returned as dtype "datetime64".

If pyarrow=True, a date column is returned as dtype "object" with datetime.date objects as data values.

Not sure what is the way to go, but probably it would be better if the behaviour is the same.

theroggy:
Additional complication: columns of type "datetime64" are written by pyogrio to a column of Date type (tested with shapefile), while "object" columns with datetime.date values seem to be written as a "String" column.

theroggy:
Script to reproduce both aspects:

```
from pathlib import Path
import tempfile

import pyogrio

tmp_dir = Path(tempfile.gettempdir())
url = "https://github.com/theroggy/pysnippets/raw/main/pysnippets/pyogrio/polygon-parcel_31370.zip"
for use_arrow in [True, False]:
    gdf = pyogrio.read_dataframe(url, use_arrow=use_arrow)
    print(
        f"input_file, use_arrow: {use_arrow}, column dtype: {gdf['DATUM'].dtype}, "
        f"value: {gdf['DATUM'][0]}, value type: {type(gdf['DATUM'][0])}"
    )
    written_path = tmp_dir / f"output_{use_arrow}.shp"
    pyogrio.write_dataframe(gdf, written_path)
    gdf = pyogrio.read_dataframe(written_path, use_arrow=False)
    print(
        f"written_file, use_arrow: {use_arrow}, column dtype: {gdf['DATUM'].dtype}, "
        f"value: {gdf['DATUM'][0]}, value type: {type(gdf['DATUM'][0])}"
    )
```
(relevant) output:
```
input_file, use_arrow: True, column dtype: object, value: 2020-05-01, value type: <class 'datetime.date'>
written_file, use_arrow: True, column dtype: object, value: 2020-05-01, value type: <class 'str'>
input_file, use_arrow: False, column dtype: datetime64[s], value: 2020-05-01 00:00:00, value type: <class 'pandas._libs.tslibs.timestamps.Timestamp'>
written_file, use_arrow: False, column dtype: datetime64[s], value: 2020-05-01 00:00:00, value type: <class 'pandas._libs.tslibs.timestamps.Timestamp'>
```

theroggy:
In [geofileops](https://github.com/geofileops/geofileops) I'm thinking about dealing with it as follows, so convert these columns to datetime64:
```
[...]

PR #273, opened 2023-08-17
Add param arrow_to_pandas_kwargs to read_dataframe + decrease memory usage

### LINKED ISSUE #241: Load GeoDataFrame with arrow attribute dtypes?
Now that Pandas 2.0 is out, it would seem to be faster if pyogrio created a geodataframe with arrow dtypes, since they could be zero-copy from what GDAL emits. Presumably this should be opt-in for backwards compatibility?

theroggy:
* Add option to pass parameters to `to_pandas` to influence e.g. the way data is returned.
* Based on the `arrow` documentation, passing split_blocks=True and self_destruct=True should decrease peak memory usage of `to_pandas` in some cases. In read_dataframe it seems more logical to use those defaults:
https://arrow.apache.org/docs/python/pandas.html#reducing-memory-use-in-table-to-pandas

Related to #262 and resolves #241


jorisvandenbossche:
There are a lot of options in the arrow -> pandas `to_pandas` conversion that users might want to tweak. So maybe a more general solution is to provide a way to pass through keywords for `to_pandas`.

For this `split_blocks` keyword explicitly, I would prefer to follow the default of pyarrow (maybe pyarrow should consider changing that default, though). 

The `self_destruct` should indeed be possible to always call. Although I wonder how much difference that makes, since the `table` variable goes out of scope anyway when leaving the function, so I would expect that this should get cleaned-up that way.

theroggy:
I did a quick test to see the impact of this change on the peak memory usage of read_dataframe... I restarted the python process between each test, to avoid any influence of the order I ran the tests.

Obviously it is just one specific case, so not sure if it is comparable for other files, but it is better to have one test than no test :-).

I used the following script/file, as it was the motive to make the change: the script crashed with memory errors on my laptop. I ran the test on a real computer.
The file read is an openstreetmap (.pbf) file of 540 MB with 5.732.130 rows and 26 columns.
``` python
import psutil
import pyogrio

url = "https://download.geofabrik.de/europe/germany/baden-wuerttemberg-latest.osm.pbf"
pgdf = pyogrio.read_dataframe(url, use_arrow=True, sql="SELECT * FROM multipolygons")
print(psutil.Process().memory_info())
```

Results:
```
# split_blocks=True, self_destruct=True: peak_wset=9.296.629.760 -> 9,3 GB
# split_blocks=False, self_destruct=True: peak_wset=10430320640 -> 10,4 GB
# split_blocks=False, self_destruct=False: peak_wset=12530679808 -> 12,5 GB
```
@jorisvandenbossche do you see any disadvantages to use `split_blocks=True`, as it does seem to make a measurable difference in peak memory usage.

theroggy:
> There are a lot of options in the arrow -> pandas `to_pandas` conversion that users might want to tweak. So maybe a more general solution is to provide a way to pass through keywords for `to_pandas`.

Something like a an extra parameter for `read_dataframe` called e.g. `arrow_to_pandas_kwargs: Dict[str, Any]` we can pass to the `to_pandas` call or ?

theroggy:
Related to #262

pyogrio/geopandas.py
```diff
@@ -127,10 +128,12 @@ def read_dataframe(
     fid_as_index : bool, optional (default: False)
         If True, will use the FIDs of the features that were read as the
         index of the GeoDataFrame.  May start at 0 or 1 depending on the driver.
-    use_arrow : bool, default False
+    use_arrow : bool, optional (default: False)
         Whether to use Arrow as the transfer mechanism of the read data
         from GDAL to Python (requires GDAL >= 3.6 and `pyarrow` to be
         installed). When enabled, this provides a further speed-up.
+    arrow_to_pandas_kwargs: dict, optional (default: None)
```
jorisvandenbossche, line 135:
```suggestion
    arrow_to_pandas_kwargs : dict, optional (default: None)
```

pyogrio/geopandas.py
```diff
@@ -127,10 +128,12 @@ def read_dataframe(
...
+        When use_arrow is True, these kwargs will be passed to the _TO_PANDAS call.
```
jorisvandenbossche, line 136:
```suggestion
        When use_arrow is True, these kwargs will be passed to the ``to_table()`` call for the arrow to pandas conversion .
```

jorisvandenbossche (commented):
Sorry for the late reply, and thanks for the update!

> do you see any disadvantages to use split_blocks=True, as it does seem to make a measurable difference in peak memory usage.

For typical usage, I don't expect much difference, but with many columns there can be a benefit of having consolidated columns in pandas (so for actually benchmarking the impact, you also need to consider potential follow-up operations on the pandas DataFrame ..). Now I personally think we should consider switching the default, but I would prefer to follow pyarrow on this for consistency, and see if we we want to change this on the pyarrow side.

theroggy:
> Sorry for the late reply, and thanks for the update!
> 
> > do you see any disadvantages to use split_blocks=True, as it does seem to make a measurable difference in peak memory usage.
> 
> For typical usage, I don't expect much difference, but with many columns there can be a benefit of having consolidated columns in pandas (so for actually benchmarking the impact, you also need to consider potential follow-up operations on the pandas DataFrame ..). Now I personally think we should consider switching the default, but I would prefer to follow pyarrow on this for consistency, and see if we we want to change this on the pyarrow side.

OK

pyogrio/geopandas.py
theroggy: The method is called [to_pandas()](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas). I gave it a quick try, but it seems that a link cannot be visualized like this ``      ``.

jorisvandenbossche: Whoops, sorry, that was meant to be `to_pandas()` (I mostly wanted to comment that we should not use `_TO_PANDAS`)

pyogrio/geopandas.py
```diff
@@ -130,10 +131,13 @@ def read_dataframe(
     fid_as_index : bool, optional (default: False)
         If True, will use the FIDs of the features that were read as the
         index of the GeoDataFrame.  May start at 0 or 1 depending on the driver.
-    use_arrow : bool, default False
+    use_arrow : bool, optional (default: False)
         Whether to use Arrow as the transfer mechanism of the read data
         from GDAL to Python (requires GDAL >= 3.6 and `pyarrow` to be
         installed). When enabled, this provides a further speed-up.
+    arrow_to_pandas_kwargs : dict, optional (default: None)
+        When use_arrow is True, these kwargs will be passed to the `to_pandas`_ ()
```
jorisvandenbossche, line 139:
```suggestion
        When `use_arrow` is True, these kwargs will be passed to the `to_pandas`_
```

jorisvandenbossche (approved):
Thanks, looks good to me!

cc @brendan-ward are you ok with the `arrow_to_pandas_kwargs`? It's quite long, but explicit

brendan-ward (approved):
Thanks @theroggy !

`arrow_to_pandas_kwargs` is fine by me.

PR #282, opened 2023-09-05
ENH: Add support for skip_features, max_features for read_arrow

brendan-ward:
Progress toward feature parity between Arrow and non-Arrow read interfaces.

This adds support for `skip_features` and `max_features` to `read_arrow`, which enables these to be passed through via `read_dataframe(path, use_arrow=True, skip_features=10, max_features=2)` so that the behavior is the same with and without `use_arrow`.

I added a note to the introduction to describe the overhead involved:`use_arrow` is `True`, `skip_features` and `max_features` will incur additional overhead because all features up to the next batch size above `max_features` (or size of data layer) will be read prior to slicing out the requested range of features. If `max_features` is less than the maximum Arrow batch size (65,536 features) only `max_features` will be read. All features up to `skip_features` are read from the data source and later discarded because the Arrow interface does not support randomly seeking a starting feature.

This overhead is relative to reading via Arrow; based on my limited tests so far, it is still generally a lot faster to use Arrow even with these parameters than without Arrow.

This also drops a validation requirement that `skip_features` be less than the number of features available to read (originally we raised a `ValueError`).  Since using a `.slice` on a pyarrow Table with a value larger than the size of the original table happily returns an empty table, it made sense to take this approach throughout: if you ask for more features than available, you get back empty arrays / pyarrow Tables / (Geo)DataFrames.

### DIFF
--- a/pyogrio/raw.py
+++ b/pyogrio/raw.py
@@ -196,6 +196,17 @@ def read_arrow(
             "geometry_name": "<name of geometry column in arrow table>",
         }
     """
+    from pyarrow import Table
+
+    # limit batch size to max_features if set
+    if "batch_size" in kwargs:
+        batch_size = kwargs.pop("batch_size")
+    else:
+        batch_size = 65_536
+
+    if max_features is not None and max_features < batch_size:
+        batch_size = max_features
+
     with open_arrow(
         path_or_buffer,
         layer=layer,
@@ -203,18 +214,45 @@ def read_arrow(
         columns=columns,
         read_geometry=read_geometry,
         force_2d=force_2d,
-        skip_features=skip_features,
-        max_features=max_features,
         where=where,
         bbox=bbox,
         fids=fids,
         sql=sql,
         sql_dialect=sql_dialect,
         return_fids=return_fids,
+        batch_size=batch_size,
         **kwargs,
     ) as source:
         meta, reader = source
-        table = reader.read_all()
+
+        if max_features is not None:
+            batches = []
+            count = 0
+            while True:
+                try:
+                    batch = reader.read_next_batch()
+                    batches.append(batch)
+
+                    count += len(batch)
+                    if count > max_features:
+                        break
+
+                except StopIteration:
+                    break
+
+            # use combine_chunks to release the original memory that included
+            # too many features
+            table = (
+                Table.from_batches(batches)
+                .slice(skip_features, max_features)
+                .combine_chunks()
+            )
+
+        elif skip_features > 0:
+            table = reader.read_all().slice(skip_features).combine_chunks()
+
+        else:
+            table = reader.read_all()
 
     return meta, table
 
--- a/pyogrio/raw.py
+++ b/pyogrio/raw.py
@@ -242,6 +242,12 @@ def read_arrow(
     if max_features is not None and max_features < batch_size:
         batch_size = max_features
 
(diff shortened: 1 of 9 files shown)


theroggy:
I think this PR adds support for this in gdal 3.8: https://github.com/OSGeo/gdal/pull/8306

jorisvandenbossche:
Specifically one commit (https://github.com/OSGeo/gdal/pull/8306/commits/248cf600ff40dbd2cccc5f4f74a8a8e82d5f804e): " OGRLayer::GetArrowStream(): do not issue ResetReading() at beginning of iteration, but at end instead, so SetNextByIndex() can be honoured". 
That's nice! (although of course will still take some time before we can use in released version, but we can already start using dependent on the GDAL version if we want)

jorisvandenbossche:
> This also drops a validation requirement that `skip_features` be less than the number of features available to read

+1

jorisvandenbossche:
Something else: if `max_features` is small, you can set the `MAX_FEATURES_IN_BATCH` option to avoid to get too much data from GDAL. We still need the current code to iterate until we have enough data and slice the last batch, because it is not guaranteed that that option is honored (I think it can depend on the driver)

pyogrio/raw.py
```diff
@@ -231,25 +231,63 @@ def read_arrow(
...
+        if max_features is not None:
+            batches = []
+            count = 0
+            while True:
+                try:
+                    batch = reader.read_next_batch()
+                    batches.append(batch)
+
+                    count += len(batch)
+                    if count > max_features:
```
jorisvandenbossche, line 272:
```suggestion
                    if count >= (skip_features + max_features):
```

`max_features` is on top of the ones that are skipped, right?

pyogrio/raw.py
```diff
@@ -231,25 +231,63 @@ def read_arrow(
...
+            # use combine_chunks to release the original memory that included
+            # too many features
+            table = (
+                Table.from_batches(batches)
+                .slice(skip_features, max_features)
+                .combine_chunks()
+            )
+
+        elif skip_features > 0:
+            table = reader.read_all().slice(skip_features).combine_chunks()
```
jorisvandenbossche, line 287: Side thought: we should have a better way to release the sliced memory, without having to copy the whole table (and with "we", I mean pyarrow should enable this). 
Because right now, assume you have a million rows, and you skip the first 10,000, then only the first batch is sliced, and we unnecessarily copy all the other data.

pyogrio/tests/test_arrow.py
```diff
@@ -31,6 +31,20 @@ def test_read_arrow(naturalearth_lowres_all_ext):
...
+    table = read_arrow(naturalearth_lowres, skip_features=skip_features)[1]
+    assert len(table) == expected
+
+
+@pytest.mark.parametrize(
+    "max_features, expected", [(0, 0), (10, 10), (200, 177), (100000, 177)]
+)
+def test_read_arrow_max_features(naturalearth_lowres, max_features, expected):
+    table = read_arrow(naturalearth_lowres, max_features=max_features)[1]
+    assert len(table) == expected
```
jorisvandenbossche, line 45: Can you also add a test for a combination of skip and max features?

pyogrio/raw.py
brendan-ward: Good catch!

jorisvandenbossche (approved):

PR #284, opened 2023-09-05
FIX: Raise error from GDALClose after write

### LINKED ISSUE #221: Ensure to catch and propagate errors when calling GDALClose
See https://github.com/Toblerity/Fiona/issues/1169 (and fix in fiona: https://github.com/Toblerity/Fiona/commit/04ad6c8960c84ee6a2d29acd9e35e79bde193750) for context. 

I think we have the same issue as fiona that we are not checking for errors when "closing" the dataset when writing, while the writing can still go wrong at that point, and so this wouldn't bubble up to users.

brendan-ward:
Resolves #221

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,8 +18,8 @@
     `use_arrow=True` (#265)
 -   Fix errors reading OSM data due to invalid feature count and incorrect
     reading of OSM layers beyond the first layer (#271)
--   Raise an exception if a data source could not be written when error was not
-    raised by GDAL until closing the data source
+-   Always raise an exception if there is an error when writing a data source
+    (#284)
 
 ## 0.6.0 (2023-04-27)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@
     `use_arrow=True` (#265)
 -   Fix errors reading OSM data due to invalid feature count and incorrect
     reading of OSM layers beyond the first layer (#271)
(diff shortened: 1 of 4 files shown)


pyogrio/_io.pyx
```diff
@@ -1925,3 +1925,9 @@ def ogr_write(
     ### Final cleanup
     if ogr_dataset != NULL:
         GDALClose(ogr_dataset)
+
+        # GDAL will set an error if there was an error writing the data source
+        # on close (some operations in some drivers only happen
```
theroggy, line 1930: Sentence doesn't seem finished? 
```suggestion
        # In some cases, GDAL will only set an error on close when there was an error writing the data source
```

CHANGES.md
```diff
@@ -18,6 +18,8 @@
     `use_arrow=True` (#265)
 -   Fix errors reading OSM data due to invalid feature count and incorrect
     reading of OSM layers beyond the first layer (#271)
+-   Raise an exception if a data source could not be written when error was not
+    raised by GDAL until closing the data source
```
theroggy, line 22: Probably because I'm not a native english speaker, but it took me a few reads before I understood the sentence. Maybe this way it is easier to read?
```suggestion
-   When a datasource is written, also raise an exception if GDAL only gives an error when closing it (#284)
```

theroggy (approved):

theroggy (approved):

PR #285, opened 2023-09-06
ENH: Add support for mask parameter

### LINKED ISSUE #277: ENH: support `mask` keyword for compatibility with Fiona
Current plan is to use bounding box of mask when reading features with GDAL, then filter using shapely.

The issue is that OGR will give different results if GEOS is or is not compiled into GDAL.  Using shapely afterward ensures precise results.

brendan-ward:
Resolves #277

Adds `mask` parameter to `read`, `read_dataframe`, `open_arrow`, and `read_bounds`.

This takes an easier path to supporting geometries than Fiona (which supports GeoJSON-like inputs).  This requires Shapely `Geometry` objects, and internally converts them to WKB to create the geometry within GDAL.

I have not finished checking compatibility with other parameters, like `where` or `sql`.

I am currently seeing inconsistent behavior when using Arrow, where it will return more results than expected for GPKG.

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -198,6 +198,32 @@ will be returned; if GEOS is not available or not used by GDAL, all geometries
 with bounding boxes that intersect this bbox will be returned.
 `pyogrio.__gdal_geos_version__` will be `None` if GEOS is not detected.
 
+## Filter records by a geometry
+
+You can use the `mask` parameter to select only those features that intersect
+with a Shapely (>= 2.0) geometry.
+
+```python
+>>> mask = shapely.Polygon(([-80,8], [-80, 10], [-85,10], [-85,8], [-80,8]))
+>>> read_dataframe('ne_10m_admin_0_countries.shp', mask)
+```
+
+Note: the `mask` values must be in the same CRS as the dataset.
+
+If your mask geometry is in some other representation, such as GeoJSON, you will
+need to convert it to a Shapely geometry before using `mask`.
+
+```python
+>>> mask_geojson = '{"type":"Polygon","coordinates":[[[-80.0,8.0],[-80.0,10.0],[-85.0,10.0],[-85.0,8.0],[-80.0,8.0]]]}'
+>>> mask = shapely.from_geojson(mask_geojson)
+>>> read_dataframe('ne_10m_admin_0_countries.shp', mask)
+```
+
+Note: if GEOS is present and used by GDAL, only geometries that intersect `mask`
+will be returned; if GEOS is not available or not used by GDAL, all geometries
+with bounding boxes that intersect the bounding box of `mask` will be returned.
+`pyogrio.__gdal_geos_version__` will be `None` if GEOS is not detected.
+
 ## Execute a sql query
 
 You can use the `sql` parameter to execute a sql query on a dataset.
@@ -322,6 +348,7 @@ This function supports options to subset features from the dataset:
 -   `max_features`
 -   `where`
 -   `bbox`
+-   `mask`
 
 ## Write a GeoPandas GeoDataFrame
 
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -364,7 +364,7 @@ You can write a `GeoDataFrame` `df` to a file as follows:
 By default, the appropriate driver is inferred from the extension of the filename:
 
 -   `.fgb`: [FlatGeobuf](https://gdal.org/drivers/vector/flatgeobuf.html)
--   `.geojson`, `.json`: [GeoJSON](https://gdal.org/drivers/vector/geojson.html)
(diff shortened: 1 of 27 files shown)


brendan-ward:
GDAL Arrow bug logged at: https://github.com/OSGeo/gdal/issues/8347

This affects both `bbox` and `mask` filters when using Arrow for both GPKG and FlatGeobuf.

normanb:
@brendan-ward there has been quite a few commits to GDAL master regarding Arrow filtering. I have been using pyogrio with TileDB in GDAL (which supports arrow) and I am 👍 on adding mask support.

docs/source/introduction.md
```diff
@@ -200,6 +200,32 @@ will be returned; if GEOS is not available or not used by GDAL, all geometries
...
 `pyogrio.__gdal_geos_version__` will be `None` if GEOS is not detected.
 
+## Filter records by a geometry
+
+You can use the `mask` parameter to select only those features that intersect
+with a Shapely (>= 2.0) geometry.
+
+```python
+>>> mask = shapely.Polygon(([-80,8], [-80, 10], [-85,10], [-85,8], [-80,8]))
+>>> read_dataframe('ne_10m_admin_0_countries.shp', mask)
```
jorisvandenbossche, line 210:
```suggestion
>>> read_dataframe('ne_10m_admin_0_countries.shp', mask=mask)
```

docs/source/introduction.md
```diff
@@ -200,6 +200,32 @@ will be returned; if GEOS is not available or not used by GDAL, all geometries
...
+
+Note: the `mask` values must be in the same CRS as the dataset.
+
+If your mask geometry is in some other representation, such as GeoJSON, you will
+need to convert it to a Shapely geometry before using `mask`.
+
+```python
+>>> mask_geojson = '{"type":"Polygon","coordinates":[[[-80.0,8.0],[-80.0,10.0],[-85.0,10.0],[-85.0,8.0],[-80.0,8.0]]]}'
+>>> mask = shapely.from_geojson(mask_geojson)
+>>> read_dataframe('ne_10m_admin_0_countries.shp', mask)
```
jorisvandenbossche, line 221:
```suggestion
>>> read_dataframe('ne_10m_admin_0_countries.shp', mask=mask)
```

jorisvandenbossche (commented):
Looks good!

pyogrio/geopandas.py
```diff
@@ -97,13 +98,23 @@ def read_dataframe(
...
-        number of features usings FIDs is also driver specific.
+        with other keywords to select a subset (``skip_features``,
+        ``max_features``, ``where``, ``bbox``, ``mask``, or ``sql``). Note that
+        the starting index is driver and file specific (e.g. typically 0 for
+        Shapefile and 1 for GeoPackage, but can still depend on the specific
+        file). The performance of reading a large number of features usings FIDs
+        is also driver specific.
     sql : str, optional (default: None)
         The SQL statement to execute. Look at the sql_dialect parameter for
         more information on the syntax to use for the query. When combined
```
theroggy, line 120: Probably ``mask`` will behave the same as ``bbox`` in combination with ``sql``? Best to test it though + add here in doc?

Same for ``where``?

pyogrio/raw.py
```diff
@@ -94,13 +100,23 @@ def read(
...
     fids : array-like, optional (default: None)
         Array of integer feature id (FID) values to select. Cannot be combined
-        with other keywords to select a subset (`skip_features`, `max_features`,
-        `where` or `bbox`). Note that the starting index is driver and file
-        specific (e.g. typically 0 for Shapefile and 1 for GeoPackage, but can
-        still depend on the specific file). The performance of reading a large
-        number of features usings FIDs is also driver specific.
+        with other keywords to select a subset (``skip_features``,
+        ``max_features``, ``where``, ``bbox``, or ``mask``). Note that the
+        starting index is driver  and file specific (e.g. typically 0 for
```
theroggy, line 116:
```suggestion
        starting index is driver and file specific (e.g. typically 0 for
```

pyogrio/geopandas.py
brendan-ward: I added tests, but I'm not clear on what we should note in the docstring.  Do you mean state that you can combine `bbox` or `mask` with `sql` or `where` parameters?

theroggy: It's apparently not possible to add review comments in code/doc that hasn't been touched... but in the sql documentation `or mask` could  be added to the list of keywords in this listed keywords:
```
... When combined with other keywords like ``columns``, ``skip_features``, ``max_features``, ``where`` or ``bbox``, those are applied after the SQL query. ...
```

jorisvandenbossche (approved):

PR #286, opened 2023-09-06
TST: drop older GDAL, require shapely for tests, consolidate arrow test parametrize

brendan-ward:
This raises the minimum supported version of GDAL to 3.5.* (note: no compatibility change, just dropping old versions from CI).

This updates the CI matrix to use newer Docker images from ghcr.io (current home of GDAL images).

This adds `_compat.py` to check the state of optional dependencies and GDAL features like Arrow API or GEOS.  This will give us a place in later PRs to set variables to use Arrow by default.

This drops shim to use PyGEOS; otherwise the conditional presence of it was cluttering up tests esp. those around #285.  While GeoPandas support should continue to work with PyGEOS instead of Shapely 2.0, we no longer test that specifically and no longer list that as a compatible setup.  The tests assume that if GeoPandas is present that Shapely >= 2.0 is present.

This adds a new parametrize `@use_arrow` which will provide a boolean value of at least `False` and will include `True` if arrow support is enabled.  This should make it easier to mark additional tests to test with and without arrow support.

This adds new pytest marks to make skipping tests a little easier:
- `requires_arrow_api`: skip tests if pyarrow is absent or GDAL < 3.6
- `requires_gdal_geos`: skip tests if GDAL was not compiled with GEOS
- `requires_shapely`: skip tests if Shapely >= 2.0 is absent

This drops direct tests that use `.json` extension without specifying driver, because recent versions of GDAL added [another driver](https://gdal.org/drivers/vector/jsonfg.html) for that extension and we raise errors when there are multiple.  Since this comes from GDAL rather than an intentional breaking change on our end, I didn't note this in the changelog (should we?).

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,9 +4,9 @@ channels:
 dependencies:
   - numpy
   - geopandas-base
-  - libgdal ==3.6.2
+  - libgdal==3.6.2
   - pyarrow
-  - shapely
+  - shapely>=2
   # Specific for dev
   - cython
   - pre-commit
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - numpy
   - geopandas-base
-  - libgdal
+  - libgdal==3.6.2
   - pyarrow
   - shapely>=2
   # Specific for dev
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - numpy
   - geopandas-base
-  - libgdal==3.6.2
+  - libgdal
   - pyarrow
   - shapely>=2
   # Specific for dev
--- a/pyogrio/tests/conftest.py
+++ b/pyogrio/tests/conftest.py
@@ -4,22 +4,12 @@ from zipfile import ZipFile, ZIP_DEFLATED
 import pytest
 
 from pyogrio import (
-    __gdal_version__,
     __gdal_version_string__,
-    __gdal_geos_version__,
     __version__,
     list_drivers,
 )
+from pyogrio._compat import HAS_ARROW_API, HAS_GDAL_GEOS, HAS_SHAPELY
 from pyogrio.raw import read, write
-
-try:
-    import pyarrow
-except ImportError:
-    pyarrow = None
-
-
-HAS_ARROW = __gdal_version__ >= (3, 6, 0) and pyarrow is not None
-HAS_GEOS = __gdal_geos_version__ is not None
 
 
 _data_dir = Path(__file__).parent.resolve() / "fixtures"
@@ -50,6 +40,18 @@ def pytest_report_header(config):
         f"GDAL {__gdal_version_string__}\n"
         f"Supported drivers: {drivers}"
     )
+
+
+# marks to skip tests if optional dependecies are not present
+requires_arrow_api = pytest.mark.skipif(
+    not HAS_ARROW_API, reason="GDAL>=3.6 and pyarrow required"
+)
+
+requires_gdal_geos = pytest.mark.skipif(
+    not HAS_GDAL_GEOS, reason="GDAL compiled with GEOS required"
+)
+
+requires_shapely = pytest.mark.skipif(not HAS_SHAPELY, reason="Shapely >= 2.0 required")
 
 
 def prepare_testfile(testfile_path, dst_dir, ext):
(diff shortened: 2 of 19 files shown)


brendan-ward:
> Is there now a test environment available in the CI where gdal is compiled without geos?

Not that I am aware of; I think we'd have to compile GDAL from source to get that setup.  For #285, I opted to use the same approach as for `bbox`, which means it is limited by GDAL rather than an operation we implement on our end (using shapely).  We can sidestep the issue a bit by making sure that GDAL in our wheels is always compiled using GEOS.

theroggy:
> > Is there now a test environment available in the CI where gdal is compiled without geos?
> 
> Not that I am aware of; I think we'd have to compile GDAL from source to get that setup. For #285, I opted to use the same approach as for `bbox`, which means it is limited by GDAL rather than an operation we implement on our end (using shapely). We can sidestep the issue a bit by making sure that GDAL in our wheels is always compiled using GEOS.

Ah, OK, I hadn't "read" the PR yet. If GDAL is being used for the filtering and it is not planned to implement a fallback (for now) if GDAL isn't compiled with geos it is not as important to have such a test environment. 

If an testing environment with GDAL without geos support would be wanted, the alpine-small probably is an easy candidate as mentioned in this comment: https://github.com/geopandas/pyogrio/issues/277#issuecomment-1702612350

pyogrio/tests/conftest.py
```diff
@@ -16,7 +21,6 @@
     ".geojsonl": "GeoJSONSeq",
     ".geojsons": "GeoJSONSeq",
     ".gpkg": "GPKG",
-    ".json": "GeoJSON",
```
jorisvandenbossche, line 19: Was this done just to reduce the test parametrization size?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -23,25 +28,22 @@
...
-# Note: this will also be false for GDAL < 3.4 when GEOS may be present but we
-# cannot verify it
-has_geos = __gdal_geos_version__ is not None
+use_arrow = pytest.mark.parametrize(
+    "use_arrow",
+    [
+        False,
+        pytest.param(True, marks=requires_arrow_api),
+    ],
+)
```
jorisvandenbossche, line 46: You could also make this a parametrized fixture, and then the `@use_arrow` is not needed, just the `use_arrow` as argument to the test function (that's also a bit less explicit though)

```suggestion
@pytest.fixture(params=[False, pytest.param(True, marks=requires_arrow_api)])
def use_arrow(request):
    return request.param
```

jorisvandenbossche (approved):
Thanks a lot! This is a nice clean-up and sets the path for more arrow True/False parametrization.

One thing I am wondering: should we maybe be a bit more conservative with dropping older GDAL from testing? (I don't know how much of a bottleneck it is to keep one more build)  
For example, GDAL 3.4 is not yet 2 years old (3.4.0 released in November 2021). Or we could also test against one minimum version that we officially want to support, and not necessarily every version in between to keep the number of builds sane.

pyogrio/tests/conftest.py
brendan-ward: No, this was because newer versions of GDAL introduced additional drivers that use the `.json` extension, so we get an exception using it because there are multiple drivers available for it (which is what we want).  Seemed easier just to avoid that issue rather than adding version-specific handling.

brendan-ward:
I added 3.4 back in as minimum supported version; the one issue I wanted out of the way - detecting GEOS via GDAL - was limited to < 3.4, so this should be fine.  Just to stay safe from version-specific behavior, it is probably a good idea for us to continue testing all versions between our minimum and latest, and they are reasonably fast.

environment-dev.yml
```diff
@@ -4,9 +4,9 @@ channels:
 dependencies:
   - numpy
   - geopandas-base
-  - libgdal ==3.6.2
+  - libgdal==3.6.2
```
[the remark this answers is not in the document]
theroggy: It is not necessary, but on Windows the gdal version needs to be specified via an environment variable to build pyogrio. The version was pinned here so it is in sync with the line in the documentation that sets this environment variable: https://pyogrio.readthedocs.io/en/latest/install.html#conda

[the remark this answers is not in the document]
jorisvandenbossche: I would say that a contributor can also check the version of gdal (`conda list`) to write the correct env variable (and we can mention that in the docs), instead of pinning it here. And if pinning, we should update it to 3.7

PR #293, opened 2023-09-23
CI: Run nightly tests on installed version of pyogrio

theroggy:
Just an idea to do this... but it can help to detect issues as early as possible when dependencies make breaking changes and/or have regression bugs that influence pyogrio.


theroggy:
As the binding on new versions of GDAL is explicit as I learned since... this isn't that interesting, so closing this.

PR #299, opened 2023-09-29
ENH Remove warning on no features

### LINKED ISSUE #294: Stop warning about no features to read
We currently issue a warning if there are no features to read from a data source unless you are using Arrow, so this is a point of inconsistency.

In practice this warning is likely not very helpful, so we should simply remove it.

brendan-ward:
Resolves #294 

Removes the `UserWarning` warning previously raised when no features were read from the data source.

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -632,7 +632,6 @@ cdef validate_feature_range(OGRLayerH ogr_layer, int skip_features=0, int max_fe
     num_features = max_features
 
     if feature_count == 0:
-        name = OGR_L_GetName(ogr_layer)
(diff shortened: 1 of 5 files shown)


pyogrio/_io.pyx
```diff
@@ -633,7 +633,6 @@ cdef validate_feature_range(OGRLayerH ogr_layer, int skip_features=0, int max_fe
 
     if feature_count == 0:
         name = OGR_L_GetName(ogr_layer)
```
jorisvandenbossche, line 635:
```suggestion
```

jorisvandenbossche (approved):
Looks good!

PR #306, opened 2023-10-02
TST: add use_arrow to geopandas tests with sql statement

theroggy:


pyogrio/_io.pyx
```diff
@@ -1353,16 +1353,16 @@ def ogr_open_arrow(
...
-
-        stream_ptr = <uintptr_t> &stream
-
         if skip_features:
-            # only supported for GDAL >= 3.8.0; have to do this after getting
+            # only supported for GDAL >= 3.8.0; have to do this before getting
             # the Arrow stream
             OGR_L_SetNextByIndex(ogr_layer, skip_features)
 
+        if not OGR_L_GetArrowStream(ogr_layer, &stream, options):
```
brendan-ward, line 1361: Based on my testing when implementing this, this line needed to come before `OGR_L_SetNextByIndex` or it didn't set the index properly.

brendan-ward (commented):
Thanks for working on this @theroggy - 👍 on more arrow tests.

The only issue I'm seeing is the order of operations around getting the arrow stream (has to happen first) vs setting the index of the next feature to read.  Did you find that you needed to change that order for the sql tests to work?

pyogrio/_io.pyx
theroggy: I apparently forgot to hit "comment" to save a comment about it.

It seems that one specific case in the test to filter using a sql statement + skip_features doesn't work for the `use_arrow=True` gdal >= 3.8 code path: the test where 1 row is returned by the sql statement and `skip_features=1`, so the result should be 0 rows but it is 1.

To understand better wat is happening in de gdal side for this case I had another quick look at the fix in gdal>=3.8 that enables using `skip_features` (https://github.com/OSGeo/gdal/pull/8306/files), and it basically only removes a `ResetReading` call in `OpenArrowStream` and the tests in gdal that were added in that PR also first set the `SetNextByIndex` and then do  `OpenArrowStream`.

So I tried if changing the order helped, but it gives the same result: the same test case still fails.

brendan-ward: Does this indicate a bug in the Arrow API in GDAL?

theroggy: I need to have another look at the gdal code to be sure, but in the failing test the `LIMIT` clause is used in the SQL statement, which the SQL equivalent of `max_features`, while `OFFSET` is equivalent of `skip_features`. Because it is a bit weird from a usage perspective to apply windowing in two ways simultaneously, I rewrote the SQL statement so it still returns 1 row, but without using a windowing function, and then the test passes.

So I think it is not (fully) supported to use both windowing clauses in the SQL statement and `skip_features`/`max_features`, but I need to test some more and read some more gdal to be sure.

I also saw that the gdal PR above also includes an optimization for GPKG to implement `skip_features` and `max_features` by appending LIMIT ... OFFSET ... to the SQL query as such, so for GPKG I'm quite sure this combo will give issues, but some more testing needed there as well as I rather expected an SQL error than a wrong result.

theroggy: I did some more tests... Not really practical without a local gdal 3.8 version, but I think I know more.

There seems to only be a minor difference in support of this feature in gdal 3.8 versus before: only whether you can put the setNextByID before OpenArrowStream or not.

The main problem is as I mentioned above that for OGRSQL sql statements (SQLITE dialect seems ok), that you cannot combine an OFFSET clause with skip_fearures. If you do, you get a wrong result.

Do for now I changed the code so that sql statements take the  local filtering code path.

I think its best to post an issue for gdal, but is a bit of a pain to create a reproducable code sample for this without using pyogrio...

theroggy: I reported the issue in the gdal issue tracker and has already been solved: [#8585](https://github.com/OSGeo/gdal/issues/8585)

It also got backported to 3.7... so will be fixed in >= 3.7.3 (once it is released).

theroggy:
Hmm... the test still fails on gdal 3.8. The PR was merged 4 days ago, is it possible the container used here doesn't contain the fix yet?

brendan-ward:
The GDAL latest container hasn't been updated in some time (last update 9/16): https://github.com/OSGeo/gdal/pkgs/container/gdal/128542179?tag=ubuntu-small-latest

I thought those would be built automatically on merge to GDAL `master`, but apparently not.

pyogrio/_io.pyx
```diff
@@ -1390,8 +1390,7 @@ def ogr_open_arrow(
         stream_ptr = <uintptr_t> &stream
 
         if skip_features:
-            # only supported for GDAL >= 3.8.0; have to do this after getting
-            # the Arrow stream
+            # For GDAL < 3.8.0 this needs to be set after getting the Arrow stream
```
brendan-ward, line 1393: I don't think this change to the comment is needed?  If the goal is to clarify the previous comment, we could use something like:

```
# skip_features is only supported for GDAL >= 3.8.0; otherwise this is responsibility of the caller
# need to set OGR_L_SetNextByIndex after getting the Arrow stream
```

brendan-ward (commented):
Apologies for the long delay in review!

It looks like this now passes on CI, so other than updating a comment, this looks ready to go.

pyogrio/_io.pyx
theroggy: If I recall correctly from when testing this, in GDAL < 3.8 there were 2 issues:

- the combination of a sql filter with a LIMIT clause + OGR_L_SetNextByIndex gave wrong results. This was fixed in gdal >= 3.8.
- OGR_L_SetNextByIndex had to be run after OGR_L_GetArrowStream because in OGR_L_GetArrowStream the index was reset to 0. This was removed in GDAL 3.8 so for gdal > 3.8 OGR_L_SetNextByIndex can be called before OGR_L_GetArrowStream and it will work.

Because `skip_features` is only enabled from GDAL >= 3.8 it all doesn't really matter... because in that version all bugs were fixed. So, I think the current comment isn't really accurate, but I suppose it is not really important.

theroggy: Anyway, I'll give it another try based on your new suggestion.

brendan-ward (approved):
Thanks @theroggy !

Issue #309, opened 2023-10-09
BUG: Handle negative skip_features


brendan-ward:
Negative values of `skip_features` results in skipping all records, which is not intuitive.

We need to decide if we want to raise an error when `skip_features` is a negative value, or if we want to interpret it as a negative slice.

The variable name doesn't lend itself nicely to implying negative slicing compared to something like `start_index`.

Crossref [geopandas #3039](https://github.com/geopandas/geopandas/pull/3039)

PR #312, opened 2023-10-10
FIX: Disallow negative skip_features

brendan-ward:
This sidesteps making a decision around implementing support for negative `skip_features` in #309 for now and simply disallows that, which would have been equivalent to the behavior prior to #282.  We can add support for negative slicing at a different time, if we so choose.

This raises a specific `ValueError` and expands tests to match.  I also added checks for negative `max_features` and associated tests.

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -490,7 +490,7 @@ def test_read_max_features(naturalearth_lowres_all_ext, use_arrow, max_features)
     )
 
 
-def testt_read_negative_max_features(naturalearth_lowres, use_arrow):
+def test_read_negative_max_features(naturalearth_lowres, use_arrow):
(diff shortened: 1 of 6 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -429,15 +429,70 @@ def test_read_fids_force_2d(test_fgdb_vsi):
...
+        df,
+        expected,
+        check_less_precise=is_json,
+        check_index_type=False,
+        check_dtype=not is_json,
+        normalize=is_jsons,
+    )
+
+
+def testt_read_negative_max_features(naturalearth_lowres, use_arrow):
```
jorisvandenbossche, line 493:
```suggestion
def test_read_negative_max_features(naturalearth_lowres, use_arrow):
```

Issue #310, opened 2023-10-09
Support for forcing driver in `read_dataframe`


kylebarron:
I have some CSV data that uses a file extension that is not `.csv`. It appears that the OGR CSV driver only auto-detects CSV data if the extension is `.csv`. Is there a way to force the use of a specific driver when not auto-detected? It appears there is not already. This would be akin to [`driver` in `fiona.open`](https://fiona.readthedocs.io/en/stable/fiona.html#fiona.open).

brendan-ward:
It looks like Fiona uses the `enabled_drivers` to restrict the drivers that are checked by GDAL when opening a file for reading (not `driver`, which is used for writing).  It is unclear from the example in the docstring if it actually allows overriding the driver used to open the file, or if it simply constraints the set of drivers checked - which would still depend on file extension.

In pyogrio we have it enable all drivers at startup; I'm not sure we want to selectively enable some of them.

The GDAL Python bindings let you first obtain [a driver instance and then use that to open a path](https://gdal.org/api/python/osgeo.ogr.html#osgeo.ogr.Driver.Open), so it may provide a better example to check.

theroggy:
Have you tried this, from the [GDAL CSV driver documentation](https://gdal.org/drivers/vector/csv.html)?

> For files structured as CSV, but not ending with .CSV extension, the 'CSV:' prefix can be added before the filename to force loading by the CSV driver.

kylebarron:
Ooo... the driver documentation page is very long and I missed that. But yes that does work!

I'm happy to close this or keep it open in case forcing a driver is useful in other cases

brendan-ward:
We can close until there is an urgent need to add that capability for other drivers.

theroggy:
This works for all drivers already... the prefix to use is defined with the GDAL_DMD_CONNECTION_PREFIX metadata item of each driver, e.g.:
```
poDriver->SetMetadataItem(GDAL_DMD_CONNECTION_PREFIX, "MSSQL:");
```

PR #313, opened 2023-10-11
DOC: mention option of prefixing a path with the driver in ogr_open exception 

theroggy:
As most drivers seem to support this, it might be practical to suggest this option when a file cannot be opened?

For writing, the driver detection doesn't support this at the moment, and gdal doesn't seem to support this either for writing... so where it is probably easier to just keep using the `driver` parameter.

reference #310

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -161,6 +161,11 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         raise DataSourceError("Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))) from None
 
     except CPLE_BaseError as exc:
+        if str(exc).endswith("not recognized as a supported file format."):
+            raise DataSourceError(
+                f"{str(exc)} It might help to specify the correct driver explicitly by "
+                "prefixing the file path with `<DRIVER>:`"
+            )
         raise DataSourceError(str(exc))
 
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -164,7 +164,7 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         if str(exc).endswith("not recognized as a supported file format."):
             raise DataSourceError(
                 f"{str(exc)} It might help to specify the correct driver explicitly by "
-                "prefixing the file path with `<DRIVER>:`"
+                "prefixing the file path with '<DRIVER>:', e.g. 'GPKG:path'."
             )
         raise DataSourceError(str(exc))
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -166,7 +166,7 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         if str(exc).endswith("not recognized as a supported file format."):
             raise DataSourceError(
                 f"{str(exc)} It might help to specify the correct driver explicitly by "
-                "prefixing the file path with '<DRIVER>:', e.g. 'GPKG:path'."
+                "prefixing the file path with '<DRIVER>:', e.g. 'CSV:path'."
             ) from None
         raise DataSourceError(str(exc)) from None
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -158,15 +158,17 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         return ogr_dataset
 
     except NullPointerError:
-        raise DataSourceError("Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))) from None
+        raise DataSourceError(
+            "Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))
+        ) from None
 
     except CPLE_BaseError as exc:
         if str(exc).endswith("not recognized as a supported file format."):
             raise DataSourceError(
                 f"{str(exc)} It might help to specify the correct driver explicitly by "
                 "prefixing the file path with '<DRIVER>:', e.g. 'GPKG:path'."
-            )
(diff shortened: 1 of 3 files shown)


pyogrio/_io.pyx
```diff
@@ -161,6 +161,11 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         raise DataSourceError("Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))) from None
 
     except CPLE_BaseError as exc:
+        if str(exc).endswith("not recognized as a supported file format."):
+            raise DataSourceError(
+                f"{str(exc)} It might help to specify the correct driver explicitly by "
+                "prefixing the file path with `<DRIVER>:`"
```
martinfleis, line 167:
```suggestion
                "prefixing the file path with `<DRIVER>:`, e.g., `GPKG:path"
```
Giving an example may help. I can imagine people just copy pasting `<DRIVER>:`.

theroggy: Good idea!

pyogrio/tests/test_raw_io.py
```diff
@@ -906,6 +906,25 @@ def test_read_datetime_millisecond(test_datetime):
     assert field[1] == np.datetime64("2020-01-01 10:00:00.000")
 
 
+@pytest.mark.parametrize("prefix_driver", [True, False])
+def test_read_unsupported_ext(tmp_path, prefix_driver):
```
brendan-ward, line 910: I'd suggest splitting this into 2 test cases, since the logic is a bit different:

`test_read_unsupported_ext` (the error case)
`test_read_unsupported_ext_with_prefix`

theroggy: Sure, I first wrote it like that, but merged them to reuse the test data and make it clearer that with the same data it works with prefixing versus without.

But it doesn't really matter...

pyogrio/_io.pyx
```diff
@@ -161,6 +161,11 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
         raise DataSourceError("Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))) from None
 
     except CPLE_BaseError as exc:
+        if str(exc).endswith("not recognized as a supported file format."):
+            raise DataSourceError(
+                f"{str(exc)} It might help to specify the correct driver explicitly by "
+                "prefixing the file path with '<DRIVER>:', e.g. 'GPKG:path'."
+            )
```
brendan-ward, line 168:
```suggestion
            ) from None
```

We appear to be getting nested exceptions here.  Can you also please add this to line 169:

`raise DataSourceError(str(exc)) from None`

brendan-ward (commented):
Thanks @theroggy !

From a bit of testing, it looks like the `FGB:` prefix doesn't work properly.  FGB files seem to be detected correctly regardless of file extension, but adding the `FGB:` prefix causes us to return a file not found error.

Most of the others return some sort of error (varying by type) if you specify the wrong type in the prefix.

Given that, I wonder if it might be good to mention this in the introduction.  Maybe a new subsection under reading

```
## Read without a supported extension

GDAL typically uses the file extension to infer an appropriate driver to use when reading the file.  If you have a file path for a file format that is supported by your version of GDAL that is lacking this extension, GDAL may return an error about your file not being a recognized file format.

For some drivers, you may be able to get around this by adding a driver prefix to the path:
...
```

And then direct folks to check the documentation for their specific format to determine if the prefix is supported.  Unfortunately it doesn't appear these are all documented, e.g., GPKG doesn't mention this.

theroggy:
Yes, it is supported for quite some formats, but definitely not for all, so it is definitely not a good idea to advice the user to structurally specify the driver this way...

While testing some things I also noticed that for some formats gdal also does some automatic detection even if the extension isn't correct... so I think it isn't easy to give advice beyond "if reading a file doesn't seem to work, try specifying the driver like this."

This is why I thought it was a good idea to give the advice in the exception, then the user is in a situation that it is useful to try if he can solve his problem by specifying the driver like this.

Obviously this doesn't mean it cannot be useful to mention this in the documentation somewhere as well. I might be wrong about this, but I interpreted the "introduction" rather like a quick-start. If this is a correct interpretation, I'm not sure if it is the ideal place to document something 99.9% of users probably won't ever encounter?
Maybe the [error handling page](https://pyogrio.readthedocs.io/en/latest/errors.html) could be renamed to "Troubleshooting" and some info about this could be added there?

brendan-ward:
> so it is definitely not a good idea to advice the user to structurally specify the driver this way

Good point.  Let's leave this out of the documentation for now until there is additional demand / confusion around this; having it in the error handler only seems reasonable for now.

brendan-ward (approved):

pyogrio/_io.pyx
```diff
@@ -158,10 +158,17 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
+        raise DataSourceError(
+            "Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))
+        ) from None
 
     except CPLE_BaseError as exc:
-        raise DataSourceError(str(exc))
+        if str(exc).endswith("not recognized as a supported file format."):
+            raise DataSourceError(
+                f"{str(exc)} It might help to specify the correct driver explicitly by "
+                "prefixing the file path with '<DRIVER>:', e.g. 'GPKG:path'."
```
jorisvandenbossche, line 169:
```suggestion
                "prefixing the file path with '<DRIVER>:', e.g. 'CSV:path'."
```

Let's maybe take an example where it matters? (from reading the other comments, it seems that for specifically GPKG this is not documented on the GDAL side and potentially also not needed because GDAL recognizes GPKG regardless of the extension?)

theroggy: True! Changed.

PR #323, opened 2023-10-31
MNT: Update release actions

### LINKED ISSUE #317: MNT: Update Github actions for Release
See warnings emitted at: https://github.com/geopandas/pyogrio/actions/runs/6644743658

brendan-ward:
Resolves #317

This updates the release creation / deployment actions to supported versions, and migrates to trusted publishing on PyPI.

I enabled trusted publishing on PyPI for this project.

Not quite sure how best we can test this other than creating alpha / beta releases via tags.  We could potentially temporarily switch this to point to the test PyPI server, and delete any releases this creates from Github.


martinfleis:
If it won't work, then we can always delete the tag, fix the action and try again.

brendan-ward:
Running into [issues with TestPyPI](https://github.com/pypi/warehouse/issues/14780) in trying to get setup to test uploading packages there instead in order to test this PR.  Looks like we'll need to wait for that to get fixed first.

martinfleis:
@brendan-ward can you check again?

brendan-ward:
Head's up: going to create (possibly multiple) 0.7.3a* tags to test upload of release artifacts to test pypi repository and upload of release artifacts to new versions on Github.  Will delete test tags / versions as part of working this out.

Apologies in advance for the noise from Github about this.

brendan-ward:
These updates successfully uploaded to the test PyPI repository and created a version (name had issues but that might have been due to running in a PR or config; will watch on next release).

Once this is merged, we can create a throw-away tag on `main` (maybe `v0.7.3a4`) to verify this works properly on the public PyPI repository.

.github/workflows/release.yml
```diff
@@ -248,45 +266,32 @@ jobs:
...
 
       - name: Create GitHub Release
         id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        uses: softprops/action-gh-release@v1
         with:
           tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
```
jorisvandenbossche, line 273: In the email notification I got from your test, the title was "Release v0.7.3a3 - refs/tags/v0.7.3a3", while before this was something like "Release v0.7.2 - v0.7.2". I am not sure what the second part (that is now with "refs/tags/..") is used for in the UI (since you deleted it). Ideally this would still just be the version name, but this is of course also something easily edited manually (since we need to edit the release anyway to add some release notes)

jorisvandenbossche (approved):
Looks good!

martinfleis (approved):

.github/workflows/release.yml
brendan-ward: In the UI, it named it the same as the email, which is indeed not ideal.  We changed to a different action for creating the release, so that may have something to do with it.  I'll do a bit more experimenting to see if we can get it to name it nicely; will be a bit noisy with release announcement emails though!

brendan-ward: Turns out I was pointing it at the wrong variable previously, which is why it started including `refs/tags` prefix.

It is easy to add "Version <tag>" as the release; it does not appear easy to either use the message set on the tag per our instructions (e.g., "Version 0.7.3a3") nor does it appear easy to strip the leading "v" from the tag in order to create the name.  So now our release names by default are e.g., "Version v0.7.3a3".

This at least makes the messages sent out by github better and the default name of the version better; we can always edit it to remove the "v" when adding the changes, if we are that annoyed about the "v".

brendan-ward:
Here's how it looks in the UI:
image

Note: it appears to show the tag message as the body by default until we set the body manually; this appears to be coming from Github automatically rather than from the release action.

Going to delete the release / tag because I don't want anyone to try to pull this.

.github/workflows/release.yml
```diff
@@ -248,45 +266,33 @@ jobs:
...
+        uses: pypa/gh-action-pypi-publish@release/v1
 
       - name: Create GitHub Release
         id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        uses: softprops/action-gh-release@v1
         with:
+          name: Version ${{ github.ref_name }}
```
jorisvandenbossche, line 293: If we don't want the duplicate "version" and "v", I think just using the ref_name without "Version " before it, would also be fine. Anyway, either way is good!

jorisvandenbossche (approved):

PR #325, opened 2023-11-01
FIX: Fix error in `write_dataframe` if input has a date column and non-consecutive index values

### LINKED ISSUE #324: BUG: error when writing a dataframe with a date column and non-consecutive index values
I updated a test as a reproducable test case:

```
def test_read_datetime_tz(test_datetime_tz, tmp_path):
    df = read_dataframe(test_datetime_tz)
    # Make the index non-consecutive to test this case as well
    df = df.set_index(np.array([0, 2]))
    raw_expected = ["2020-01-01T09:00:00.123-05:00", "2020-01-01T10:00:00-05:00"]

    if PANDAS_GE_20:
        expected = pd.to_datetime(raw_expected, format="ISO8601").as_unit("ms")
    else:
        expected = pd.to_datetime(raw_expected)
    expected = pd.Series(expected, name="datetime_col", index=non_consecutive_indices)
    assert_series_equal(df.datetime_col, expected, check_index=False)
    # test write and read round trips
    fpath = tmp_path / "test.gpkg"
    write_dataframe(df, fpath)
    df_read = read_dataframe(fpath)
    assert_series_equal(df_read.datetime_col, expected)
```

theroggy:
resolves #324

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 0.7.3 (???)
+
+### Bug fixes
+
+-   Fix error in `write_dataframe` if input has a date column and
+    non-consecutive index values (#325).
+
 ## 0.7.2 (2023-10-30)
 
 ### Bug fixes
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -439,6 +439,7 @@ def write_dataframe(
     # TODO: may need to fill in pd.NA, etc
     field_data = []
     field_mask = []
+    df = df.reset_index(drop=True)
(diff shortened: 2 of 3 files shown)


brendan-ward (changes requested):
Good catch on isolating the issue.  I think the real fix needs to be something like

`gdal_tz_offsets[name] = gdal_offset_representation` => `gdal_tz_offsets[name] = gdal_offset_representation.values`  on line 459 (geopandas.py)

So that we force it into a contiguous array where incremental indexing works.

theroggy:
That's indeed a cleaner solution... thanks.

brendan-ward (approved):
Thanks @theroggy !

PR #329, opened 2023-12-11
CI: use setup-micromamba action

martinfleis:
`provision-with-micromamba` has been deprecated for some time. `setup-micromamba` is the new action to be used.

### DIFF
--- a/.github/workflows/tests-conda.yml
+++ b/.github/workflows/tests-conda.yml
@@ -28,7 +28,7 @@ jobs:
         include:
           # environment with lower versions of optional dependencies
           - python: "3.10"
-            extra: |
+            extra: >-
               pandas=1.5
               geopandas=0.12
           # minimal environment without optional dependencies


martinfleis:
I got no clue why Python 3.10 does not work when loaded from matrix but does with the one `include` run...

theroggy:
I don't remember anymore why, but you need to use >- instead of | with create-args:
```
          create-args: >-
            python=${{ matrix.python }}
            ${{ matrix.extra }}
```
instead of 
```
          create-args: |
            python=${{ matrix.python }}
            ${{ matrix.extra }}
```

theroggy:
@martinfleis I think `python==${{ matrix.python }}` should be `python=${{ matrix.python }}` , so only one `=`?

martinfleis:
It required consistent use of `>-` everywhere in the yaml.

Now we have a segfault coming from reading GPKG with mask on macOS under Python 3.12. And that is something I will happily leave to someone else to look at as I know near nothing on how to fix that one.

martinfleis:
Solved by re-running the env. All green!

jorisvandenbossche:
And FWIW, for other people wondering what's the difference between `|` or `>-` -> https://yaml-multiline.info/ (essentially we needed a single line command here, instead of executing as multiple lines)

jorisvandenbossche (approved):

Issue #334, opened 2023-12-21
BUG: use_arrow=True changes boolean values when reading Geopackage


Lotuss91:
Hi, this is my first time submitting an issue on GitHub so I apologize in advance in case I do not follow the etiquette.

When reading a geopackage with (generated using `geopandas.to_file`) with `geopandas.read_file`, the the values of all the boolean columns are changed. The weird thing not all values of the columns are changes but most of them are.
Is it possible that the issue is related to the PyArraow own datatypes?

Below I provide the use case that generates the issue.
```python
# normal behaviour
basic = gpd.read_file(input_folder / file_name)
pyogr = gpd.read_file(input_folder / file_name, engine="pyogrio")
arrow = gpd.read_file(input_folder / file_name, use_arrow=True)

# boolean values changes
pyogr_arrow = gpd.read_file(input_folder / file_name, engine="pyogrio", use_arrow=True)
``` 

If useful, I can provide a sample file.

Thank You

martinfleis:
> the values of all the boolean columns are changed

Can you be more specific?

> If useful, I can provide a sample file.

If it is a small one, it may help, thanks!

Lotuss91:
Thank you for the quick reply.
By "the boolean columns are changed" I mean that several element (more than 50% but not 100%) of the columns are flipped from `False` to `True`(or viceversa)

I sent you a sample file to you e-mail.

martinfleis:
Here's a sample file attached.

```py
basic = pyogrio.read_dataframe("sample.gpkg.zip")
w_arrow =  pyogrio.read_dataframe("sample.gpkg.zip", use_arrow=True)
```

this results each in different values in the boolean column `foo`.

```
>>> basic
     id    foo     geometry
0    18  False  POINT (0 0)
1    48   True  POINT (0 0)
2    49   True  POINT (0 0)
3    50   True  POINT (0 0)
4    51   True  POINT (0 0)
..  ...    ...          ...
95  349   True  POINT (0 0)
96  350   True  POINT (0 0)
97  351   True  POINT (0 0)
98  352   True  POINT (0 0)
99  393   True  POINT (0 0)

[100 rows x 3 columns]

>>> w_arrow
     id    foo     geometry
0    18   True  POINT (0 0)
1    48  False  POINT (0 0)
2    49  False  POINT (0 0)
3    50  False  POINT (0 0)
4    51  False  POINT (0 0)
..  ...    ...          ...
95  349  False  POINT (0 0)
[...]

PR #335, opened 2023-12-22
ENH: Add test of Arrow roundtrip of boolean values, raise exception on GDAL <3.8.3 for bool values via Arrow API

brendan-ward:
Test to reproduce #334 incorrect handling of boolean values using Arrow I/O (mostly to see if this also fails on GDAL >= 3.8.1).

As the root cause has been solved and released in GDAL, this PR now resolves https://github.com/geopandas/pyogrio/issues/334


theroggy:
[GDAL 3.8.3](https://github.com/OSGeo/gdal/blob/v3.8.3/NEWS.md#vector-drivers) fixes this and has just been released...

brendan-ward:
Because misreading boolean values is a **bad** error, this adds an exception to `read_arrow` if boolean values are detected in the pyarrow table read from GDAL.  I didn't add an exception to `open_arrow` because we don't know the contents of the stream it returns.  I'm assuming that given the severity of this bug, it is better to raise an exception and instruct users to avoid the arrow API for GDAL < 3.8.3 than to do something less intrusive like raise a warning that might be ignored by the user.

I used an `xfail` for the failing test case so that we can detect if this fix gets backported to other GDAL versions (e.g., if there is a 3.7.4 that includes it)

theroggy:
> Because misreading boolean values is a **bad** error, this adds an exception to `read_arrow` if boolean values are detected in the pyarrow table read from GDAL. I didn't add an exception to `open_arrow` because we don't know the contents of the stream it returns. I'm assuming that given the severity of this bug, it is better to raise an exception and instruct users to avoid the arrow API for GDAL < 3.8.3 than to do something less intrusive like raise a warning that might be ignored by the user.

Good idea!

> I used an `xfail` for the failing test case so that we can detect if this fix gets backported to other GDAL versions (e.g., if there is a 3.7.4 that includes it)

I'm not sure if the xfail is that interesting as the test testing for the exception for GDAL < 3.8 will fail anyway. It feels like it's a pity that it is not possible to specify a "match" in the xfail decorator. If that would be possible I think the "exception" test could be replaced by that using only the xfail, or am I missing something?

theroggy (approved):

jorisvandenbossche:
According to the GDAL PR fixing this, this only affects reading GPKG/FlatGeoBuf files. Would we have a way to check for that? (we don't know the driver because that gets inferred while reading, except if we explicitly get that info (``read_info`` returns it?). But maybe we could then only do that when we detect there is a boolean column)

brendan-ward:
Indeed; I had assumed it applied to all the bool drivers because it was in `OGRArrowArrayHelper::SetBoolOn()` (even though it did specifically mention only those drivers).

It looks like boolean values roundtripped properly for SQLite / GeoJSON / GML (not able to easily test postgres) - so our exception would be raised when the functionality is just fine.  Will add a fix for this.

jorisvandenbossche:
Maybe it is because that utility function is only used by drivers that have a custom implementation of GetArrowStream that sidesteps the internal GDAL data model (which are only GPKG and FlatGeoBuf, AFAIK, apart from Parquet/Arrow, but those already are already in arrow format after reading, of course). Other drivers fall back to a default implementation based on the standard APIs to read data (like we also use)

PR #336, opened 2024-01-11
CI: Add env name to conda test names

theroggy:
This way it is immediately clear when tests fail what environments (latest vs minimal vs libgdal3.5) they are failing for...


brendan-ward (approved):
Thanks @theroggy !

PR #342, opened 2024-01-23
MNT: Switch from flake8 to ruff for linting

### LINKED ISSUE #341: MNT: Change linting / precommit to use ruff
GeoPandas switched to ruff a while back; we should do the same.

brendan-ward:
Resolves #341 

Similar to changes in [GeoPandas #2921](https://github.com/geopandas/geopandas/pull/2921) but without needing to configure additional rules.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,5 +3,5 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.1.13"
     hooks:
-      - id: ruff
       - id: ruff-format
+      - id: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,11 @@
 files: 'pyogrio\/'
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.1.0
     hooks:
       - id: black
         language_version: python3
-  - repo: https://github.com/pycqa/flake8
-    rev: 5.0.4
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: "v0.1.13"
     hooks:
-      - id: flake8
-        language: python_venv
+      - id: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,7 @@
 files: 'pyogrio\/'
 repos:
-  - repo: https://github.com/psf/black
-    rev: 23.1.0
-    hooks:
-      - id: black
-        language_version: python3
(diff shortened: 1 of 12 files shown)


martinfleis (approved):
Thanks! We could potentially also consider switching from black to ruff-format. I am using it in some other projects and the transition was flawless.

brendan-ward:
Switched to ruff-format and removed unnecessary `# noqa` statements.

theroggy:
Maybe add `ruff` as dependency in environment-dev.yml?

.pre-commit-config.yaml
```diff
@@ -1,12 +1,7 @@
...
     hooks:
-      - id: black
-        language_version: python3
-  - repo: https://github.com/pycqa/flake8
-    rev: 5.0.4
-    hooks:
-      - id: flake8
-        language: python_venv
+      - id: ruff
+      - id: ruff-format
```
jorisvandenbossche, line 7: Should we put the formatter first? (eg I can imagine that the linter complains about a too long line, which would already be solved by the formatter afterwards)

jorisvandenbossche:
> Maybe add `ruff` as dependency in environment-dev.yml?

`pre-commit` is included, and then the idea is that you install the pre-commit, which will handle getting ruff (that's also how it went for black and flake8, which aren't included in the yml)

Of course if you don't use pre-commit and want to run ruff manually, you also need it in your env.

theroggy:
> > Maybe add `ruff` as dependency in environment-dev.yml?
> 
> `pre-commit` is included, and then the idea is that you install the pre-commit, which will handle getting ruff (that's also how it went for black and flake8, which aren't included in the yml)
> 
> Of course if you don't use pre-commit and want to run ruff manually, you also need it in your env.

I personally use visual studio code and then it is convenient that formatter and linter are installed so the formatting is automatically applied and linting errors are shown. Once you commit pre-commit will check this as well... but this is later in the process.

But, if this is rather typical for my workflow, no problem not to include it... not problem to install it myself...

.pre-commit-config.yaml
brendan-ward: Sure; was just following the [example](https://github.com/astral-sh/ruff-pre-commit).

I'm assuming we don't want to auto-fix lint errors, which is where putting the linter first is recommended.

brendan-ward:
Given that we had `black` in the `environment-dev.yml` (and I removed here), I have no issues with adding ruff as a development dependency.  For vscode, I've been using the [ruff extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff), which I believe ships with its own install of ruff that would be used to highlight things in the vscode UI.

jorisvandenbossche (approved):

PR #343, opened 2024-01-24
MNT: use gdal 3.8.3 in environment-dev.yml

theroggy:
There are quite some new features that are depending on GDAL 3.8, so seems logical to use GDAL 3.8 during development...


jorisvandenbossche (approved):

brendan-ward (approved):

brendan-ward:
Thanks @theroggy !

Issue #314, opened 2023-10-12
ENH: Support writing using the Arrow API


brendan-ward:
GDAL recently introduced support for writing files using the Arrow interface in [GDAL #8517](https://github.com/OSGeo/gdal/pull/8517).

kylebarron:
GDAL 3.8.0 is out! https://github.com/OSGeo/gdal/releases/tag/v3.8.0

kylebarron:
If no one else is interested in taking this, I may give a shot at this before 0.8. At the current release cadence I'd like to get this into 0.8 instead of 0.9.

PR #346, opened 2024-01-25
Write Arrow Table/RecordBatchReader to GDAL

kylebarron:
A _**very**_ WIP implementation for writing Arrow. This is mostly to start discussion and is largely inspired/copied from https://github.com/OSGeo/gdal/pull/9133.

Notes:

- This uses the [Arrow PyCapsule interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) so that it can read from _any_ object that exposes itself via Arrow. This means that it doesn't have a pyarrow dependency, and can read from `pyarrow.Table`, `pyarrow.RecordBatchReader`, and geoarrow-rs' [`GeoTable`](https://geoarrow.org/geoarrow-rs/python/v0.1.0/api/core/table/#geoarrow.rust.core.GeoTable.__arrow_c_stream__). The `RecordBatchReader` means that it can handle an _iterator_ of batches, so they don't all have to be materialized at once in a Python Table.

Todo:

- Figure out how to create the OGR Layer first
- I'm sure my use of pointers is wildly incorrect.

References:

- https://github.com/OSGeo/gdal/pull/9133
- [`OGR_L_WriteArrowBatch`](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_L_WriteArrowBatch9OGRLayerHPK11ArrowSchemaP10ArrowArrayPPc) 
- https://gdal.org/tutorials/vector_api_tut.html#writing-to-ogr-using-the-arrow-c-data-interface

Closes https://github.com/geopandas/pyogrio/issues/314.


pyogrio/_io.pyx
```diff
@@ -1678,6 +1678,117 @@ cdef infer_field_types(list dtypes):
...
+    return write_arrow_stream_capsule(stream_capsule)
+
+cdef OGRErr write_arrow_stream_capsule(OGRLayerH destLayer, object capsule):
+    cdef ArrowArrayStream stream
+    cdef ArrowSchema schema
+    cdef ArrowArray array
+
+    stream = PyCapsule_GetPointer(capsule, "arrow_array_stream")
+    if not stream:
+        raise RuntimeError("Accessing PyCapsule pointer named 'arrow_array_stream' failed.")
```
jorisvandenbossche, line 1698: It's probably fine to check for a NULL stream, but that the message can say something about "no valid stream" (the accessing of the pointer itself will have succeeded at this point, because if that fails, `PyCapsule_GetPointer` already raises an error which cython translates to a Python exception)

pyogrio/_ogr.pxd
```diff
@@ -191,11 +191,20 @@ cdef extern from "ogr_srs_api.h":
 
 
 cdef extern from "arrow_bridge.h":
+    struct ArrowArray:
+        void (*release)(struct ArrowArray*)
```
jorisvandenbossche, line 195:
```suggestion
        void (*release)(ArrowArray*)
```

If you want to see a working version, see eg https://github.com/apache/arrow/blob/c67d0260d4e96472b5cbdff66ca67ead2b9abe4c/python/pyarrow/includes/libarrow.pxd#L2795-L2803

Cython generally leaves out "struct" in signatures

I think this causes some of the compilation errors

jorisvandenbossche: Or nanoarrow-python has the full declaration, but it's not committed to github but autogenerated, so you can only see it locally. Copying what I have:

```
    struct ArrowSchema:
        const char* format
        const char* name
        const char* metadata
        int64_t flags
        int64_t n_children
        ArrowSchema** children
        ArrowSchema* dictionary
        void (*release)(ArrowSchema*)
        void* private_data

    struct ArrowArray:
        int64_t length
        int64_t null_count
        int64_t offset
        int64_t n_buffers
        int64_t n_children
        const void** buffers
        ArrowArray** children
        ArrowArray* dictionary
        void (*release)(ArrowArray*)
        void* private_data

    struct ArrowArrayStream:
        int (*get_schema)(ArrowArrayStream*, ArrowSchema* out)
        int (*get_next)(ArrowArrayStream*, ArrowArray* out)
        const char* (*get_last_error)(ArrowArrayStream*)
        void (*release)(ArrowArrayStream*)
        void* private_data
``

File: pyogrio/_io.pyx [hunk 3]
```diff
@@ -1678,6 +1678,117 @@ cdef infer_field_types(list dtypes):
...
+    IF CTE_GDAL_VERSION < (3, 8, 0):
+        raise RuntimeError("Need GDAL>=3.8 for Arrow write support")
+
+    stream_capsule = obj.__arrow_c_stream__()
+    return write_arrow_stream_capsule(stream_capsule)
+
+cdef OGRErr write_arrow_stream_capsule(OGRLayerH destLayer, object capsule):
+    cdef ArrowArrayStream stream
+    cdef ArrowSchema schema
+    cdef ArrowArray array
```
jorisvandenbossche, line 1694:
```suggestion
    cdef ArrowArrayStream* stream
    cdef ArrowSchema* schema
    cdef ArrowArray* array
```

(I would expect we are working with pointers here, but didn't look in detail at how those variables are used in the code)

jorisvandenbossche: Or at least for the stream, maybe not for the schema/array

pyogrio/_io.pyx
```diff
@@ -1678,6 +1678,117 @@ cdef infer_field_types(list dtypes):
...
+    char** options
+):
+    # The schema object is a struct type where each child is a column.
+    for child in schema.n_children:
+        # Access the metadata for this column
+        const char *metadata = child.metadata
+
+        # TODO: I don't know how to parse this metadata in C... I guess I can just use Python APIs for this in Cython?
+        # https://github.com/OSGeo/gdal/pull/9133/files#diff-37bedc92ae1d5e04706c7b9f8ea9e9fcccf984ca0c9997e2020ff85f1b958433R1159-R1185
+        # if metadata:
```
jorisvandenbossche, line 1754: Hmm, it's a pity that GDAL doesn't provide a helper to create a full layer definition from a schema, instead of only field by field (that would us to access the individual children of the ArrowSchema). 

We might want to vendor some helpers from nanoarrow-c to extract the children, and the metadata etc.

pyogrio/_io.pyx
```diff
@@ -1678,6 +1678,117 @@ cdef infer_field_types(list dtypes):
...
+    stream.release(stream)
+
+# Create output fields using CreateFieldFromArrowSchema()
+static bool create_fields_from_arrow_schema(
+    OGRLayerH destLayer,
+    const struct ArrowSchema* schema,
+    char** options
+):
+    # The schema object is a struct type where each child is a column.
+    for child in schema.n_children:
```
jorisvandenbossche, line 1748:
```suggestion
    for child in range(schema.n_children):
```

pyogrio/_io.pyx
jorisvandenbossche: Accessing the child is something we can probably do quite easily, as it's something like `schema.children[i]`

For a first prototype, I think we could also ignore the metadata for a moment, and manually specify which are the geometry columns.

jorisvandenbossche (commented):
Thanks for starting this! 
It have wanted to look at this as well, but as you might have noticed in other projects, my time for geo is a bit limited at the moment ;) But so a starting WIP PR is always a good reason to take a look! I hope my comments provide some useful pointers

pyogrio/_ogr.pxd
kylebarron: Updated in https://github.com/geopandas/pyogrio/pull/346/commits/331599852e1b0ac8a9bddd686fa834f24d604939

pyogrio/_io.pyx
kylebarron: I meant to change from

```diff
- for i in range(schema.n_children):
+ for child in schema.children:
```

I'm not sure if cython lets you iterate through a list of pointers as if it's a python list?

pyogrio/_io.pyx
jorisvandenbossche: Yes, but there we allocate those structs ourselves (to let them be filled in by GDAl and to export them), while here we get it from the capsule

pyogrio/_io.pyx
jorisvandenbossche: > I'm not sure if cython lets you iterate through a list of pointers as if it's a python list?

No idea either, you will have to try and see ;)

pyogrio/_io.pyx
kylebarron: 😅 I can't write anything without the Rust IDE telling me exactly what to do

kylebarron:
This now compiles on 3.8! I still need to figure out how to exclude this code from trying to compile on earlier GDAL versions, and I still need to test it.

jorisvandenbossche:
>  I still need to figure out how to exclude this code from trying to compile on earlier GDAL versions, and I still need to test it.

Hmm, looking at how it is done on the reading side, it's only the declaration in pxd that we put behind an IF, and then in the actual code we just raise an error at the beginning of the function, but still compile it regardless of the GDAL version (wondering if cython does something smart here)

kylebarron:
Yeah I was confused because I thought I was doing the same process as we have for reading

kylebarron:
I was able to compile and install it locally with

```
virtualenv env
source ./env/bin/activate
pip install --no-cache-dir -e '.[dev,test,geopandas]'
```

but trying to use it with

```py
import pyarrow as pa
from pyogrio._io import ogr_write_arrow

import geopandas as gpd
from geopandas.io.arrow import _geopandas_to_arrow
gdf = gpd.read_file(gpd.datasets.get_path('nybb'))
table = _geopandas_to_arrow(gdf)

new_schema = table.schema.set(4, pa.field("geometry", pa.binary(), True, {"ARROW:extension:name": "ogc.wkb"}))
new_table = pa.Table.from_batches(table.to_batches(), new_schema)

test = ogr_write_arrow('test.fgb', 'test', 'FlatGeobuf', new_table, 'MultiPolygon', {}, {})
```

is crashing Python. Does anyone have any suggestions for how to debug? I don't know where to go from here.

jorisvandenbossche:
I fixed the segfault (gdb was pointing at exporting the schema, so `stream.get_schema(stream, schema)`. My understanding is that we need to pass a schema as a pointer, but we need to have allocated it (and get_schema "fills it in"). So changed the declaration of schema and array a bit. 
That fixed the segfault. Then there was still an error in `create_fields_from_arrow_schema` always erroring because of checking for a wrong error code. But with that fixed, your example script runs without any error. The strange thing is that it only doesn't write any file .. Not fully sure what I am missing here.

jorisvandenbossche:
Those annoying error return codes ... ;) We need to very carefully inspect every one while writing: the GDAL function `OGR_L_WriteArrowBatch` is returning a bool (true on success), but the `ArrowArrayStream.get_next` uses "0 on success, a non-zero error code otherwise.". While the code was doing it the other way around ..

Now it _is_ writing actual content to the file. Just not for FlatGeobuf (that gives some error while writing), but for eg GPKG it is working.

I also opened https://github.com/geopandas/pyogrio/pull/351 to more easily integrate this with all options for setting up the layer/dataset.

kylebarron:
I merged in https://github.com/geopandas/pyogrio/pull/351 which simplified this a lot.

I'm still struggling with my dev environment though. In particular getting 
```
ModuleNotFoundError: No module named 'pyogrio._ogr'
```
when trying to `import pyogrio.raw`. This is after `pip install '.[dev,test,geopandas]'`.

kylebarron:
Ah yes I've hit this so many times. It's also why compiled rust projects advise to use a different name for the folder containing python code.

kylebarron:
I added some conditional compilation statements around other functions that were trying to use GDAL Arrow struct definitions

```
IF CTE_GDAL_VERSION >= (3, 8, 0):
```

so let's see if we can get it to compile with that

kylebarron:
I'll try to add some tests, but the meat of the PR should be ready for review

kylebarron:
It looks like the geometry is not being picked up correctly. E.g trying to write GeoJSON I see:
```
{ "type": "Feature", "properties": { "pop_est": 3856181, "continent": "Europe", "name": "Bosnia and Herz.", "iso_a3": "BIH", "gdp_md_est": 42530.0, "geometry": "0103000000010000001700000080C2F5285C8F32403033333333534540C0FFD3A7C7AC314090BAD7EFA783454030CE3DAB204C3140982FB2B021B9454080F5993A89EA3040E0D322EE77D54540F0F63671D9743040480BF3574705464000E525605A3D3040B82A7843F22C46404070F66A03802F403840268BCB684640809EEF3032EB2F40D04E9A65EC9D4640E00AE5B87251304090BB3A3987804640D01CF6C9F1883040B877F6F4159B4640802C6AA48C003140D04E9A65EC9D464090639DD79DDC314090D385B7AB884640300037719F8D3240A870BC87718A46403F244770670133404254F4291C6E464050C8242367013340E08442041C6E464050F2CD36375E33406039B4C8766E4640D0AE5FB01B1E3340A8F5622827364640007311DF8999334040F0BF95EC0446406062105839743340D86D3480B7C8454090853DEDF0373340182B6A300DC3454030EBE2361A083340F8449E245DB7454040478FDFDBB4324020CD58349D99454080C2F5285C8F32403033333333534540" }, "geometry": null },
```
I was hoping that the geometry column would automatically be picked up, but it looks like no. Presumably we need [this section](https://github.com/OSGeo/gdal/pull/9133/files#diff-37bedc92ae1d5e04706c7b9f8ea9e9fcccf984ca0c9997e2020ff85f1b958433R1187-R1223) of the upstream GDAL implementation

kylebarron:
With the latest changes, I require a `geometry_name` argument and it appears to now be correctly interpreting the geometry column. _But_ the written file is still getting corrupted. Here the GeoJSON ends abruptly

```json
{ "type": "Feature", "properties": { "pop_est": 1218208, "continent": "North America", "name": "Trinidad and Tobago", "iso_a3": "TTO", "gdp_md_est": 43570.0 }, "geometry": { "type": "Polygon", "coordinates": [ [ [ -61.68, 10.76 ], [ -61.105, 10.89 ], [ -60.895, 10.855 ], [ -60.935, 10.11 ], [ -61.77, 10.0 ], [ -61.95, 10.09 ], [ -61.66, 10.365 ], [ -61.68, 10.76 ] ] ] } },
{ "type": "Feature", "properties": { "pop_est": 13026129, "continent": "Africa", "name": "S. Sudan", "iso_a3": "SSD", "gdp_md_est": 20880.0 }, "geometry": { "type": "Polygon", "coordinates": [ [ [ 30.833852421715427, 3.509171604222463 ], [ 29.953500197069474, 4.173699042167684 ], [ 29.71599531425602, 4.600804755060153 ], [ 29.1590784034465, 4.389267279473231 ], [ 28.696677687298802, 4.455077215996937 ], [ 28.428993768026913, 4.287154649264494 ], [ 27.979977247842811, 4.408413397637375 ], [ 27.37422610851749, 5.233944403500061 ], [ 27.21340905122517, 5.550953477394557 ], [ 26.465909458123235, 5.94671743410187 ], [ 26.213418409945117, 6.546603298362072 ], [ 25.796647983511178, 6.979315904158071 ], [ 25.124130893664727, 7.500085150579437 ], [ 25.114932488716789, 7.825104071479174 ], [ 24.567369012152085, 8.229187933785468 ], [ 23.886979580860668, 8.619729712933065 ], [ 24.19406772118765, 8.728696472403897 ], [ 24.53741516360202, 8.91753756573172 ], [ 24.794925745412684, 9.810240916008695 ], [ 25.069603699343986, 10.273759963267992 ], [ 25.790633328413946, 10.411098940233728 ], [ 2%
```

I really don't know what's going on there.

Ideally we would also be able to infer the geometry column name from the presence of the geometry extension name, but I don't know how to port [these lines](https://github.com/OSGeo/gdal/pull/9133/files#diff-37bedc92ae1d5e04706c7b9f8ea9e9fcccf984ca0c9997e2020ff85f1b958433R1187-R1223).

jorisvandenbossche:
Looking at how this is used in GDAL itself (the test from the PR you linked above), I first didn't understand how this works there without never setting any geometry field. But I assume that it works because the arrow table that emitted by GDAL itself uses the extension type that it then recognizes on input? 

Anyway, looking a bit more at the documentation (https://gdal.org/tutorials/vector_api_tut.html#writing-to-ogr-using-the-arrow-c-data-interface and https://gdal.org/api/ogrlayer_cpp.html#_CPPv4N8OGRLayer15WriteArrowBatchEPK11ArrowSchemaP10ArrowArray12CSLConstList), it seems that we need to set the `GEOMETRY_NAME` option, if not creating the geometry field on the layer definition explicitly. The example of `copy_layer` in the WriteArrowBatch docs either explicitly creates the geometry fields with CreateGeomField if there are multiple geometry fields, or otherwise sets the `GEOMETRY_NAME` option. 
The docs indicate that it should fall back to recognize a default "wkb_geometry" named column, but that didn't seem to work for me (the created GeoJSON still had a "null" geometry, like in your output above)

Adding a small patch to set the option, now it works for any geometry column name. The GeoJSON output is for me also randomly cut off at some point creating a corrupted file. But that might be a bug on the GDAL side. Because writing to GPKG _does_ create a file that correctly roundtrips.

jorisvandenbossche:
> The docs indicate that it should fall back to recognize a default "wkb_geometry" named column, but that didn't seem to work for me (the created GeoJSON still had a "null" geometry, like in your output above)

Ah, that's probably because I didn't specify the column name to `geometry_name` argument, and as a result the code here still creates a field with that name. And I suppose that if an attribute field exists with that name, it will not use that column anymore to write the geometry field.

But so with this patch

```diff
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1706,9 +1706,10 @@ def ogr_write_arrow(
         dataset_metadata, layer_metadata,
         &ogr_dataset, &ogr_layer,
     )
+    cdef char **options = dict_to_options({"GEOMETRY_NAME": geometry_name})
 
     stream_capsule = arrow_obj.__arrow_c_stream__()
-    return write_arrow_stream_capsule(ogr_layer, stream_capsule, geometry_name)
+    return write_arrow_stream_capsule(ogr_layer, stream_capsule, geometry_name, options)
 
 
 IF CTE_GDAL_VERSION >= (3, 8, 0):
```

I could successfully roundtrip to GPKG:

```
In [2]: gdf = pyogrio.read_dataframe(geopandas.datasets.get_path("naturalearth_lowres"))

In [3]: table = geopandas.io.arrow._geopandas_to_arrow(gdf)

In [6]: pyogrio.raw.write_arrow("test3.gpkg", table, geometry_type="MultiPolygon", geometry_name="geometry")
.. warnings ..

In [7]: pyogrio.read_dataframe("test3.gpkg")
Out[7]: 
         pop_est      continent                      name iso_a3  gdp_md_est                                           geometry
0       889953.0        Oceania                      Fiji    FJI        5496  MULTIPOLYGON (((180.00000 -16.06713, 180.00000...
1     58005463.0         Africa                  Tanzania    TZA       63177  POLYGON ((33.90371 -0.95000, 34.07262 -1.05982...
2       603253.0         Africa                 W. Sahara    ESH         907  POLYGON ((-8.66559 27.65643, -8.66512 27.58948...
3     37589262.0  North America                    Canada    CAN     1736425  MULTIPOLYGON (((-122.84000 49.00000, -122.9742...
4    328239523.0  North America  United States of America    USA    21433226  MULTIPOLYGON (((-122.84000 49.00000, -120.0000...
..           ...            ...                       ...    ...         ...                                                ...
172    6944975.0         Europe                    Serbia    SRB       51475  POLYGON ((18.82982 45.90887, 18.82984 45.90888...
173     622137.0         Europe                Montenegro    MNE        5542  POLYGON ((20.07070 42.58863, 19.80161 42.50009...
174    1794248.0         Europe                    Kosovo    -99        7926  POLYGON ((20.59025 41.85541, 20.52295 42.21787...
175    1394973.0  North America       Trinidad and Tobago    TTO       24269  POLYGON ((-61.68000 10.76000, -61.10500 10.890...
176   11062113.0         Africa                  S. Sudan    SSD       11998  POLYGON ((30.83385 3.50917, 29.95350 4.17370, ...

[177 rows x 6 columns]
```

jorisvandenbossche:
> Ideally we would also be able to infer the geometry column name from the presence of the geometry extension name, but I don't know how to port [these lines](https://github.com/OSGeo/gdal/pull/9133/files#diff-37bedc92ae1d5e04706c7b9f8ea9e9fcccf984ca0c9997e2020ff85f1b958433R1187-R1223).

I was going to say: we can do that on the python side where this should be simple, but of course I was thinking about a pyarrow table where we can easily check the schema. To make it generic for any object implementing the Arrow PyCapsule protocol, it indeed needs to be done similarly like the above. Will try to take a stab at this.

But so summarizing, we can probably do something along those lines:
- Provide a `geometry_name` keyword for if you want to specify this explicitly as a user (eg if your column has a custom name / doesn't have geoarrow metadata set)
- When looping over the schema's fields (in `create_fields_from_arrow_schema` to call `OGR_L_CreateFieldFromArrowSchema`), we try to infer if a column is a geometry column (with similar logic as the snippet from GDAL, i.e. checking the metadata). And when a column name is equal to the one passed as `geometry_name` or inferred as geometry type, we skip creating a field for it.
- Pass the inferred/specified geometry column name to GDAL through the option (I suppose if we infer it, GDAL would also infer it, but just setting it anyway shouldn't hurt). 
  - Future enhancement can then be to support multiple geometry fields (but I think that not all formats support this, so it's still good to use the simpler option for the case of 1 geometry field)

jorisvandenbossche:
Not the nicest piece of code, but it seems to be working ;)

(writing a table that was read with `pyogrio.raw.read_arrow`, and thus has a "ogc.wkb" column, works without specifying the geometry column name)

jorisvandenbossche:
> _But_ the written file is still getting corrupted. Here the GeoJSON ends abruptly

I think the last commit should fix this, by ensuring we properly close the GDAL dataset object.

pyogrio/_io.pyx
```diff
@@ -1767,6 +1783,30 @@ IF CTE_GDAL_VERSION >= (3, 8, 0):
...
+            key = metadata[pos:pos+length_key]
+            pos += length_key
+            length_value = int.from_bytes(metadata[pos:pos+4], byteorder=sys.byteorder)
+            pos += 4
+            value = metadata[pos:pos+length_value]
+            pos += length_value
+
+            if key == b"ARROW:extension:name":
+                if value == b"geoarrow.wkb" or value == b"ogc.wkb":
+                    return True
```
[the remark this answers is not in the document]
kylebarron: It would probably also be good to check here for any geoarrow extension arrays that are _not_ `geoarrow.wkb` and error. I'm sure someone will try to pass e.g. a `geoarrow.point` array in at some point and be confused why no geometry was written.

kylebarron:
> I think the last commit should fix this, by ensuring we properly close the GDAL dataset object.

Do you have an idea why CI is failing?

jorisvandenbossche:
The latest run seems to be mostly passing, except for linting (I think we switched to ruff since the PR was started, so might have to merge in main to get those changes and run pre-commit), and except for on Mac build (but that's a flaky segfault that has been happening lately, independent from this PR, so something we should try to investigate and fix, but nothing to worry about for this PR)

pyogrio/_io.pyx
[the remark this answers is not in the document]
jorisvandenbossche: We currently set the CRS at the moment of creating the layer, so that is before passing any Arrow data. So I would _expect_ it cannot deal with the CRS embedded in the schema metadata at that point

pyogrio/raw.py
```diff
@@ -559,3 +559,116 @@ def write(
...
+            if k in dataset_option_names:
+                dataset_kwargs[k] = v
+            elif k in layer_option_names:
+                layer_kwargs[k] = v
+            else:
+                raise ValueError(f"unrecognized option '{k}' for driver '{driver}'")
+
+    # TODO: pass these as dataset kwargs?
+    # promote_to_multi=promote_to_multi,
+    # nan_as_null=nan_as_null,
```
jorisvandenbossche, line 659: I suppose those can't be supported with Arrow writing. For `promote_to_multi`, we actually call a GDAL function on each individual geometry while looping for writing row by row. So with the batch writing of `write_arrow`, this would need to be done ahead of writing (and so for now I would leave that up to the user), or it would need to be done by GDAL under the hood (but currently there is no option for that).

For `nan_as_null`, if you already start from Arrow data, there is less need for this option, because you can already have proper nulls then. It was specifically added for numpy-based systems that use NaN as the null indicator.

jorisvandenbossche:
Pushed a few more changes:

- In the `write_arrow()` signature, changed they first keywords order to `data, path` instead of `path, data`. That makes it consistent with the `write_dataframe` (not consistent with the "raw" write function, but in functionality it is closer to the geopandas-based version)
- Added a docstring. Mostly copied from `write_dataframe`, just updating for the data keyword, and for `geometry_type` and `geometry_name`
- Added a few more tests

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+            "append to FlatGeobuf is not supported for GDAL <= 3.5.0 due to segfault"
+        )
+
+    if "promote_to_multi" in kwargs:
+        raise ValueError(
+            "The 'promote_to_multi' option is not supported when writing using Arrow"
+        )
+
+    if geometry_type is None:
+        raise ValueError("Need to specify 'geometry_type'")
```
jorisvandenbossche, line 673: This is a bit of a dilemma. 
In theory, we can pass through the None, which gets translated to `wkbNone` on the GDAL side, which is documented as "non-standard, for pure attribute records". And that works fine for writing data without any geometry column.

However, if you forget to specify this and you _have_ a geometry the column (which is by far the typical use case), you get a generic *"RuntimeError: Error while writing batch to OGR layer."* error (i.e. the error we raise if `OGR_L_WriteArrowBatch` returns a non-success code). And that is not very helpful or clear that this is due not specifying the geometry type.

So for now I opted to require passing this (and so essentially not enabling to write no geometry column. You can still specify a geometry type, but then eg with GPKG you get an all-null geometry column)

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+        layer name
+    driver : str, optional (default: None)
+        The OGR format driver used to write the vector file. By default write_dataframe
+        attempts to infer driver from path.
+    geometry_name : str, optional (default: None)
+        The name of the column in the input data that will be written as the
+        geometry field. Will be inferred from the input data if the geometry
+        column is annotated as an "geoarrow.wkb" or "ogc.wkb" extension type.
+        Otherwise needs to be specified explicitly.
+    geometry_type : str, optional (default: None)
```
kylebarron, line 603:
```suggestion
    geometry_type : str
```

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+    layer_options=None,
+    **kwargs,
+):
+    """
+    Write an Arrow compatible data source to an OGR file format.
+
+    Parameters
+    ----------
+    arrow_obj
+        The Arrow data to write. This can be any Arrow-compatible taular data
```
kylebarron, line 587:
```suggestion
        The Arrow data to write. This can be any Arrow-compatible tabular data
```

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+        object that implements the Arrow PyCapsule Protocol (i.e. has an
```
kylebarron, line 588: Maybe link to the PyCapsule protocol document?

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+        ``__arrow_c_stream__`` method), for example a pyarrow Table or
+        RecordBatchReader.
+    path : str
+        Path to file.
+    layer : str, optional (default: None)
+        layer name
+    driver : str, optional (default: None)
+        The OGR format driver used to write the vector file. By default write_dataframe
```
kylebarron, line 596:
```suggestion
        The OGR format driver used to write the vector file. By default write_arrow
```

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+    append=False,
+    dataset_metadata=None,
+    layer_metadata=None,
+    metadata=None,
+    dataset_options=None,
+    layer_options=None,
+    **kwargs,
+):
+    """
+    Write an Arrow compatible data source to an OGR file format.
```
kylebarron, line 582:
```suggestion
    Write an Arrow-compatible data source to an OGR file format.
```

pyogrio/raw.py
```diff
@@ -559,3 +559,173 @@ def write(
...
+        raise DataSourceError(
+            f"{driver} does not support write functionality in GDAL "
+            f"{get_gdal_version_string()}"
+        )
+
+    # prevent segfault from: https://github.com/OSGeo/gdal/issues/5739
+    if append and driver == "FlatGeobuf" and get_gdal_version() <= (3, 5, 0):
+        raise RuntimeError(
+            "append to FlatGeobuf is not supported for GDAL <= 3.5.0 due to segfault"
+        )
```
kylebarron, line 665: We can probably remove this if the arrow write api is only gdal 3.8+?

pyogrio/raw.py
himikof: Maybe the error message in the described case would be more clear and useful if the inner error details were preserved? Something like "Cannot find OGR geometry field for Arrow array {geometry_name}".

Also, is `wkbNone` actually the most useful default here in the low-level interface? It means exactly "there are zero geometry fields", which should be uncommon. Maybe `wkbUnknown` (meaning "there is a single geometry field with no further restrictions") would be more useful?

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+            if is_geometry_field(child):
+                continue
+
+            errcode = OGR_L_CreateFieldFromArrowSchema(
+                destLayer, child, options)
+            if errcode == 0:
+                raise RuntimeError(
+                    f"error while creating field from Arrow for field {i} with name "
+                    f"{get_string(child.name)} and type {get_string(child.format)}"
+                )
```
himikof, line 2361: This code should also use `exc_check()` to surface the inner error details.

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+        OGRLayerH destLayer,
+        const ArrowSchema* schema,
+        char** options,
+        str geometry_name
+    ):
+        # The schema object is a struct type where each child is a column.
+        cdef ArrowSchema* child
+        for i in range(schema.n_children):
+            child = schema.children[i]
+            metadata = child.metadata
```
brendan-ward, line 2345:
```suggestion
```

(unused)

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+
+            # Don't create property for geometry column
+            field_name = get_string(child.name)
```
brendan-ward, line 2348: Could be simplified a bit:

```Python
if get_string(child.name) == geometry_name or is_geometry_field(child):
```

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+            if field_name == geometry_name:
+                continue
+
+            if is_geometry_field(child):
+                continue
+
+            errcode = OGR_L_CreateFieldFromArrowSchema(
```
brendan-ward, line 2355:
```suggestion
            if !OGR_L_CreateFieldFromArrowSchema(...):
```

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+        dataset_metadata, layer_metadata,
+        &ogr_dataset, &ogr_layer,
+    )
+    if geometry_name:
+        opts = {"GEOMETRY_NAME": geometry_name}
+    else:
+        opts = {}
+    cdef char **options = dict_to_options(opts)
+
+    stream_capsule = arrow_obj.__arrow_c_stream__()
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, yes, we should of course check that the method is present to provide a better error message if not. Could also already do that at the Python level in `write_arrow`?

For validating that the method actually returns a capsule, I am not sure that is needed (it would be very strange for someone to use exactly that method name but not intent to implement the protocol). And in any case Python already will raise "ValueError: PyCapsule_GetPointer called with invalid PyCapsule object"

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+        GDALClose(ogr_dataset)
+
+        # GDAL will set an error if there was an error writing the data source
+        # on close
+        exc = exc_check()
+        if exc:
+            raise DataSourceError(f"Failed to write features to dataset {path}; {exc}")
+
+
+IF CTE_GDAL_VERSION >= (3, 8, 0):
```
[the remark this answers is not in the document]
jorisvandenbossche: I think that's what @kylebarron originally tried (mimicking how we did it for reading), but didn't get that working and so ended up with the current approach of putting the whole function in the IF. But can try once more with the other approach

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+            schema.release(&schema)
+            stream.release(stream)
+            raise e
+
+        while True:
+            errcode = stream.get_next(stream, &array)
+            if errcode != 0:
+                schema.release(&schema)
+                stream.release(stream)
+                raise RuntimeError("Error while accessing batch from stream.")
```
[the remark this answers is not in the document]
jorisvandenbossche: I would assume `array` doesn't need to be released here, as the error should mean that `array` didn't get filled in, so there might not be a valid release callback to call

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+
+            # We've reached the end of the stream
+            if array.release == NULL:
+                break
+
+            errcode = OGR_L_WriteArrowBatch(destLayer, &schema, &array, options)
```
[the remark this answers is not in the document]
jorisvandenbossche: (needs to use `if not OGR_..`, `!` is C/C++, but yes, that is cleaner!)

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+            if not errcode:
+                if array.release != NULL:
+                    array.release(&array)
+
+                schema.release(&schema)
+                stream.release(stream)
+                raise RuntimeError("Error while writing batch to OGR layer.")
```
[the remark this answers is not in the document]
jorisvandenbossche: @himikof good point. It would be good to come up with a test case that would trigger an error, to test this (some unsupported data type for a certain driver?), because it's not super clear to me if GDAL actually sets an error in that case (since this function returns a bool and not OGRErr (like other OGR functions where we check the set exception)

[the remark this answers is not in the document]
jorisvandenbossche: Found an example with FGB (first tried shapefile, but it seems GDAL just converts anything Shapefile doesn't support into its string representation and write it anyway ..):

```
In [1]: import shapely

In [2]: import pyogrio

In [3]: table = pa.table({"geometry": shapely.to_wkb(shapely.from_wkt(["POINT (0 0)", "POINT (1 1)"])), "col": [[0,1], [2,3,4]]})

In [4]: pyogrio.raw.write_arrow(table, "test_list.fgb", geometry_type="Point", geometry_name="geometry")
...
CPLE_AppDefinedError: ICreateFeature: Missing implementation for OGRFieldType 13
```

So with checking `exc_check()`, we indeed get a more targetted error message (it's not a super clear error message TBH, but that's something to improve on the GDAL side ;))

pyogrio/_io.pyx
```diff
@@ -2214,3 +2216,190 @@ def ogr_write(
...
+            if not OGR_L_WriteArrowBatch(destLayer, &schema, &array, options):
+                if array.release != NULL:
+                    array.release(&array)
+
+                schema.release(&schema)
+                stream.release(stream)
+
+                exc = exc_check()
+                if exc:
+                    raise exc
```
jorisvandenbossche, line 2322: I think we actually typically catch the `CPLE_..` error and reraise that with one of our generic error classes?

For example I see this pattern:

```
            except CPLE_BaseError as exc:
                raise FeatureError(str(exc))
```

But then what error class to use? (`FeatureError` or `FieldError`?)

jorisvandenbossche:
@himikof @brendan-ward thanks a lot for the extensive review! Already addressed a few of the comments, will need to go through the bulk of them still

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+    if not encoding:
+        encoding = locale.getpreferredencoding()
+
+    layer_created = create_ogr_dataset_layer(
+        path, layer, driver, crs, geometry_type, encoding,
+        dataset_kwargs, layer_kwargs, append,
+        dataset_metadata, layer_metadata,
+        &ogr_dataset, &ogr_layer,
+    )
+    if geometry_name:
```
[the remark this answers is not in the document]
jorisvandenbossche: `dict_to_options` already encodes it as UTF-8, always. But does that mean `dict_to_options` should be updated to not always do that, and to actually pass it an encoding?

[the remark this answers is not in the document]
brendan-ward: Sorry for the confusion, what I meant was this block for the dataset-level encoding, after the `layer_created = create_ogr_dataset_layer(...)` block:

```Python
    if driver == 'ESRI Shapefile':
        # force encoding for remaining operations to be in UTF-8 (even if user
        # provides an encoding) because GDAL will automatically convert those to
        # the target encoding because ENCODING is set as a layer creation option
        encoding = "UTF-8"

    else:
        # Now the dataset and layer have been created, we can properly determine the
        # encoding. It is derived from the user, from the dataset capabilities / type,
        # or from the system locale
        encoding = encoding or detect_encoding(ogr_dataset, ogr_layer)
```

`dict_to_options` should always encode keys and values to `UTF-8` before sending to GDAL, since those are used by GDAL rather than written out based on the target encoding of the dataset.

pyogrio/_io.pyx
brendan-ward: Probably `DataSourceError` since it is for the batch of features rather than individual features.

pyogrio/_io.pyx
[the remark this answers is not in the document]
brendan-ward: I don't have a terribly strong opinion about where validation should occur, but our general tendency for the I/O functions is to do as much validation as possible (e.g., for parameters not being inspected / modified in the Python tier) in the lower-level Cython code rather than in the Python tier.

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+    cdef get_extension_metadata(const ArrowSchema* schema):
+        cdef const char *metadata = schema.metadata
+
+        extension_name = None
+        extension_metadata = None
+
+        if metadata == NULL:
+            return extension_name, extension_metadata
+
+        n = int.from_bytes(metadata[:4], byteorder=sys.byteorder)
```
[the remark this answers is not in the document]
jorisvandenbossche: AFAIU the problem is that `metadata` is not a null terminated string, so you can't ask python the length of it. You can only determine the length by parsing it, and thus essentially relying on it being properly formed

(in any case, doing `len(metadata)` always gives 1, and doesn't actually calculate the length)

pyogrio/_io.pyx
```diff
@@ -2178,3 +2180,182 @@ def ogr_write(
...
+            pos += length_key
+            length_value = int.from_bytes(metadata[pos:pos+4], byteorder=sys.byteorder)
+            pos += 4
+            value = metadata[pos:pos+length_value]
+            pos += length_value
+
+            if key == b"ARROW:extension:name":
+                extension_name = value
+            elif key == b"ARROW:extension:metadata":
+                extension_metadata = value
```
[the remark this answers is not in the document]
jorisvandenbossche: The order isn't guaranteed, but indeed once we have both keys extracted, we can stop parsing the remainder

pyogrio/_io.pyx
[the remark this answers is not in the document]
brendan-ward: This seems dangerous though - doesn't this risk segfault / buffer overflow type issues (if the metadata is malformed and there aren't 4 bytes available)?  But this also might be a necessary part of reading a binary encoded format.

Sorry, I missed that metadata is a `char *`, which would be length 1, rather than `char **`, length TBD.

I'm not familiar with the memory layout of the schema; are we able to trust that a valid schema writer will always write either NULL or at least 4 bytes per the spec?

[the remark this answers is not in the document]
jorisvandenbossche: I think in general we have to assume properly formed data. I did a test with patching an ArrowSchema with invalid metadata (`b"\x01\x01\x01\x01"`, 4 bytes that represent a large integer value), and the code I wrote here segfaults on that, but pyarrow does as well, and nanoarrow does not segfault but reads into random memory (not sure which of the two is better .. ;))

pyogrio/_io.pyx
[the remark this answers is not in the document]
jorisvandenbossche: `encoding` isn't used anymore after the dataset/layer has been created. I am not entirely sure how GDAL handles that for writing the data through the Arrow API for a dataset/layer that isn't in UTF-8, though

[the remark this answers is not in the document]
brendan-ward: Ah, good point!  We ran into the same issue reading Arrow data as well, and pyarrow breaks badly (as we have it now) if the data read from the data source is not UTF-8 (e.g., if we forced it to be written that way for a non-shapefile driver)

pyogrio/_io.pyx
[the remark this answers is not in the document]
jorisvandenbossche: > I'm not familiar with the memory layout of the schema; are we able to trust that a valid schema writer will always write either NULL or at least 4 bytes per the spec?

Yes, that's what the spec requires: https://arrow.apache.org/docs/dev/format/CDataInterface.html#c.ArrowSchema.metadata (*"If omitted, this field MUST be NULL (not an empty string)."*)

pyogrio/_io.pyx
jorisvandenbossche: Used DataLayerError following our documentation, as it involves an error with a specific layer ("Errors relating to working with a single OGRLayer") and not with the whole dataset ("Errors relating to opening or closing an OGRDataSource (with >= 1 layers)")

pyogrio/_io.pyx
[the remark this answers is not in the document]
jorisvandenbossche: > Do you anticipate `write_arrow_stream_capsule` and `create_fields_from_arrow_schema` being reusable? It would make `ogr_write_arrow` a bit longer, but in theory those could just be in the body of that function.

I inlined `write_arrow_stream_capsule`, but kept `create_fields_from_arrow_schema` separate (I don't think we will directly reuse it elsewhere, but that feels a bit easier to follow code-organization wise, because it is a nice separate piece of functionality)

> It looks like at least `is_geometry_field` and `get_extension_metadata` could just be regular top-level functions and not need to be in this block.

Yes, moved those out of the IF block (and also dedented `create_fields_from_arrow_schema` by adding an IF raising an error inside it)

pyogrio/raw.py
brendan-ward: Would a 3-state variable make sense here, to enable writing without geometry?
- `""` (or some other arbitrary value to indicate unset): default, indicates not set by user and will raise exception
- `None`: provided by user to indicate there is no geometry to write
- `Point` ... `GeometryCollection`: valid geometry type

pyogrio/tests/test_arrow.py
```diff
@@ -304,3 +310,270 @@ def test_arrow_bool_exception(tmpdir, ext):
     else:
         with open_arrow(filename):
             pass
+
+
+# Point(0, 0)
```
brendan-ward, line 315: This also gets used in a variety of places, perhaps this could be added as a function in `conftest.py` and then just call it when used below?

```Python
def get_wkb_points(count=1):
    return np.array([bytes.fromhex("010100000000000000000000000000000000000000")]*count, dtype=object)
```

We could hook it up to other usages in a different PR, this would just set us up for that while simplifying here.

pyogrio/tests/test_arrow.py
```diff
@@ -304,3 +310,270 @@ def test_arrow_bool_exception(tmpdir, ext):
...
+points = np.array(
+    [bytes.fromhex("010100000000000000000000000000000000000000")] * 3,
+    dtype=object,
+)
+
+
+@requires_arrow_write_api
+def test_write(tmpdir, naturalearth_lowres):
```
brendan-ward, line 323:
```suggestion
def test_write_shp(tmpdir, naturalearth_lowres):
```

brendan-ward (commented):
Thanks for the updates!

Can you please add a test when `crs` is not passed in to verify the warning is raised?

If possible, would be good to have that also test (or in a separate test case) if GDAL auto detects CRS from metadata on the geometry column (per the `TODO` comment).

Can you please add a test for a driver that doesn't support write?  See `test_raw_io.py::test_write_unsupported`.

It would probably also be good to probe at close errors on write, similar to `test_raw_io.py::test_write_gdalclose_error`.

It might be good to test other drivers similar to `test_raw_io.py::test_write_supported`, though it looks like only FGB would be particularly unique there.

If possible, please add tests for append capability supported / unsupported, similar to  `test_raw_io.py::test_write_append` / `test_raw_io.py::test_write_append_unsupported`

Is the idea to add support to `write_dataframe` in a later PR?  (would be a good idea, this is already getting pretty big)

pyogrio/_io.pyx
```diff
@@ -2214,3 +2216,186 @@ def ogr_write(
...
+):
+    """Create output fields using CreateFieldFromArrowSchema()"""
+
+    IF CTE_GDAL_VERSION < (3, 8, 0):
+        raise RuntimeError("Need GDAL>=3.8 for Arrow write support")
+
+    # The schema object is a struct type where each child is a column.
+    cdef ArrowSchema* child
+    for i in range(schema.n_children):
+        child = schema.children[i]
```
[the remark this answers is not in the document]
jorisvandenbossche: That would again be an invalid struct, but for this case it's easy to add some validation, so added that (I checked Arrow C++ and they only have a debug check when importing an ArrowSchema, so meaning they would segfault if receiving such an invalid struct in released packages)

jorisvandenbossche:
> Can you please add a test when `crs` is not passed in to verify the warning is raised?
> 
> If possible, would be good to have that also test (or in a separate test case) if GDAL auto detects CRS from metadata on the geometry column (per the `TODO` comment).

Added a test for not providing a `crs`. That directly tests the second aspect as well, because the table that we are writing in test was written with the result of `read_arrow` which will have this CRS metadata in its schema.

> Can you please add a test for a driver that doesn't support write? See `test_raw_io.py::test_write_unsupported`.
> 
> It would probably also be good to probe at close errors on write, similar to `test_raw_io.py::test_write_gdalclose_error`.
> 
> It might be good to test other drivers similar to `test_raw_io.py::test_write_supported`, though it looks like only FGB would be particularly unique there.
> 
> If possible, please add tests for append capability supported / unsupported, similar to `test_raw_io.py::test_write_append` / `test_raw_io.py::test_write_append_unsupported`

Copied and adapted those tests from raw io (and fixed the `append=True` case, we were still incorrectly trying to add new fields to the existing layer). Only for geojson-like formats I get some error that I still have to report upstream.

> Is the idea to add support to `write_dataframe` in a later PR? (would be a good idea, this is already getting pretty big)

Yes, that was my idea. I already have a branch locally that starts testing that (most errors are about the missing support for promoting the geometry type, but which is something for which we could add some custom shapely code in the geopandas -> arrow conversion), but indeed was planning to leave that for a separate PR ;)

pyogrio/raw.py
```diff
@@ -559,3 +579,145 @@ def write(
...
+        specified explicitly when creating a new layer.
+        Possible values are: "Unknown", "Point", "LineString", "Polygon",
+        "MultiPoint", "MultiLineString", "MultiPolygon" or "GeometryCollection".
+
+        This parameter does not modify the geometry, but it will try to force the layer
+        type of the output file to this value. Use this parameter with caution because
+        using a non-default layer geometry type may result in errors when writing the
+        file, may be ignored by the driver, or may result in invalid files.
+    encoding : str, optional (default: None)
+        Only used for the .dbf file of ESRI Shapefiles. If not specified,
```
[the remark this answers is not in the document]
jorisvandenbossche: I think in the case of writing with Arrow, it's actually correct to say that `encoding` is only relevant for Shapefiles? Because we only use it to set up the GDAL option for Shapefiles, but then afterwards don't use it to actually pass the string data bytes to GDAL, given we pass the data as Arrow data.

pyogrio/_io.pyx
```diff
@@ -2291,3 +2293,199 @@ def ogr_write(
...
+    dataset_metadata=None,
+    layer_metadata=None,
+):
+    IF CTE_GDAL_VERSION < (3, 8, 0):
+        raise RuntimeError("Need GDAL>=3.8 for Arrow write support")
+
+    cdef OGRDataSourceH ogr_dataset = NULL
+    cdef OGRLayerH ogr_layer = NULL
+    cdef char **options = NULL
+    cdef ArrowArrayStream* stream
```
brendan-ward, line 2319:
```suggestion
    cdef ArrowArrayStream* stream = NULL
```

pyogrio/tests/test_arrow.py
```diff
@@ -339,3 +354,446 @@ def test_arrow_bool_exception(tmpdir, ext):
...
+        geometry_name=meta["geometry_name"] or "wkb_geometry",
+    )
+    assert filename.exists()
+
+
+@requires_arrow_write_api
+def test_write_unsupported(tmpdir, naturalearth_lowres):
+    meta, table = read_arrow(naturalearth_lowres)
+
+    filename = os.path.join(str(tmpdir), "test.gdb")
```
brendan-ward, line 499:
```suggestion
    filename = os.path.join(str(tmpdir), "test.json")
```

(`.gdb` is FileGDB)

pyogrio/raw.py
```diff
@@ -593,3 +613,152 @@ def write(
...
+        )
+
+    if geometry_type is None:
+        raise ValueError("'geometry_type' keyword is required")
+
+    dataset_metadata, layer_metadata = _validate_metadata(
+        dataset_metadata, layer_metadata, metadata
+    )
+
+    # TODO: does GDAL infer CRS automatically from geometry metadata?
```
brendan-ward, line 737: From the associated test case, it looks like GDAL does not infer this automatically; comment can be removed?

brendan-ward (commented):
Thanks for the updates @jorisvandenbossche , this is getting really close.  A few minor outstanding comments, but otherwise this looks ready to merge (and we can deal with outstanding issues around encoding - namely not allowing encoding for anything other than shapefiles -  in #384 after this is merged)

brendan-ward (approved):
Thanks @jorisvandenbossche @kylebarron !  🚀

kylebarron:
Honestly it was mostly @jorisvandenbossche , but I'm happy to have got the ball rolling!

Issue #357, opened 2024-01-31
DOC: Documentation lists incorrect key for results of read_info() method


ebkurtz:
documentation states the dictionary returned by `pyogrio.read_info()` has entry: "geometry": "<geometry type>".  The key is actually called "geometry_type"

jorisvandenbossche:
Closed by  #358

PR #358, opened 2024-01-31
DOC: Correct key name in pyogrio.read_info() docstring 

ebkurtz:
Corrected key "geometry" to "geometry_type" in docstring for pyogrio.io.read_info() method. #357


jorisvandenbossche (approved):
Thanks for the catch and fix!

PR #139, opened 2022-07-08
FIX: use user-provided encoding instead of default encoding of data source

brendan-ward:
fixes #138

Issue #138, opened 2022-07-08
BUG:  parameter encoding not work when engine='pyogrio'


CNFeffery:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

![image](https://user-images.githubusercontent.com/49147660/177937074-c9510e40-0f02-42aa-b290-8665613eae36.png)

#### Problem description

parameter `encoding` not work when set `engine='pyogrio'` in `gpd.read_file()`

#### Output of ``geopandas.show_versions()``

<details>

```
SYSTEM INFO
-----------
python     : 3.8.13 (default, Mar 28 2022, 06:59:08) [MSC v.1916 64 bit (AMD64)]
executable : C:\Conda\envs\geopandas-env\python.exe
machine    : Windows-10-10.0.22000-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.5.0
GDAL data dir: None
PROJ       : 9.0.1
PROJ data dir: C:\Conda\envs\geopandas-env\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.0
pandas     : 1.4.3
fiona      : 1.8.21
numpy      : 1.23.0
shapely    : 1.8.2
rtree      : 1.0.0
pyproj     : 3.3.1
matplotlib : 3.5.2
mapclassify: 2.4.3
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : 0.12.0
```

</details>

CNFeffery:
@jorisvandenbossche  https://github.com/CNFeffery/geopandas-encoding-issues-demo/blob/main/demo.zip
![image](https://user-images.githubusercontent.com/49147660/178016037-e3f0684a-43ca-4a96-a25c-5cb39da2d374.png)

brendan-ward:
`gdalinfo` seems to automatically detect the correct encoding of the test file.  We're first trying that ourselves via `OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8)`, but perhaps this is returning `False` in this case, in which case 
[...]

PR #361, opened 2024-02-23
BUG: fix encoding issues on windows for some formats

theroggy:
I noticed in some [geofileops tests](https://github.com/geofileops/geofileops/actions/runs/8110958569/job/22169297289?pr=505) that using pyogrio to write/read dataframes to ".csv" files gives encoding issues.

This PR fixes those.

Odd detail: I only saw this behaviour on the github CI windows systems: locally I couldn't reproduce. Apparently:
- when running tests on my local windows `locale.getprefferedencoding()` returns "UTF-8", even though when I run `locale.getprefferedencoding()` in a seperate script it returns the typical one for windows: "cp1252".
- in the tests on the github windows machines, `locale.getprefferedencoding()` returns "cp1252".

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -476,7 +476,7 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
     str or None
     """
 
-    if ogr_layer is not NULL and OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8):
+    if OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8):
         return 'UTF-8'
 
     driver = get_driver(ogr_dataset)
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1953,12 +1953,7 @@ def ogr_write(
     # Now the dataset and layer have been created, we can properly determine the
     # encoding. It is derived from the user, from the dataset capabilities / type,
     # or from the system locale
-    encoding = (
-        encoding
-        or detect_encoding(ogr_dataset, ogr_layer)
-        or locale.getpreferredencoding()
-    )
-
+    
     ### Create the fields
     field_types = None
     if num_fields > 0:
@@ -2006,6 +2001,12 @@ def ogr_write(
 
     ### Create the features
     ogr_featuredef = OGR_L_GetLayerDefn(ogr_layer)
+
+    encoding = (
+        encoding
+        or detect_encoding(ogr_dataset, ogr_layer)
+        or locale.getpreferredencoding()
+    )
 
     supports_transactions = OGR_L_TestCapability(ogr_layer, OLCTransactions)
     if supports_transactions:
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -462,24 +462,40 @@ cdef get_metadata(GDALMajorObjectH obj):
 
 
 cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
-    """Attempt to detect the encoding of the layer.
+    """Attempt to detect the encoding of the dataset or layer.
     If it supports UTF-8, use that.
     If it is a shapefile, it must otherwise be ISO-8859-1.
 
     Parameters
     ----------
     ogr_dataset : pointer to open OGR dataset
-    ogr_layer : pointer to open OGR layer
+    ogr_layer : NULL or pointer to open OGR layer
 
     Returns
     -------
     str or None
     """
 
-    if OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8):
+    if ogr_layer is not NULL and OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8):
         return 'UTF-8'
 
     driver = get_driver(ogr_dataset)
+    return get_encoding_for_driver(driver)
+
+
+cdef get_encoding_for_driver(str driver):
+    """Get the typical encoding for the driver, or None.
+
+    Parameters
+    ----------
+    driver : str
+        driver to get the encoding for
+    
+    Returns
+    -------
+    str or None
+        the typical encoding if this is applicable for the driver, or None
+    """
     if driver == 'ESRI Shapefile':
         return 'ISO-8859-1'
 
@@ -1906,11 +1922,10 @@ def ogr_write(
         gdal_tz_offsets = {}
 
     ### Setup up dataset and layer
-    # Encoding is derived from the user, from the dataset capabilities / type,
-    # or from the system locale
+    # Encoding is derived from the user, from the driver or from the system locale
     encoding = (
         encoding
-        or detect_encoding(ogr_dataset, ogr_layer)
+        or get_encoding_for_driver(driver)
         or locale.getpreferredencoding()
     )
 
@@ -2037,7 +2052,7 @@ def ogr_write(
                             field_value = str(field_value)
 
                         try:
-                            value_b = field_value.encode(encoding)
+                            value_b = field_value.encode("UTF-8")
                             OGR_F_SetFieldString(ogr_feature, field_idx, value_b)
 
                         except AttributeError:
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -504,6 +504,9 @@ cdef get_encoding_for_driver(str driver):
         # per https://help.openstreetmap.org/questions/2172/what-encoding-does-openstreetmap-use
         return "UTF-8"
 
+    if driver == "XLSX":
+        return "UTF-8"
+
     return None
 
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2002,15 +2002,15 @@ def ogr_write(
     ### Create the features
     ogr_featuredef = OGR_L_GetLayerDefn(ogr_layer)
 
+    supports_transactions = OGR_L_TestCapability(ogr_layer, OLCTransactions)
+    if supports_transactions:
+        start_transaction(ogr_dataset, 0)
+
     encoding = (
         encoding
         or detect_encoding(ogr_dataset, ogr_layer)
         or locale.getpreferredencoding()
     )
-
-    supports_transactions = OGR_L_TestCapability(ogr_layer, OLCTransactions)
-    if supports_transactions:
-        start_transaction(ogr_dataset, 0)
 
     for i in range(num_records):
         try:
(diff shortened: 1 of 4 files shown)


theroggy:
XLSX gave UTF-8 decoding errors when reading a file written after the change because it doesn't say it needs UTF-8 data (via the OLCStringsAsUTF8 capability).

I now hardcoded the XLSX needs UTF-8 data, but I also opened a PR upstream in GDAL: https://github.com/OSGeo/gdal/pull/9295

pyogrio/tests/test_geopandas_io.py
```diff
@@ -62,6 +62,25 @@ def spatialite_available(path):
         return False
 
 
+def test_read_csv_encoding(tmp_path):
+    # Write csv test file. Depending on the os this will be written in a different
+    # encoding: for linux and macos this is utf-8, for windows it is cp1252.
```
jorisvandenbossche, line 67: In one of the next Python versions, this will become utf-8 for windows as well. Do we want to have an additional test where we manually specify the encoding to test a specific one?

pyogrio/_io.pyx
```diff
@@ -1906,22 +1940,27 @@ def ogr_write(
         gdal_tz_offsets = {}
 
     ### Setup up dataset and layer
-    if not encoding:
-        encoding = locale.getpreferredencoding()
-
     layer_created = create_ogr_dataset_layer(
```
jorisvandenbossche, line 1943: This should be a change (or fix) in behaviour for Shapefiles, given that now with this PR, if encoding is not set, it will always use 'ISO-8859-1' (as fixed in `get_encoding_for_driver`)? While before it would have used UTF8 (or cp1252 on Windows) from `locale.getpreferredencoding()`?

(not saying is wrong or so, just trying to understand the code and the change ;))

pyogrio/_io.pyx
```diff
@@ -2032,7 +2070,7 @@ def ogr_write(
                             field_value = str(field_value)
 
                         try:
-                            value_b = field_value.encode("UTF-8")
+                            value_b = field_value.encode(encoding)
```
jorisvandenbossche, line 2073: I am not fully sure this should be changed. It's an old RFC, but according to https://gdal.org/development/rfc/rfc23_ogr_unicode.html#oftstring-oftstringlist-fields field _values_ (so not field _names_) in the GDAL API's are always UTF8?

pyogrio/_io.pyx
```diff
@@ -488,6 +504,16 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
...
         return "UTF-8"
 
+    if driver in ("XLSX", "ODS"):
+        # TestCapability for OLCStringsAsUTF8 for XLSX and ODS was False for new files
+        # being created for GDAL < 3.8.5. Once these versions of GDAL are no longer
+        # supported, this can be removed.
+        return "UTF-8"
+
+    if driver == "GeoJSONSeq":
+        # In old gdal versions, OLCStringsAsUTF8 wasn't advertised yet.
```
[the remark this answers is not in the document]
theroggy: No, I have been searching but based on the git history of gdal it has never been different... but the gdal 3.5 tests fail on it?

pyogrio/_io.pyx
theroggy: Yes indeed. Not sure if this is explicitly wanted, but this is [what fiona does](https://github.com/Toblerity/Fiona/blob/21ec544adceb95622de026eb06944e6671fd4d76/fiona/ogrext.pyx#L668) (if I understand correctly) and is also what ogr2ogr does based on the [shapefile driver docs](https://gdal.org/drivers/vector/shapefile.html#layer-creation-options).

pyogrio/_io.pyx
theroggy: It's indeed not 100% clear. Based on the next paragraph in that document, regarding the [olcstringsasutf8-capability-flag](https://gdal.org/development/rfc/rfc23_ogr_unicode.html#olcstringsasutf8-capability-flag), I interprete that for drivers/layers that return True for this flag, the field values should always be in UTF-8 (for most drivers), and for other drivers it should be in the encoding you want it to be written (e.g. for .csv). This is how I intended it to work.

What do you think?

pyogrio/tests/test_geopandas_io.py
theroggy: OK, the test now tests encodings "UTF-8", "cp1252" and None.

pyogrio/_io.pyx
brendan-ward: > for drivers/layers that return True for this flag, the field values should always be in UTF-8 (for most drivers), and for other drivers it should be in the encoding you want it to be written

My interpretation of the docs is in line with this.

pyogrio/_io.pyx
```diff
@@ -1791,6 +1823,8 @@ cdef create_ogr_dataset_layer(
         if driver == 'ESRI Shapefile':
             # Fiona only sets encoding for shapefiles; other drivers do not support
             # encoding as an option.
+            if encoding is None:
```
brendan-ward, line 1826: I'm not quite sure about the order of operations and the interactions around setting the `"ENCODING"` option for writing datasets.

I think the previous implementation was setting this to blank (which then I think means refer to the `.cpg` file), and using `UTF-8` for shapefiles if supported, falling back to `ISO-8859-1` when `UTF-8` is not supported.

Now, this will always use `ISO-8859-1` regardless of whether the capability for `UTF-8` exists.

This came up in for reading [#138](https://github.com/geopandas/pyogrio/issues/138#issuecomment-1179288076) and we fixed the order of operations for reading in #139.

brendan-ward (commented):
Thanks for working on this!  Still a bit unsure of the correct behavior for shapefiles when `encoding` is not provided by the user, but otherwise this looks reasonable.

pyogrio/_io.pyx
theroggy: In the previous implementation, right before calling create_ogr_dataset_layer this was done to the encoding:
``` py
    if not encoding:
        encoding = locale.getpreferredencoding()
```
So, in effect, when creating a shapefile, if the encoding wasn't specified, the file would be created as UTF-8 on linux/MacOS and in cp1552 for windows (the preferred encodings of the platform.

UTF-8 being supported or not isn't really a thing when creating a new file. There doesn't exist a `.cpg` file yet, do it can also not be checked whether it contains 'UTF-8' or not. So, the `.cpg` and the "UTF-8" capability is only relevant for reading.

theroggy:
> Thanks for working on this! Still a bit unsure of the correct behavior for shapefiles when `encoding` is not provided by the user, but otherwise this looks reasonable.

I'm not sure about the best way forward either. I now implemented it the same as the behaviour in fiona and the default of gdal, but e.g. just always using UTF-8 sound reasonable to me as well. 

Not sure why fiona and gdal default to the 1970's encoding, but possibly some applications e.g. don't use the ".cpg" properly when reading leading to more risk of compatibility issues?

jorisvandenbossche:
From an ESRI page (https://support.esri.com/en-us/knowledge-base/read-and-write-shapefile-and-dbase-files-encoded-in-var-000013192):

> Shapefiles can now be stored in UTF-8. However, shapefiles encoded in UTF-8 are only recognized in ArcMap, ArcCatalog and ArcGIS Pro.

Now, I don't know what this "only" means (i.e. what other ESRI products are not included in that list).

Anyway, given we were already using UTF-8 before and didn't yet get any complaints about that, I would personally keep that. That seems like a better default nowadays (while the default in fiona and GDAL probably stems from many years ago)

pyogrio/_io.pyx
jorisvandenbossche: Indeed, that sounds correct

jorisvandenbossche:
However, I would then maybe do that for all platforms, including Windows?

theroggy:
> However, I would then maybe do that for all platforms, including Windows?

I agree. If we would go for "UTF-8", I would also vote to do it for all platforms.

brendan-ward:
All right - let's go with `UTF-8` as the default for shapefiles on all platforms and revisit (by setting to `ISO-8859-1`) if we get errors reported by users.

theroggy:
UTF-8 is now the default for Shapefile on all platforms...

pyogrio/_io.pyx
```diff
@@ -480,6 +480,22 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
...
+    driver : str
+        driver to get the encoding for
+    
+    Returns
+    -------
+    str or None
+        the typical encoding if this is applicable for the driver, or None
+    """
     if driver == 'ESRI Shapefile':
         return 'ISO-8859-1'
```
brendan-ward, line 500: This should probably be set to `UTF-8` with a comment indicating that we are intentionally always setting `UTF-8` rather than the legacy value of `ISO-8859-1`.

pyogrio/_io.pyx
```diff
@@ -1914,22 +1948,27 @@ def ogr_write(
...
         &ogr_dataset, &ogr_layer,
     )
 
+    # Now the dataset and layer have been created, we can properly determine the
+    # encoding. It is derived from the user, from the dataset capabilities / type,
+    # or from the system locale
+    encoding = (
+        encoding
+        or detect_encoding(ogr_dataset, ogr_layer)
+        or locale.getpreferredencoding()
```
brendan-ward, line 1964: It looks like we always fall back to `locale.getpreferredencoding()` everywhere we are trying to get the encoding; what do you think about centralizing this in `detect_encoding`?

```Python
cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
    ...

    if OGR_L_TestCapability(ogr_layer, OLCStringsAsUTF8):
        return 'UTF-8'

    driver = get_driver(ogr_dataset)
    encoding = get_encoding_for_driver(driver)

   return encoding or locale.getpreferredencoding()
```

brendan-ward (commented):
Thanks for the updates!

Can you please add a test that Shapefile is always written to `UTF-8` by default (since it wasn't necessarily set that way on Windows before) unless encoding is passed by user.

pyogrio/_io.pyx
theroggy: At least for shapefiles, this code is now only reached when reading... and I think it is safer to keep the default for reading to ISO-... rather than UTF-8?

pyogrio/_io.pyx
theroggy: OK!

brendan-ward (approved):
Thanks @theroggy !  Apologies for the very slow final review!

Planning to merge once CI is green so that we can have this in place before changes needed for #380

PR #362, opened 2024-02-24
ENH: warn when reading a multilayer file without specifying layer

### LINKED ISSUE #355: ENH?: make `layer=` mandatory if multiple layers in a file?
At the moment, when a file being read contains > 1 layer and the `layer` parameter isn't specified, the first layer encountered in the file is being read.

This isn't really transparent behaviour, and I think it would be better if an error would be thrown when trying to read a multi-layer file without specifying `layer`. This isn't backwards compatible, but maybe now is the time still to change this?

An alternative that wouldn't break backwards compatibility would be to raise a warning.

Something to consider: what is a "layer": e.g. Geopackage files can also contain attribute tables which can be read by pyogrio. Are they to be considered as layers or can they be ignored if there is only one layer with geometries?

theroggy:
resolves #355 , by giving a warning if a multilayer file is being read without specifying the layer to be read.


pyogrio/_io.pyx
```diff
@@ -1126,6 +1126,14 @@ def ogr_read(
         if sql is None:
             # layer defaults to index 0
             if layer is None:
+                layers = ogr_list_layers_dataset(ogr_dataset)
+                if len(layers) > 1:
+                    warnings.warn(
+                        f"More than one layer found in '{path}', layer {layers[0][0]} "
+                        f"is read, available layers: {layers[:, 0].tolist()}. Specify "
+                        "layer parameter to avoid this warning.",
```
martinfleis, line 1134: This can be quite verbose. Can we try to use some shorter version? I don't think we need to print `path` (user knows what they're reading). Same for the last sentence. We could say something like "Unspecified layer. Using  {layers[0][0]}, available layers: {layers[:, 0].tolist()}". It tells the same but is much shorter and it implies that specifying the layer silences the warning.

theroggy: No problem to try making the warning shorter, but I'm not sure about removing the path. Because there are sometimes different code paths in pyogrio that lead to the warning being triggered, the stacklevel isn't always good enough to lead to the line number in the user's script being shown. In a larger script where many files are being read I can imagine that it can be a hunt to find the relevant file.

Maybe just print the file name? Something like: "Unspecified layer, using {layers[0][0]}. Available layers in file {filename}: {{layers[:, 0].tolist()}"

pyogrio/_io.pyx
```diff
@@ -1126,6 +1126,14 @@ def ogr_read(
         if sql is None:
             # layer defaults to index 0
             if layer is None:
+                layers = ogr_list_layers_dataset(ogr_dataset)
```
brendan-ward, line 1129: This block could probably be consolidated into a function:

```
cdef get_default_layer(OGRDataSourceH ogr_dataset):
    layers = ogr_list_layers_dataset(ogr_dataset)
    first_layer_name = layers[0][0]

    if len(layers) > 1:
        dataset_name = os.path.basename(OGR_DS_GetName(ogr_dataset).decode('UTF-8'))

        other_layer_names = ', '.join([f"'{l}'" for l in layers[1:, 0]])
        warnings.warn(
            f"More than one layer found in '{dataset_name}': '{first_layer_name}' (default), {other_layer_names}.  "
            "Specify layer parameter to avoid this warning.",
            stacklevel=2,
        )

    return first_layer_name
```
(not tested against all the various VSI filenames, but worked properly for the `test_fgdb_vsi` fixture)


You'll need to add the following to `_ogr.pxd` in the `cdef extern from "ogr_api.h":` block:

```
const char*     OGR_DS_GetName(OGRDataSourceH)
```


This returns a somewhat more succinct warning:

```
UserWarning: More than one layer found in 'sample.osm.pbf': 'points' (default), 'lines', 'multilinestrings', 'multipolygons', 'other_relations'.  Specify layer parameter to avoid this warning.
```

brendan-ward (commented):
Thanks for working on this!  A few suggestions...

pyogrio/_io.pyx
theroggy: That's indeed a lot cleaner. Done.

CHANGES.md
```diff
@@ -4,7 +4,11 @@
 
 ### Improvements
 
-- `read_arrow` and `open_arrow` now provide [GeoArrow-compliant extension metadata](https://geoarrow.org/extension-types.html), including the CRS, when using GDAL 3.8 or higher.
+-   `read_arrow` and `open_arrow` now provide [GeoArrow-compliant extension metadata]
+    (https://geoarrow.org/extension-types.html), including the CRS, when using GDAL 3.8 or higher.
+-   Warn when reading from a multilayer file without specifying a layer (#362).
+
+## 0.7.3 (???)
```
brendan-ward, line 11: This probably came from a merge, but can be removed

brendan-ward (approved):

pyogrio/_io.pyx
theroggy: The warning is a bit shorter now (based on suggestion of Brendan).

martinfleis (approved):
Thanks!

PR #365, opened 2024-02-26
ENH: add some missing properties to read info

theroggy:
Add some properties to read_info that I regularly use: layer_name, fid_column and geometry_name.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,10 +4,11 @@
 
 ### Improvements
 
--   Add some missing properties to `read_info` (#365)
+-   Add some missing properties to `read_info`, including layer name, geometryname
+    and FID column name (#365).
 -   `read_arrow` and `open_arrow` now provide
     [GeoArrow-compliant extension metadata](https://geoarrow.org/extension-types.html),
-    including the CRS, when using GDAL 3.8 or higher.
+    including the CRS, when using GDAL 3.8 or higher (#366).
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
--   Add some missing properties to `read_info` (#)
+-   Add some missing properties to `read_info` (#365)
 
 ## 0.7.3 (???)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
--   Add some missing properties to `read_info`, including layer name, geometryname
+-   Add some missing properties to `read_info`, including layer name, geometry name
     and FID column name (#365).
 -   `read_arrow` and `open_arrow` now provide
     [GeoArrow-compliant extension metadata](https://geoarrow.org/extension-types.html),
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## ???
+
+### Improvements
+
+-   Add some missing properties to `read_info` (#)
+
 ## 0.7.3 (???)
 
 ### Bug fixes
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1523,7 +1523,7 @@ def ogr_read_info(
         fields = get_fields(ogr_layer, encoding)
 
         meta = {
-            "name": get_string(OGR_L_GetName(ogr_layer)),
+            "layer_name": get_string(OGR_L_GetName(ogr_layer)),
             "crs": get_crs(ogr_layer),
             "encoding": encoding,
             "fields": fields[:,2], # return only names
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1523,14 +1523,17 @@ def ogr_read_info(
         fields = get_fields(ogr_layer, encoding)
 
         meta = {
-            'crs': get_crs(ogr_layer),
-            'encoding': encoding,
-            'fields': fields[:,2], # return only names
-            'dtypes': fields[:,3],
-            'geometry_type': get_geometry_type(ogr_layer),
-            'features': get_feature_count(ogr_layer, force_feature_count),
-            'total_bounds': get_total_bounds(ogr_layer, force_total_bounds),
-            'driver': get_driver(ogr_dataset),
+            "name": get_string(OGR_L_GetName(ogr_layer)),
+            "crs": get_crs(ogr_layer),
+            "encoding": encoding,
+            "fields": fields[:,2], # return only names
+            "dtypes": fields[:,3],
+            "fid_column": get_string(OGR_L_GetFIDColumn(ogr_layer)),
+            "geometry_name": get_string(OGR_L_GetGeometryColumn(ogr_layer)),
+            "geometry_type": get_geometry_type(ogr_layer),
+            "features": get_feature_count(ogr_layer, force_feature_count),
+            "total_bounds": get_total_bounds(ogr_layer, force_total_bounds),
+            "driver": get_driver(ogr_dataset),
             "capabilities": {
                 "random_read": OGR_L_TestCapability(ogr_layer, OLCRandomRead) == 1,
                 "fast_set_next_by_index": OGR_L_TestCapability(ogr_layer, OLCFastSetNextByIndex) == 1,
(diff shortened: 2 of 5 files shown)


pyogrio/_io.pyx
```diff
@@ -1523,23 +1523,26 @@ def ogr_read_info(
...
         meta = {
-            'crs': get_crs(ogr_layer),
-            'encoding': encoding,
-            'fields': fields[:,2], # return only names
-            'dtypes': fields[:,3],
-            'geometry_type': get_geometry_type(ogr_layer),
-            'features': get_feature_count(ogr_layer, force_feature_count),
-            'total_bounds': get_total_bounds(ogr_layer, force_total_bounds),
-            'driver': get_driver(ogr_dataset),
+            "name": get_string(OGR_L_GetName(ogr_layer)),
```
brendan-ward, line 1526:
```suggestion
            "layer_name": get_string(OGR_L_GetName(ogr_layer)),
```

pyogrio/_io.pyx
```diff
@@ -1523,23 +1523,26 @@ def ogr_read_info(
...
+            "crs": get_crs(ogr_layer),
+            "encoding": encoding,
+            "fields": fields[:,2], # return only names
+            "dtypes": fields[:,3],
+            "fid_column": get_string(OGR_L_GetFIDColumn(ogr_layer)),
+            "geometry_name": get_string(OGR_L_GetGeometryColumn(ogr_layer)),
```
brendan-ward, line 1532:
```suggestion
            "geometry_field": get_string(OGR_L_GetGeometryColumn(ogr_layer)),
```

pyogrio/_io.pyx
```diff
@@ -1523,23 +1523,26 @@ def ogr_read_info(
...
-            'geometry_type': get_geometry_type(ogr_layer),
-            'features': get_feature_count(ogr_layer, force_feature_count),
-            'total_bounds': get_total_bounds(ogr_layer, force_total_bounds),
-            'driver': get_driver(ogr_dataset),
+            "name": get_string(OGR_L_GetName(ogr_layer)),
+            "crs": get_crs(ogr_layer),
+            "encoding": encoding,
+            "fields": fields[:,2], # return only names
+            "dtypes": fields[:,3],
+            "fid_column": get_string(OGR_L_GetFIDColumn(ogr_layer)),
```
brendan-ward, line 1531:
```suggestion
            "fid_field": get_string(OGR_L_GetFIDColumn(ogr_layer)),
```
(for consistency, since we refer to the list of columns as "fields")

pyogrio/core.py
```diff
@@ -217,6 +217,14 @@ def read_info(
     driver or if the data source is nonspatial. You can force it to be calculated using
     the ``force_total_bounds`` parameter.
 
+    ``fid_column`` is the name of the FID field in the data source, if the FID is
+    physically stored (e.g. in GPKG). If the FID is just a sequence, ``fid_column``
+    will be "" (e.g. ESRI Shapefile).
+
+    ``geometry_name`` is the name of the field where the main geometry is stored in the
+    data data source, if the column can by customized (e.g. in GPKG). If no custom name
```
brendan-ward, line 225:
```suggestion
    data data source, if the field name can by customized (e.g. in GPKG). If no custom name
```

pyogrio/tests/test_core.py
```diff
@@ -354,23 +354,39 @@ def test_read_bounds_bbox_intersects_vs_envelope_overlaps(naturalearth_lowres_al
...
+        assert meta["geometry_type"] == "MultiPolygon"
+        assert meta["driver"] == "GPKG"
+    elif naturalearth_lowres.suffix == ".shp":
+        # fid_column == "" for formats where fid is not physically stored
+        assert meta["fid_column"] == ""
+        # geometry_name == "" for formats where geometry column name can be customized
+        assert meta["geometry_name"] == ""
+        assert meta["geometry_type"] == "Polygon"
+        assert meta["driver"] == "ESRI Shapefile"
+        # for GPKG, is only set to True since gdal 3.8
```
brendan-ward, line 385: Since the comment applies to GPKG, it seems like this assert should be copied there as well (with the comment), and with a guard for the GDAL version, whereas it should always be `True` for shapefile.

brendan-ward (commented):
Thanks!  A few suggestions...

pyogrio/_io.pyx
jorisvandenbossche: We already use `geometry_name` in some of the Arrow code paths (`open_arrow` uses that key in the meta dict it returns). Not necessarily a reason to change it here, but it would give a slight inconsistency.

We should maybe also think about what this key should be if there are multiple geometry fields.

pyogrio/_io.pyx
jorisvandenbossche: > since we refer to the list of columns as "fields"

Yes, but on the other hand the geometry and FID columns/fields are not(?) part of the list of (attribute) fields in the `"fields"` entry.

pyogrio/core.py
```diff
@@ -217,6 +217,14 @@ def read_info(
     driver or if the data source is nonspatial. You can force it to be calculated using
     the ``force_total_bounds`` parameter.
 
+    ``fid_column`` is the name of the FID field in the data source, if the FID is
+    physically stored (e.g. in GPKG). If the FID is just a sequence, ``fid_column``
+    will be "" (e.g. ESRI Shapefile).
+
+    ``geometry_name`` is the name of the field where the main geometry is stored in the
```
jorisvandenbossche, line 224: Potentially confusing aspect here is that `read_dataframe` actually does not honor that name, but always uses "geometry"

pyogrio/_io.pyx
brendan-ward: Ah yes, fine to leave it as `geometry_name` then (for now).

To handle multiple geometries, would we want this to be a list instead? (and maybe also be a plural, e.g., `geometry_names`: [...])

So `["geometry"]` for single-geometry drivers and datasets with only geometry, and `["geom", "geom2", ...]` for multiple geometries?  Then update the Arrow implementation to use the first by default?

(still haven't gotten my head fully around multiple geometry datasets)

pyogrio/core.py
brendan-ward: Presumably it should?  Which wouldn't be a breaking change, because GeoPandas would also return it as the `.geometry` attribute of a GeoDataFrame, regardless of the name we use for the geometry field when constructing the GeoDataFrame?

pyogrio/_io.pyx
theroggy: Typically (sometimes) there is also a "primary" geometry column I think...

pyogrio/_io.pyx
theroggy: Yes, because I feel there are arguments possible in all directions, I just followed the GDAL naming, but I don't have explicit preferences...

pyogrio/core.py
jorisvandenbossche: Users still do `df["geometry"]` as well, in addition to `df.geometry`, so changing the geometry name we return in `read_dataframe` would be quite a big change, I think

theroggy: > Users still do `df["geometry"]` as well, in addition to `df.geometry`, so changing the geometry name we return in `read_dataframe` would be quite a big change, I think

I can confirm I use `df["geometry"]` regularly as well...

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## ???
+
+### Improvements
+
+-   Add some missing properties to `read_info` (#365)
```
brendan-ward, line 7:
```suggestion
-   Add additional properties to `read_info`, including layer name, geometry name, and FID column name (#365)
```

pyogrio/tests/test_core.py
```diff
@@ -354,23 +355,41 @@ def test_read_bounds_bbox_intersects_vs_envelope_overlaps(naturalearth_lowres_al
...
+        assert meta["geometry_name"] == "geom"
+        assert meta["geometry_type"] == "MultiPolygon"
+        assert meta["driver"] == "GPKG"
+        if GDAL_GE_38:
+            # this capability is only True for GPKG if GDAL >= 3.8
+            assert meta["capabilities"]["fast_set_next_by_index"] is True
+    elif naturalearth_lowres.suffix == ".shp":
+        # fid_column == "" for formats where fid is not physically stored
+        assert meta["fid_column"] == ""
+        # geometry_name == "" for formats where geometry column name can be customized
```
brendan-ward, line 385:
```suggestion
        # geometry_name == "" for formats where geometry column name cannot be customized
```

pyogrio/_io.pyx
brendan-ward: Ok - let's stick to what you have.

CHANGES.md
```diff
@@ -4,10 +4,11 @@
 
 ### Improvements
 
--   Add some missing properties to `read_info` (#365)
+-   Add some missing properties to `read_info`, including layer name, geometryname
```
brendan-ward, line 7:
```suggestion
-   Add some missing properties to `read_info`, including layer name, geometry name
```

brendan-ward (approved):
Thanks @theroggy , apologies for the slow review!

Issue #345, opened 2024-01-24
Read Arrow: Support for new options as of GDAL 3.8


kylebarron:
I was just re-reading the [`GetArrowStream` API docs](https://gdal.org/api/vector_c_api.html#_CPPv420OGR_L_GetArrowStream9OGRLayerHP16ArrowArrayStreamPPc) and saw that there are two new options as of GDAL 3.8: 

- TIMEZONE="unknown", "UTC", "(+|:)HH:MM" or any other value supported by Arrow. (GDAL >= 3.8)
- GEOMETRY_METADATA_ENCODING=OGC/GEOARROW (GDAL >= 3.8).

It would be great to support these. The metadata encoding would be useful for my purposes. I don't use timezone information as often, but someone probably would like that.

brendan-ward:
Looks like we'd need to add a keyword on `ogr_open_arrow` to set the geometry encoding, maybe `geometry_encoding="wkb"|"geoarrow"`
(not finding "ogc" as a meaningful value here compared to "wkb" which is specific).

brendan-ward:
Oops, I mistakenly conflated the actual geometry encoding - which looks like it is either WKB or native (which may be GeoArrow) without a way to specifically request GeoArrow geometry encoding - with the _metadata_ that sets extension types (which I don't entirely follow here). 

Does the `geoarrow.wkb` extension type mean that the actual geometry encoding is still (OGC) WKB, or is it serialized GeoArrow bytes.  Apologies for not being more up to speed on GeoArrow!

kylebarron:
> Looks like we'd need to add a keyword on `ogr_open_arrow` to set the geometry encoding, maybe `geometry_encoding="wkb"|"geoarrow"` (not finding "ogc" as a meaningful value here compared to "wkb" which is specific).

I agree this is confusing 🙂 . The key is that the switch does _not_ change the actual encoding of the geometry. The geometries are **always** WKB. Rather, the switch only changes the **metadata** applied to the Arrow geometry column. As it is, the geometry column is always loaded with an Arrow extension name of `ogc.wkb`, which is ["non-standard"](https://geoarrow.org/extension-types#extension-names) (if you call GeoArrow a standard 😅). 

[...]

PR #366, opened 2024-02-26
Read with GeoArrow metadata for GDAL >= 3.8

kylebarron:
Resolves half of https://github.com/geopandas/pyogrio/issues/345. @jorisvandenbossche mentioned in https://github.com/geopandas/pyogrio/issues/345#issuecomment-1909638656 that he'd suggest making the change to geoarrow metadata. I'd also propose that we make this the default.

A test is added to verify that the metadata on the geometry field is set. The name should be `geoarrow.wkb` and the extension metadata should include a CRS field with PROJJSON, which I assert is EPSG:4326 for this test dataset.

@brendan-ward would you still prefer having a flag for the user to set this? Ref https://github.com/geopandas/pyogrio/issues/345#issuecomment-1908878875

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 # CHANGELOG
 
-## 0.7.3 (???)
+## 0.8.0 (???)
+
+### Improvements
+
+- `read_arrow` and `open_arrow` now provide [GeoArrow-compliant extension metadata](https://geoarrow.org/extension-types.html), including the CRS, when using GDAL 3.8 or higher.
 
 ### Bug fixes
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -783,7 +783,7 @@ cdef process_fields(
             data[i] = bin_value[:ret_length]
 
         elif field_type == OFTDateTime or field_type == OFTDate:
-            
+
             if datetime_as_string:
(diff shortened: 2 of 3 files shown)


kylebarron:
This will be exciting to get working with [lonboard](https://github.com/developmentseed/lonboard), a visualization library I'm working on. I added [support for parsing from `geoarrow.wkb` input](https://github.com/developmentseed/lonboard/pull/372) and with the CRS information it'll do [multithreaded reprojection to WGS84](https://github.com/developmentseed/lonboard/pull/337), parallelizing over each Arrow chunk (conveniently GDAL will handle the chunking into row groups). So it should be really fast to plot (though today it still goes through shapely for the WKB -> GeoArrow parsing, which is a relative bottleneck, though still certainly fast enough)

jorisvandenbossche (approved):
Thanks! 
I am personally fine with just defaulting to geoarrow without providing an explicit keyword to control this (in the end it's just for the metadata, not the actual data you get)

brendan-ward (approved):
Thanks!

Seems reasonable to go with geoarrow as the default and no need for a keyword, especially since it makes downstream handling easier.

brendan-ward:
Could you please add an entry to the changelog?

PR #368, opened 2024-03-01
ENH: Write Unknown cartesian CRS when saving gdf without a CRS to GPKG

### LINKED ISSUE #367: Write Unknown cartesian CRS when saving gdf without a CRS to GPKG
xref [GeoPandas #3019](https://github.com/geopandas/geopandas/issues/3019)

theroggy:
resolves #367

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -537,14 +537,17 @@ def write_dataframe(
                 geometry_type = f"{geometry_type} Z"
 
     crs = None
-    if geometry_column is not None and geometry.crs:
-        # TODO: this may need to be WKT1, due to issues
-        # if possible use EPSG codes instead
-        epsg = geometry.crs.to_epsg()
-        if epsg:
-            crs = f"EPSG:{epsg}"
+    if geometry_column is not None:
+        if geometry.crs:
+            # TODO: this may need to be WKT1, due to issues
+            # if possible use EPSG codes instead
+            epsg = geometry.crs.to_epsg()
+            if epsg:
+                crs = f"EPSG:{epsg}"
+            else:
+                crs = geometry.crs.to_wkt(WktVersion.WKT1_GDAL)
         else:
-            crs = geometry.crs.to_wkt(WktVersion.WKT1_GDAL)
+            crs = 'LOCAL_CS["Undefined Cartesian SRS"]'
 
     # If there is geometry data, prepare it to be written
     if geometry_column is not None:
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -910,6 +910,24 @@ def test_write_dataframe_gpkg_multiple_layers(tmp_path, naturalearth_lowres):
     )
 
 
+def test_write_dataframe_gpkg_no_crs(tmp_path, naturalearth_lowres):
+    input_gdf = gp.GeoDataFrame(geometry=[Point(0, 1)], crs=None)
+    output_path = tmp_path / "test.gpkg"
+
+    write_dataframe(input_gdf, output_path)
+
+    assert os.path.exists(output_path)
+    result_gdf = read_dataframe(output_path)
+
+    # None crs is replaced by "Undefined Cartesian SRS" in GPKG.
+    input_gdf.crs = 'LOCAL_CS["Undefined Cartesian SRS"]'
+    assert_geodataframe_equal(
+        result_gdf,
+        input_gdf,
+        # check_index_type=False,
+    )
+
+
 @pytest.mark.parametrize("ext", ALL_EXTS)
 def test_write_dataframe_append(tmp_path, naturalearth_lowres, ext):
     if ext == ".fgb" and __gdal_version__ <= (3, 5, 0):


pyogrio/geopandas.py
```diff
@@ -537,14 +537,19 @@ def write_dataframe(
...
+            # if possible use EPSG codes instead
+            epsg = geometry.crs.to_epsg()
+            if epsg:
+                crs = f"EPSG:{epsg}"
+            else:
+                crs = geometry.crs.to_wkt(WktVersion.WKT1_GDAL)
+        elif driver == "GPKG":
+            # In GPKG, None crs must be replaced by "Undefined Cartesian SRS", otherwise
+            # the default "Undefined geographic SRS" will be used.
+            crs = 'LOCAL_CS["Undefined Cartesian SRS"]'
```
martinfleis, line 552: Is this enough? My understanding was that this is required:

```
ENGCRS["Undefined Cartesian SRS with unknown unit",
    EDATUM["Unknown engineering datum"],
    CS[Cartesian,2],
        AXIS["X",unspecified,
            ORDER[1],
            LENGTHUNIT["unknown",0]],
        AXIS["Y",unspecified,
            ORDER[2],
            LENGTHUNIT["unknown",0]]]
```

See the dicsussion in https://github.com/r-spatial/sf/issues/2049.

theroggy: I'm  not an expert on projections, but as far as I can tell both are equivalent: LOCAL_CS["Undefined Cartesian SRS"] is just a shorter name for the full wkt.

I did a manual check on this before, but to be sure I added an explicit check in the test that the srs_id in the GPKG written is -1, which is the main thing we are trying to do here.

martinfleis: The difference is that `LOCAL_CS["Undefined Cartesian SRS"]` assumes meters as units and axes as easting and northing, which is incorrect:

```
<Engineering CRS: LOCAL_CS["Undefined Cartesian SRS",UNIT["Meter",1] ...>
Name: Undefined Cartesian SRS
Axis Info [cartesian]:
- [east]: Easting (Meter)
- [north]: Northing (Meter)
```

The WKT above is explicit about making no assumptions about anything, because we don't know anything.

theroggy: I think using that WKT does not comply to the GeoPackage specs: https://www.geopackage.org/spec/#r11. Using that WKT leads to a new custom CRS being "created" in the GeoPackage with srs_id = 10000 (or whatever is the first unused srs_id in that range in the geopackage) instead of using srs_id = -1, as should be used according to the specs.

I wonder what is the rational that they went for this solution in r-spatial? I don't really found one in the post above. Maybe it should be changed in GDAL that `ENGCRS["Undefined Cartesian SRS with unknown unit"` is (also) mapped to srs_id = -1 in geopackage?

I did some extra tests and proj does seem to treat them all as equivalent, so I suppose that in practice it probably won't matter to much with the different crs's being interused?

```
from pyproj import CRS

crs_undefined = CRS('LOCAL_CS["Undefined Cartesian SRS"]')
crs_undefined_unknown = CRS('LOCAL_CS["Undefined Cartesian SRS with unknown unit"]')
try:
    crs_undefined_unknown_eng = CRS(
        'ENGCRS["Undefined Cartesian SRS with unknown unit"]'
    )
except Exception as ex:
    print(f"Exception raised: {ex}")

crs_undefined_unknown_eng_wkt = """
    ENGCRS["Undefined Cartesian SRS with unknown unit",
        EDATUM["Unknown engineering datum"],
        CS[Cartesian,2],
            AXIS["X",unspecified,
                ORDER[1],
                LENGTHUNIT["unknown",0]],
            AXIS["Y",unspecified,
                ORDER[2],
                LENGTHUNIT["unknown",0]]]
"""
crs_undefined_unknown_eng2 = CRS(crs_undefined_unknown_eng_wkt)

# Check differences
print(f"{crs_undefined_unknown.equals(crs_undefined_unknown_eng2)=}")
print(f"{crs_undefined_unknown.to_wkt() == crs_undefined_unknown_eng2.to_wkt()=}")

print(f"{crs_undefined_unknown.equals(crs_undefined)=}")
print(f"{crs_undefined_unknown.to_wkt() == crs_undefined.to_wkt()=}")

print(f"{crs_undefined_unknown_eng2.equals(crs_undefined)=}")
print(f"{crs_undefined_unknown_eng2.to_wkt() == crs_undefined.to_wkt()=}")
```
Output:
```
Exception raised: Invalid projection: ENGCRS["Undefined Cartesian SRS with unknown unit"]: (Internal Proj Error: proj_create: Missing EDATUM / ENGINEERINGDATUM node)
crs_undefined_unknown.equals(crs_undefined_unknown_eng2)=True
crs_undefined_unknown.to_wkt() == crs_undefined_unknown_eng2.to_wkt()=False
crs_undefined_unknown.equals(crs_undefined)=True
crs_undefined_unknown.to_wkt() == crs_undefined.to_wkt()=False
crs_undefined_unknown_eng2.equals(crs_undefined)=True
crs_undefined_unknown_eng2.to_wkt() == crs_undefined.to_wkt()=False
```
```

Comment (martinfleis) [reply pyogrio/geopandas.py:552]: Tricky. When using the WKT, it correctly roundtrips with all the fields Unknown but the srid is 100000. I am not sure which is better. 

@edzer are you sure the way this is handled in {sf} is actually correct? Meaning it correctly roundtrips __and__ follows the spec?

Comment (theroggy) [reply pyogrio/geopandas.py:552]: Reference to GDAL code treating the srs_id=-1, as found by @jorisvandenbossche :
- reading: https://github.com/OSGeo/gdal/pull/3286
- writing: https://github.com/OSGeo/gdal/pull/3312

**Comment by edzer** [comment]
> @edzer are you sure the way this is handled in {sf} is actually correct? Meaning it correctly roundtrips and follows the spec?

@martinfleis which spec are you referring to?

**Comment by martinfleis** [comment]
@edzer I think I meant GeoPackage spec. From what I understand when @theroggy explained that, undefined SRS shall be saved as -1. But if we replicate the WKT from the sf issue linked above, it is not the case and it is saved as a custom SRS instead with ID 10000 or something like that (@theroggy please correct me here).

**Comment by edzer** [comment]
IIRC, the problem with the undefined SRS in GeoPackages is that they are assumed (and, after roundtrip read as) geographic coordinates with undefined datum. If you want to specify they are Cartesian but otherwise unknown, you can use the WKT
```
ENGCRS["Undefined Cartesian SRS with unknown unit",
  EDATUM["Unknown engineering datum"],
  CS[Cartesian, 2],
  AXIS["X",unspecified,ORDER[1],LENGTHUNIT["unknown",0]],   
  AXIS["Y",unspecified,ORDER[2],LENGTHUNIT["unknown",0]]
]
```
which is what R package `sf` writes to geopackages when the CRS is NA: https://github.com/r-spatial/sf/blob/main/R/read.R#L548

edzer:
See also https://github.com/r-spatial/sf/issues/2049#issuecomment-1333937050

martinfleis:
> the undefined SRS in GeoPackages is that they are assumed (and, after roundtrip read as) geographic coordinates with undefined datum. If you want to specify they are Cartesian but otherwise unknown, you can use the WKT

Undefined in sense of SRS=0 is assumed geographic but SRS=-1 is assumed Cartesian. Citing Requirement 11 from the version [1.4 of the spec](https://docs.ogc.org/is/12-128r19/12-128r19.html):

> The record with an srs_id of -1 SHALL be used for undefined Cartesian coordinate reference systems. The record with an srs_id of 0 SHALL be used for undefined geographic coordinate reference systems.

That is where I think that the sf solution with custom WKT is not exactly reflecting the spec.

GDAL correctly understands -1 as Cartesian but, for unknown reasons, sets units to meters [here](https://github.com/OSGeo/gdal/blob/bb520ce83fb118078605b05c9093e68331716a7e/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp#L279-L283). We discusses on the dev call earlier today that this should probably be fixed to resolve as Undefined Cartesian SRS with unknown units rather than with meters. Then setting SRS to -1 should do what we want it to without specifying custom WKT.

edzer:
So, maybe raise an issue with GDAL?

theroggy:
As planned in the community meeting and suggested by @edzer , I opened an issue to discuss this further in GDAL: https://github.com/OSGeo/gdal/issues/9580

brendan-ward:
Just checking back on the status of this PR; it looks like this is potentially waiting on upstream changes to GDAL possibly in 3.9?  But then what is the path forward on this for GDAL < 3.9?

PR #369, opened 2024-03-02
TST: add test case to write GeoDataFrame without geometry column

theroggy:
Tested this in the context of documenting known difference between Fiona and Pyogrio... so better to add it to the test permanently.

Or... based on feedback in https://github.com/geopandas/geopandas/pull/3206 is this something that works by accident and isn't supposed to, or not guaranteed that it will keep working?


brendan-ward (approved):
Thanks!

PR #370, opened 2024-03-03
BUG: undefined variable in timezone logic

m-richards:
I don't know how I missed this in #253, but there's a logical error if an exception is thrown in this block.

I'm not quite sure how to test this - I only noticed the error in my geopandas dev environment where I had pyogrio 0.7.2 with pandas dev - so I was getting the exception fixedby #348 (not part of 0.7.2), not a future error due to  mixed time zones - which is what this try except was originally for.

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -50,7 +50,7 @@ def _try_parse_datetime(ser):
         try:
             res = pd.to_datetime(ser, **datetime_kwargs)
         except Exception:
-            pass
+            res = ser
     # if object dtype, try parse as utc instead
     if res.dtype == "object":
         try:


brendan-ward (approved):
Thanks for catching this!

Going to merge now so as to avoid this remaining undefined.  However, does this also indicate that instead of catching all exceptions (via `except Exception`) that there should be different handling for the future error of mixed time zones?

m-richards:
> Thanks for catching this!
> 
> Going to merge now so as to avoid this remaining undefined. However, does this also indicate that instead of catching all exceptions (via `except Exception`) that there should be different handling for the future error of mixed time zones?

I can have another look but I didn't think this had been done yet - the error I saw was for expiring errors="ignore" - which now throws an assertionerror, I don't imagine that convention would be copied

martinfleis:
This is now showing up in https://github.com/geopandas/geopandas/pull/3223. Might be worth trying to get a version with this out sooner than later to avoid issues when testing against geopandas main (and to get green CI in geopandas).

PR #371, opened 2024-03-04
BLD: update circleci image for arm wheel building

jorisvandenbossche:
See https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177


brendan-ward (approved):
Thanks!

Looking forward to Arm64 runners coming natively to Github (I think Mac M1 runners are available now, but Linux Arm64 runners not yet)

Issue #380, opened 2024-04-01
GBK encoded SHP file, read exception.


stonereese:
Windows 10 professional 22H2 19045.4170

pyogrio == 0.7.2
GDAL == 3.8.4
fiona == 1.9.5

When using pyogrio.read_dataframe() to read a shp file, if the encoding of the dbf file is gbk, specify the parameter encoding='gbk' or encoding='cp936'. There are two exceptional situations encountered: when specifying encoding='cp936' in fiona, there are no similar issues:
1. An encoding error will be reported: UnicodeDecodeError: 'gbk' codec can't decode byte 0xaf in position 17: illegal multibyte sequence. Specifying encoding as utf8 allows for normal reading of Chinese characters in field values;
2. No error is reported, but garbled characters appear in field values.

brendan-ward:
Thanks for the report.  Are you able to share a small sample zip file (all files related to your shp) that reproduces this issue?

Is there a corresponding `.cpg` file present alongside your shp, and if so, what are the contents of that file?

stonereese:
> Thanks for the report. Are you able to share a small sample zip file (all files related to your shp) that reproduces this issue?
> 
> Is there a corresponding `.cpg` file present alongside your shp, and if so, what are the contents of that file?

The data mentioned above do not have cpg files. Most of the SHP data without a CPG file do not have garbled characters.These several SHP data are exceptions.  Unzip the shp data in the attachment (theoretically they are all encoded in gbk). When using read_dataframe() to read, specifically: for file 01.shp, if encoding parameter is not set or set as 'utf8', there will be no garbled characters; setting encoding as 'cp936' will result in a coding error. For file 02.shp, setting encoding as 'cp936' will not cause garbled characters, while setting it as 'utf8' will. For file 03.shp, both setting encoding as 'cp936' and 'utf8' will result in garbled characters.
[encode_error_shp.zip](https://github.com/geopandas/pyogrio/files/14833610/encode_error_shp.zip)

stonereese:
> Thanks for the report.  Are you able to share a small sample zip file (all files related to your shp) that reproduces this issue?
> 
> Is there a corresponding `.cpg` file present alongside your shp, and if so, what are the contents of that file?

Hello, have you downloaded the zip attachment above? If you have, I will delete it.

brendan-ward:
Actually, I'm finding that `03.shp` reads fine with `cp936` encoding, though I thought it produced garbled characters too when I tried it last night.

[...]

PR #384, opened 2024-04-05
FIX: correctly use GDAL auto-decoding of shapefiles when encoding is set

brendan-ward:
As described in #380, GDAL attempts to auto-detect the native encoding of a shapefile and will automatically use that to decode to UTF-8 before returning strings to us.  This runs into conflicts where the user provides `encoding` because we then apply the user's `encoding` to UTF-8 text, which produces incorrect text.

This sets `SHAPE_ENCODING=""` ([GDAL config option](https://gdal.org/drivers/vector/shapefile.html#configuration-options)) to disable auto-decoding _before_  opening the shapefile.  Setting the dataset open option `ENCODING=""` also works, but is specific to Shapefiles, and we don't know the path resolves to a Shapefile until _after_ opening it - which is too late to set the option.

This now correctly handles `encoding="cp936"` for the shapefiles described in #380.  I'm still trying to create some test cases here, but running into issues saving to the correct native encoding to prove that.


theroggy:
> This sets `SHAPE_ENCODING=""` ([GDAL config option](https://gdal.org/drivers/vector/shapefile.html#configuration-options)) to disable auto-decoding _before_ opening the shapefile. Setting the dataset open option `ENCODING=""` also works, but is specific to Shapefiles, and we don't know the path resolves to a Shapefile until _after_ opening it - which is too late to set the option.

The only disadvantage I see from using the config option vs the dataset open option is that it can lead to inter-thread side effects. So if someone in one thread uses the `encoding` parameter, all other threads running at the same time to read/write data can be influenced by this and get encoding issues. 
This can be avoided by first opening the file to check the driver when encoding is specified to then be able to use the open option... but I do egree it sounds like a far fetched scenario...

brendan-ward:
Fiona used to use `CPLSetThreadLocalConfigOption` to set this option specifically within a thread; we can do that here too.

My concern with opening the file twice is that it might incur a lot of overhead for remote sources (e.g., shapefile in a zip on S3), but I don't have remote examples easy at hand to measure / prove that increase in overhead, it's just a guess.  In contrast, setting the config option (so long as it doesn't negatively clobber other threads) _should_ be low overhead.

stonereese:
> Fiona used to use `CPLSetThreadLocalConfigOption` to set this option specifically within a thread; we can do that here too.
> 
> My concern with opening the file twice is that it might incur a lot of overhead for remote sources (e.g., shapefile in a zip on S3), but I don't have remote examples easy at hand to measure / prove that increase in overhead, it's just a guess. In contrast, setting the config option (so long as it doesn't negatively clobber other threads) _should_ be low overhead.

Thank you so much for your help. When I noticed garbled text in the feedback, I realized that I forgot to mention that the `shp` file also displayed garbled text when using the `sql` parameter in `read_dataframe()`, regardless of the specified `encoding` parameter. I believe your fix has also resolved this issue. I am currently on vacation today, but I will verify it tomorrow.

brendan-ward:
Unfortunately I'm finding that using `sql` with a non-UTF-8 encoded file is returning incorrect column names / values when `encoding` is not also specified, even if GDAL detects the correct native encoding and should be decoding that for us.  We might be missing a decoding step somewhere; investigating now.

I'm also finding that it is hard to use non-UTF-8 encodings with `use_arrow=True` because we don't decode the individual values; the raw values go directly into an Arrow table and immediately break certain things because they can't be decoded nicely.  I was able to work around this for shapefiles by letting GDAL do the decoding for us, but I'm not sure how to do this for other drivers.  Writing non-UTF values for some of those drivers is likely invalid, so it might not be an issue for those, but it seems theoretically possible for some of them (e.g., GeoJSON).

We can either disable the `encoding` option altogether for `ogr_open_arrow` - though this _does_ work for shapefiles because GDAL decodes those to UTF-8 for us (per [e738f7b](https://github.com/geopandas/pyogrio/pull/384/commits/e738f7b8aec3ca540c05ece7f022cbe087510e73)).

@jorisvandenbossche I'm unfamiliar with handling of non-UTF-8 column names / text values in pyarrow - should we even be trying to allow non-UTF-8 content through the Arrow API?  I'm not seeing how we would intercept and decode them on our end like we do for the non-Arrow API.

brendan-ward:
So it turns out that - at least for shapefiles - the `OGRLayer` instance returned when using an SQL always fails the test for `UTF-8` capabilities - even if there is a valid `.cpg` file present.

The workaround was to always test capabilities for shapefiles on the only layer that is present in the shapefile (e.g., the base layer underneath the SQL query layer).

I'm not sure the degree to which this might be present for other drivers; I think most of those already return true when testing for `UTF-8` capabilities, so we didn't run into the same issue.

theroggy:
> Fiona used to use `CPLSetThreadLocalConfigOption` to set this option specifically within a thread; we can do that here too.

Yep, that would be great.

> My concern with opening the file twice is that it might incur a lot of overhead for remote sources (e.g., shapefile in a zip on S3), but I don't have remote examples easy at hand to measure / prove that increase in overhead, it's just a guess. In contrast, setting the config option (so long as it doesn't negatively clobber other threads) _should_ be low overhead.

Hmm... didn't considered this, also don't have any experience with using remote files.

If the overhead would be in the driver detection, it can be solved by specifying in GDALOpenEx already that it is an "ESRI Shapefile", so the driver detection is avoided/minimized. If this would be a significant overhead it might be a good idea in general to offer the option to specify the driver(s) to take in consideration to open the file.

theroggy:
> So it turns out that - at least for shapefiles - the `OGRLayer` instance returned when using an SQL always fails the test for `UTF-8` capabilities - even if there is a valid `.cpg` file present.

Sounds like a bug in GDAL? The problem occurs for both sql dialects ('OGRSQL' and 'SQLITE')?

> The workaround was to always test capabilities for shapefiles on the only layer that is present in the shapefile (e.g., the base layer underneath the SQL query layer).
> 
> I'm not sure the degree to which this might be present for other drivers; I think most of those already return true when testing for `UTF-8` capabilities, so we didn't run into the same issue.

brendan-ward:
> Sounds like a bug in GDAL? 

Not sure, it could also be a bug in how we are trying to use it on our end in this case.  The GDAL Python bindings also return `False` for the capability but still return the correctly decoded value:

```python
from osgeo import ogr

drv = ogr.GetDriverByName("ESRI Shapefile")
ds = drv.Open("/tmp/test.shp", 0)

lyr = ds.GetLayerByIndex(0)
print(f"Supports UTF-8: {lyr.TestCapability(ogr.OLCStringsAsUTF8)}")
# True
print(lyr.schema[0].name)
# 中文

lyr = ds.ExecuteSQL("select * from test where \"中文\" = '中文' ", None, "")
print(f"Supports UTF-8: {lyr.TestCapability(ogr.OLCStringsAsUTF8)}")
# False
print(lyr.schema[0].name)
# 中文
```

The Python bindings are a bit hard to trace through, I'm still trying to find where it determines the schema / field name in this case to see if it is doing something different than we are here.

pyogrio/_io.pyx
```diff
@@ -486,6 +486,9 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
         # read without recoding. Hence, it is up to you to supply the data in the
         # appropriate encoding. More info:
         # https://gdal.org/development/rfc/rfc23_ogr_unicode.html#oftstring-oftstringlist-fields
+        # NOTE: this always returns False for the layer returned by executing SQL,
+        # even when they indeed support UTF-8; the layer underying the SQL must
```
theroggy, line 490:
```suggestion
        # even when they indeed support UTF-8; the layer underlying the SQL must
```

pyogrio/_io.pyx
```diff
@@ -1156,7 +1189,22 @@ def ogr_read(
...
+        else:
+            if sql is not None and get_driver(ogr_dataset) == "ESRI Shapefile":
+                # in order to properly test certain capabilities, we need to have the
+                # underlying layer referenced by the SQL query
+                base_layer = GDALDatasetGetLayer(ogr_dataset, 0)
+
+            else:
+                base_layer = ogr_layer
+
+            encoding = detect_encoding(ogr_dataset, base_layer)
```
theroggy, line 1207: Maybe this could be moved to a function to avoid duplication of this block in the four functions?

pyogrio/_io.pyx
```diff
@@ -1142,9 +1149,35 @@ def ogr_read(
...
+            if prev_shape_encoding_c != NULL:
+                # strings returned from config options may be replaced via
+                # CPLSetConfigOption() below; GDAL instructs us to save a copy
+                # in a new string
+                prev_shape_encoding = CPLStrdup(prev_shape_encoding_c)
+
+            # set encoding used automatically by GDAL
+            encoding_b = encoding.encode('UTF-8')
+            encoding_c = encoding_b
+            CPLSetThreadLocalConfigOption("SHAPE_ENCODING", <const char*>encoding_c)
```
theroggy, line 1170: Maybe move this to a helper function SetThreadLocalConfigOption that returns the previous set value as well?

pyogrio/_io.pyx
```diff
@@ -1156,7 +1189,22 @@ def ogr_read(
...
         # or from the system locale
-        encoding = encoding or detect_encoding(ogr_dataset, ogr_layer)
+        if encoding:
+            if get_driver(ogr_dataset) == "ESRI Shapefile":
+                # Because SHAPE_ENCODING is set above, GDAL will automatically decode
+                # to UTF-8; ignore any encoding set by user
+                encoding = "UTF-8"
+
+        else:
+            if sql is not None and get_driver(ogr_dataset) == "ESRI Shapefile":
```
theroggy, line 1199: Maybe mention here that this is fixed in GDAL and can be removed once GDAL3.9? becomes the lowest supported version?

brendan-ward: Added note that it may be fixed in 3.9.0 (hasn't been merged yet); I'm not setup to compile GDAL from source locally (don't have the time available to get everything ready for that), so I'm not able to definitely test the fix.

pyogrio/_io.pyx
```diff
@@ -1142,9 +1186,26 @@ def ogr_read(
...
+        if encoding:
+            override_shape_encoding = True
+
+            # for shapefiles, SHAPE_ENCODING must be set before opening the file
+            # to prevent automatic decoding to UTF-8 by GDAL, so we save previous
+            # SHAPE_ENCODING so that it can be restored later
+            # (we do this for all data sources where encoding is set because
+            # we don't know the driver until after it is opened, which is too late)
+            encoding_b = encoding.encode("UTF-8")
+            encoding_c = encoding_b
```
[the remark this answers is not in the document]
brendan-ward: I originally wanted to keep `override_threadlocal_config_option` so that it only worked with C types, in case we ever wanted to pass in C values equivalent to what you'd get back from the function.  Otherwise, I think we'd want to have it encode both the key and value, even if the key passed in is a string defined in code (already a `const char*` I think?) rather than runtime.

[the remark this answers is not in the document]
brendan-ward: > even if the key passed in is a string defined in code (already a const char* I think?)

Hmm, not entirely sure that is correct, [cython docs](https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#string-literals) say those are `str` (unicode) instances but yet show examples like this:  `cdef char* hello_world = 'hello world'` which implies that we at least get the conversion without extra overhead when defining string literals.

pyogrio/_io.pyx
```diff
@@ -1142,9 +1186,26 @@ def ogr_read(
...
+            prev_shape_encoding = override_threadlocal_config_option("SHAPE_ENCODING", encoding_c)
+
         dataset_options = dict_to_options(dataset_kwargs)
         ogr_dataset = ogr_open(path_c, 0, dataset_options)
 
+        if encoding and get_driver(ogr_dataset) == "ESRI Shapefile":
+            # Because SHAPE_ENCODING is set above, GDAL will automatically decode
+            # to UTF-8
+            encoding = "UTF-8"
+
```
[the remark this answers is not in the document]
brendan-ward: Good catch!  Was leftover from an earlier attempt to try and set it before getting the SQL layer, to no avail.

jorisvandenbossche (commented):
Very clear and informative comments and tests!

What I don't fully understand is how this works for non-Shapefiles. From the recent discussions/PRs I had thought that eventually this `encoding` keyword actually only makes sense for shapefiles. But you document and test that you can use it (in theory) also for other formats. But since for other formats there is no such thing as `SHAPE_ENCODING`, GDAL will still assume it is reading those files as UTF-8 and is returning UTF-8 values to us, right? How does that not give errors within GDAL?

Something else I am wondering: do we actually test a format that does not have `OLCStringsAsUTF8` (or where we override GDAL like OSM and xlsx and GeoSJONseq)? i.e. a format where we actually end up using `locale.getpreferredencoding()` 
CSV is an example for that?

jorisvandenbossche:
> @jorisvandenbossche I'm unfamiliar with handling of non-UTF-8 column names / text values in pyarrow - should we even be trying to allow non-UTF-8 content through the Arrow API? I'm not seeing how we would intercept and decode them on our end like we do for the non-Arrow API.

By definition Arrow strings are UTF-8, using anything else would be considered as invalid Arrow data. The only solution one has it to store it as a column with variable size binary data type (and keep track of the actual encoding as a user). But that only works for column values, I don't think there is a workaround for column names.

Given that you already solved this for Shapefiles, and that for other file formats this seems a dubious use case anyway(?), it's probably fine to not support this for reading with Arrow? 
I assume the only option to get this working is to have an upstream change in GDAL, for example by having an `encoding` option in `OGR_L_GetArrowStream` to allow the user to override this and let GDAL decode the read values to UFT-8 before putting that in the Arrow data.

jorisvandenbossche:
> But since for other formats there is no such thing as `SHAPE_ENCODING`, GDAL will still assume it is reading those files as UTF-8 and is returning UTF-8 values to us, right? How does that not give errors within GDAL?

Or GDAL just reads the data and passes them on, assuming it is UTF-8 but not actually doing anything with it? So as long as we then decode those bytes with the user provided encoding instead of UTF-8, this just happens to "work"? (but so eg looking at that file with `ogrinfo` would still given you garbage, because at that point GDAL will assume UTF-8?)

brendan-ward:
> Or GDAL just reads the data and passes them on, assuming it is UTF-8 but not actually doing anything with it? So as long as we then decode those bytes with the user provided encoding instead of UTF-8, this just happens to "work"

This is my assumption, based on the testing here.  GDAL and other readers of those other drivers (GPKG, FGB, etc) will show miscoded field names / values.  So it isn't terribly useful - seems like it enables breaking other tools in the ecosystem - but we don't (yet) specifically prevent it.  I don't know if other tools are able to validly write non-UTF-8 encodings to some of these, but it appears to be possible in practice by converting shapefiles to some of them without correctly accounting for alternative encodings (e.g., [GDAL #7458](https://github.com/OSGeo/gdal/issues/7458)).

It would take a bit of research to determine the list, but it would in theory be possible to define a list of drivers that we allow to be non-UTF-8 on _write_ (e.g., shapefile, maybe CSV, XLSX?), and raise an error on others.

We can leave it a bit more open-ended for _reading_, in part to allow users to specifically force a recoding to cleanup data.  But we have to restrict that to the non-Arrow API.

theroggy:
My current general understanding is as follows. There are 2 general types of format:
1. "rude" formats that don't have a fixed encoding and don't have metadata with them to indicate the encoding.
    - OLCStringsAsUTF8 is False
    - e.g. "CSV"
    - for writing: GDAL doesn't do any conversions, it just writes the string/column data as it is.
    - for reading: GDAL doesn't do any conversions, it just writes the string/column data as it is.
2. formats that support one or more encodings and have metadata about it, or the encoding can be deduced by GDAL.
    - OLCStringsAsUTF8 is True
    - e.g. "ESRI Shapefile", most xml or json formats,...
    - for writing: GDAL expects to receive UTF-8 and will write it in the default encoding of the format. As far as I know/found, "ESRI Shapefile" is the only one that supports to specify the encoding to use for writing (via layer creation option "ENCODING"), and then GDAL will do the conversion to the format specified.
    - for reading: GDAL will by default (try to) convert the data to UTF-8 based on the metadata found in the file. Depending on the file format, there are dataset open options to specify the to be used for reading, and then GDAL will do the conversion to UTF-8. 
        - For "ESRI Shapefile", this options is called ENCODING, and you can specify ENCODING="" to avoid any recoding. Probably it is a pity this name is the same as the encoding parameter in pyogrio. If this wasn't the case, we could have just said: please specify the `SHP_ENCODING` open option.
        - For "DXF", this option is called DXF_ENCODING, and you can specify DXF_ENCODING="UTF-8" to avoid any recoding.
        - For most XML-based files (e.g. "GPX", "KML",...), the encoding should be correct in the xml file, you cannot overrule it software.
        - For json-based formats, I suppose GDAL autodetects the encoding based on the content.

jorisvandenbossche:
> * For "ESRI Shapefile", this options is called ENCODING, and you can specify ENCODING="" to avoid any recoding. Probably it is a pity this name is the same as the encoding parameter in pyogrio. If this wasn't the case, we could have just said: please specify the `SHP_ENCODING` open option.

BTW if there is a clash with one of our own keywords, you can always explicitly pass the GDAL layer/dataset open option through `layer_options=dict(encoding="..")`

theroggy:
> BTW if there is a clash with one of our own keywords, you can always explicitly pass the GDAL layer/dataset open option through `layer_options=dict(encoding="..")`

Indeed, but only for `write_dataframe` at the moment, not (yet) for `read_dataframe`.

jorisvandenbossche:
> Indeed, but only for `write_dataframe` at the moment, not (yet) for `read_dataframe`.

Ah, that's something we should probably add then.

--- 

For this PR, shall we merge it? (it just needs some fixing merge conflicts)

theroggy:
Am I right that if a user passes `encoding=` for a file format having `OLCStringsAsUTF8 == True` that is not "ESRI Shapefile", it will be ignored?
If so, maybe it is useful to show a warning in that case?

CHANGES.md
```diff
@@ -19,6 +18,9 @@
 -   Raise exception in `read_arrow` or `read_dataframe(..., use_arrow=True)` if
     a boolean column is detected due to error in GDAL reading boolean values (#335)
     this has been fixed in GDAL >= 3.8.3.
+-   Properly handle decoding of ESRI Shapefiles with user-provided `encoding`
+    option for `read`, `read_dataframe`, and `open_arrow`, and correctly encode
+    Shapefile field names and text values to the user-provided `encoding` (#384).
```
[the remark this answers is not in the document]
brendan-ward: Yes, that was for writing.  Will make more explicit.

brendan-ward:
I've tried to address some of the discussion in the latest changes:

For the Arrow API (open / read arrow) it will raise an exception if an alternative encoding is provided and driver is not shapefile

We now specifically raise exception for shapefiles if both `encoding` parameter and `"ENCODING"="<whatever>"` open / layer creation options are combined, because this leads to order of operations issues because we are specifically setting `SHAPE_ENCODING` or `ENCODING` on the user's behalf.  Since `SHAPE_ENCODING` may be set in advance as a configuration option by the user (possibly re-used across many data sources), we don't raise an exception, we just silently override it (so there isn't a case where they can set it to `SHAPE_ENCODING=""` to disable recoding by GDAL and then pass in an alternative `encoding` parameter).  This seems like advanced territory, so I didn't think it necessary to raise a warning until we start seeing issues.

This does not disable writing to an alternative encoding if provided by the user (hopefully they know the right thing to do for the target driver); I think we can leave this as is until we see issues.

brendan-ward:
Added a specific test to verify that `locale.getpreferredencoding()` is used for CSV files (only runs on Windows though) when not providing `encoding` parameter.

brendan-ward:
Added more tests for the Arrow write API, slimmed down the comments and repeated code around encoding in `_io.pyx`, and fixed an issue that would have bitten us immediately with the forthcoming GDAL 3.9: FlatGeobuff no longer allows writing non-UTF-8 values.  We're already likely in unsupported territory with the tests that prove we can write non-UTF-8 values to drivers that likely expect UTF-8, but I left the other drivers in place for now because it proves the encoding works.  Alternatively, we could simply block writing non-UTF-8 values for non-shapefile drivers where `OLCStringsAsUTF8` is `True`.

jorisvandenbossche (approved):

jorisvandenbossche:
Didn't notice it before (and haven't yet look in detail what might be causing it), but merging this caused a bunch failures when testing the wheels for ubuntu and older macos: https://github.com/geopandas/pyogrio/actions/runs/8848749467/job/24299655017

PR #391, opened 2024-04-17
FIX: correctly ignore fields when reading data

### LINKED ISSUE #388: BUG: reading with arrow returns empty GeoDataFrame when column referenced in where parameter is not included in results
Observed with GDAL 3.8.3 on MacOS

```Python
from pyogrio import read_dataframe

filename = "pyogrio/tests/fixtures/naturalearth_lowres/naturalearth_lowres.shp"
df = read_dataframe(
    filename, where=""" "iso_a3" = 'CAN' """, use_arrow=True, columns=[]
)
```

yields 

```Python
Empty GeoDataFrame
Columns: [geometry]
Index: []
```

when it should have one record.

Unclear if this is an error on our side our in GDAL.

brendan-ward:
Resolves #388 

Drivers will return inconsistent results when `columns` does not include the columns referenced in `where` expression.  For some drivers, this will either return 0 results or raise an exception; other drivers appear to work without issue (GPKG, GeoJSON).  Added documentation to indicate that users should include the column used by `where` in their list of columns.

The test added here verifies that fields are indeed ignored by GDAL (previously they weren't properly ignored).


jorisvandenbossche (approved):
Looks good, thanks!

theroggy (approved):

PR #396, opened 2024-04-26
Refactor cleanup of GDAL objects / close of dataset on write

### LINKED ISSUE #394: Need to call GDALClose when writing fails?
Currently in the writing code (eg `ogr_write`) we have a final cleanup block where we close the dataset (and check for errors, which was corrected in https://github.com/geopandas/pyogrio/pull/284):

https://github.com/geopandas/pyogrio/blob/154466ab555f395d24b3d4873e685a841dd81fa6/pyogrio/_io.pyx#L2277-L2295

But when an error happens while writing (and this is already bubbled up), we don't close the dataset, see the `finally:` block in the snippet above (we only clean up the current Feature variable).

When I compare that with the various read paths, it seems we always close the dataset in the finally block, eg in `ogr_read`:

https://github.com/geopandas/pyogrio/blob/154466ab555f395d24b3d4873e685a841dd81fa6/pyogrio/_io.pyx#L1245-L1255

brendan-ward:
Resolves #394 

This reorganizes the write functions a bit to make sure that we're always cleaning up allocated objects (feature, geometry, etc) in the `finally` block in case of exception, which reduces the amount of inner `try` / `catch` blocks we need to have.  This should then correctly release those within the loops under the non-exception path.

This also strips out the `OGRReleaseDataSource` we were using in various places for `GDALClose`, which is the current recommended function.

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2064,11 +2064,13 @@ cdef create_ogr_dataset_layer(
                 ogr_crs = create_crs(crs)
 
             except Exception as exc:
-                OGRReleaseDataSource(ogr_dataset)
-                ogr_dataset = NULL
                 if dataset_options != NULL:
                     CSLDestroy(dataset_options)
                     dataset_options = NULL
+
+                GDALClose(ogr_dataset)
+                ogr_dataset = NULL
+
                 raise exc
 
         # Setup other layer creation options
@@ -2114,7 +2116,7 @@ cdef create_ogr_dataset_layer(
         set_metadata(ogr_layer, layer_metadata)
 
     except Exception as exc:
-        OGRReleaseDataSource(ogr_dataset)
+        GDALClose(ogr_dataset)
         ogr_dataset = NULL
         raise DataLayerError(str(exc))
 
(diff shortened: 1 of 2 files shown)


pyogrio/_io.pyx
```diff
@@ -2196,79 +2198,68 @@ def ogr_write(
...
-            except:
-                if ogr_fielddef != NULL:
-                    OGR_Fld_Destroy(ogr_fielddef)
-                    ogr_fielddef = NULL
+                    if width:
+                        OGR_Fld_SetWidth(ogr_fielddef, width)
 
-                OGRReleaseDataSource(ogr_dataset)
-                ogr_dataset = NULL
-                raise FieldError(f"Error creating field '{fields[i]}' from field_data") from None
```
brendan-ward, line 2248: We weren't specifically testing for this exception, and it doesn't seem particularly helpful compared to the following one raised when actually adding the fields, so I removed it.

rouault:
I'm perhaps hijacking this PR a bit but just to mention that starting with GDAL 3.7, GDALClose() returns a CPLErr return code (previously was void). This can be a way of checking that everything has been successfullly written (some formats completely defer writing at close time). Cf https://gdal.org/development/rfc/rfc91_dataset_close.html

pyogrio/_io.pyx
```diff
@@ -208,6 +208,25 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
 
 
+cdef ogr_close(GDALDatasetH ogr_dataset):
+    """Close the dataset and raise exception if that fails.
+    NOTE: some drivers only raise errors on write when calling GDALClose()
+    """
+    if ogr_dataset != NULL:
+        IF CTE_GDAL_VERSION >= (3, 7, 0):
+            if GDALClose(ogr_dataset) != CE_None:
+                return exc_check()
```
brendan-ward, line 218: We still need this to capture the error message emitted by GDAL, we just know to specifically look for it because `GDALClose` raised an error.

jorisvandenbossche (approved):
Looks good!

Issue #249, opened 2023-05-01
ENH: support writing to in-memory (byte) objects


jorisvandenbossche:
We support reading from an in-memory buffer / bytes object (https://github.com/geopandas/pyogrio/issues/22), but not yet writing to it. 

As a starter, the write path assumes the path is a string (or we will convert it to a string in several places), and so passing a BytesIO object doesn't work (currently we will actually create a file in the current directory with a name like "<_io.BytesIO object at 0x7f229d2a1a80>" because of calling `str(path)`). 

If we want to support writing to a buffer, our current code for handling this on the read path (`buffer_to_virtual_file` to create a /vsimem/.. file) will not be sufficient, because this creates a `VSIMemFile` that doesn't own the buffer's data, and thus can't expand that size of the buffer (which will be needed to write to an empty BytesIO).

From a quick look, two potential strategies:

- When writing, create a custom VSIMemFile that owns its memory, and at the end after writing the file, copy out the buffer from the gdal in-memory file, and write that data into the Python in-memory file. 
  This is probably simpler to implement, but incurs an extra copy of the data when you want the result to end up in a Python object (like BytesIO) that was provided by the user. Alternatively, we could also have a way to just return the buffer as bytes, which shouldn't require an extra copy (if GDAL allows to transfer the ownership). 
  I _think_ this is also what fiona currently does: https://github.com/Toblerity/Fiona/blob/54428a1d39d1115d3e5e7158e09f37ac3623de23/fiona/__init__.py#L277-L283
- Wrap the Python file-like object (like BytesIO for in-memory) in a custom GDAL filesystem and file object plugin (through callbacks that link the Python object `read`/`seek`/`write` methods to the methods in C++), which is covered by https://github.com/geopandas/pyogrio/issues/42 for the read side (and which was implemented in rasterio)

brendan-ward:
The second option looks reasonable based on the way this was implemented rasterio, assuming that adding write functionality on top of the read functionality implemented there is relatively straightfroward.  We could start with re-implementing the read interface in pyogrio first (#42) and adjust to that, then extend it to enable write.

jorisvandenbossche:
@brendan-ward in case this is useful, what I started writing (very draft): https://github.com/geopandas/pyogrio/compare/main...jorisvandenbossche:pyogrio:write-vsimem?expand=1 
[...]

PR #397, opened 2024-04-30
ENH: enable support for writing to memory

brendan-ward:
Resolves https://github.com/geopandas/pyogrio/issues/249

This enables basic writing to memory using an `io.BytesIO` object.  This does not currently support append / adding layers; I tried working that out unsuccessfully in earlier commits, but we could return to this in a later PR.  It seems like it should be possible in theory.

This specifically blacklists ESRI Shapefile and OpenFileGDB because they write multiple files, which can't be returned successfully as a single memory file.

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -1991,6 +1991,53 @@ cdef infer_field_types(list dtypes):
     return field_types
 
 
+cdef str get_ogr_vsimem_path(object path_or_fp, str driver, VSILFILE **vsi_handle):
+    """ Return the original path or a /vsimem/ path, and update vsi_handle if needed
+
+    Parameters
+    ----------
+    path_or_fp : str or io.BytesIO object
+        If io.BytesIO object, this will create a new in-memory file at the returned
+        path and set vsi_handle to it
+    driver : str
+        name out output driver
+    vsi_handle : **VSIFILE
+        pointer to VSI file handle that will be updated to point to in-memory file
+    """
+
+    cdef char *path_c = NULL
+    cdef bytes bytes_buffer
+    cdef unsigned char *bytes_view = NULL
+    cdef unsigned char *vsi_buffer = NULL
+    cdef int num_bytes = 0
+
+    if not isinstance(path_or_fp, BytesIO):
+        return path_or_fp
+
+    # Create in-memory directory to contain auxiliary files
+    memfilename = uuid4().hex
+    VSIMkdir(f"/vsimem/{memfilename}".encode("utf-8"), 0666)
+
+    # file extension is required for some drivers, set it based on driver metadata
+    ext = ''
+    recommended_ext = _get_driver_metadata_item(driver, "DMD_EXTENSIONS")
+    if recommended_ext is not None:
+        ext = "." + recommended_ext.split(' ')[0]
+
+    path = f"/vsimem/{memfilename}/{memfilename}{ext}"
+    path_b = path.encode('UTF-8')
+    path_c = path_b
+
+    num_bytes = path_or_fp.getbuffer().nbytes
+    if num_bytes > 0:
+        raise NotImplementedError("writing to existing in-memory object is not supported")
+
+    # Open new VSI file in write-only mode
+    vsi_handle[0] = VSIFOpenL(path_c, "w")
+
+    return path
+
+
 cdef create_ogr_dataset_layer(
     str path,
     bint is_vsi,
@@ -2234,8 +2281,8 @@ def ogr_write(
     cdef int num_records = -1
     cdef int num_field_data = len(field_data) if field_data is not None else 0
     cdef int num_fields = len(fields) if fields is not None else 0
+    cdef bint is_vsi = False
     cdef VSILFILE *vsi_handle = NULL
-    cdef bytes vsi_bytes_buffer
     cdef unsigned char *vsi_buffer = NULL
     cdef vsi_l_offset vsi_buffer_size = 0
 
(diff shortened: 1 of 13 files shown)


jorisvandenbossche:
> Some things we still need to add / figure out:
> 
>     * append / add layers (still trying to figure out how to make this work)

FWIW, I would personally first focus on getting the plain writing to memory work without append (just raise for now if the mode is append)  
(fiona also raises an error if `mode="a"` for writing to file-like objects)

brendan-ward:
Fiona has [specific tests for appending](https://github.com/Toblerity/Fiona/blob/main/tests/test_memoryfile.py#L273-L289) so I assumed it was possible and expected behavior, but indeed in the interest of getting this out sooner we can more strictly limit this to write-only mode.

jorisvandenbossche:
> Fiona has [specific tests for appending](https://github.com/Toblerity/Fiona/blob/main/tests/test_memoryfile.py#L273-L289) so I assumed it was possible and expected behavior

I _assume_ this only works when directly using the `MemoryFile` interface. Because at that point you have the same object, where you first write to, and then append to. 
But when creating a _new_ memfile, you can't append to it ([test](https://github.com/Toblerity/Fiona/blob/a99c7370b07f0c490dfb50c6ef290e1c69178e2f/tests/test_memoryfile.py#L302-L308)). And that is essentially what happens in geopandas when people pass a BytesIO object to `to_file`, we each time create a new MemoryFile object through `fiona.open()`, and thus only pure writing is supported (fiona [explicitly raise in case of append mode](https://github.com/Toblerity/Fiona/blob/a99c7370b07f0c490dfb50c6ef290e1c69178e2f/fiona/__init__.py#L228-L256) in `open()` for a writeable object). 

So longer term that is also something we _could_ do, but it might requiring creating some similar custom object like fiona's MemoryFile?

pyogrio/_io.pyx
```diff
@@ -2523,6 +2673,29 @@ def ogr_write_arrow(
...
+            VSIFSeekL(vsi_handle, 0, 0)
+
+            # Take ownership of the buffer to avoid a copy; GDAL will automatically
+            # unlink the memory file
+            vsi_buffer = VSIGetMemFileBuffer(path.encode("UTF-8"), &vsi_buffer_size, 1)
+            if vsi_buffer == NULL:
+                raise RuntimeError("could not read bytes from in-memory file")
+
+            # reset path_or_fp and write bytes to it
+            path_or_fp.seek(0)
```
jorisvandenbossche, line 2694: This is the user-provided BytesIO object at this point, right? 
In that case, should we seek into it? I would expect we just write to it as the user provided it (they might want to write to a specific place)

brendan-ward: Yes, it is the user-provided BytesIO object.  However, we don't allow it to have existing bytes, so this first seek is unnecessary.  I don't think we want to support allowing writes to arbitrary locations within a passed-in BytesIO object; if we ever allowed it to be non-empty (e.g., if we could get append working) then we'd want to completely overwrite it.

pyogrio/_io.pyx
```diff
@@ -2023,25 +2094,38 @@ cdef create_ogr_dataset_layer(
...
+
 
     path_b = path.encode('UTF-8')
     path_c = path_b
 
     driver_b = driver.encode('UTF-8')
     driver_c = driver_b
 
+    if is_vsi:
+        path_exists = VSIStatExL(path_c, &st_buf, VSI_STAT_EXISTS_FLAG) == 0
```
jorisvandenbossche, line 2107: Do you if this is actually needed for the current scope of this PR? Because right now you just created this memfile with a unique uuid, so it will never already exist?

pyogrio/_io.pyx
```diff
@@ -2418,6 +2528,29 @@ def ogr_write(
...
+        if exc:
+            raise DataSourceError(f"Failed to write features to dataset {path}; {exc}")
+
+        ### Copy in-memory file back to path_or_fp object
+        if vsi_handle != NULL:
+            # rewind to beginning of file and read bytes back from memory
+            VSIFSeekL(vsi_handle, 0, 0)
+
+            # Take ownership of the buffer to avoid a copy; GDAL will automatically
+            # unlink the memory file
```
jorisvandenbossche, line 2543: Should we then set `vsi_handle` to null to avoid we unlink it afterwards again? (although that doesn't seem to be giving errors right now)

pyogrio/_io.pyx
brendan-ward: Correct; no longer needed, it was left over from trying to append (can add back in later as part of that)

pyogrio/raw.py
```diff
@@ -527,6 +524,67 @@ def _preprocess_options_kwargs(driver, dataset_options, layer_options, kwargs):
...
+            raise DataSourceError(
+                f"{driver} does not support ability to write in-memory in GDAL "
+                f"{get_gdal_version_string()}"
+            )
+
+        if append:
+            raise NotImplementedError("append is not supported for in-memory files")
+
+    else:
+        path = str(path)
```
jorisvandenbossche, line 567: Should this still call `vsi_path(..)`? (what it did before in the deleted code below)

brendan-ward: I took it out because I wasn't convinced that it would work properly and my instead emit confusing errors (though stringifying something that would have gotten a VSI path would be confusing too).  We don't have any tests that prove writing to `/vsi*/` paths would be successful.  I also didn't see support for writing to a `/vsi*/` path in Fiona; they are largely used for reading.

For instance, it looks like it is not possible to write via the `/vsizip/` handler:

```Python
write_dataframe(df, "/vsizip//tmp/test.shp.zip", driver="ESRI Shapefile")

...
pyogrio.errors.DataSourceError: Driver ESRI Shapefile does not support write functionality
```

Are you aware of any things that we can write to a `/vsi*/` path that should succeed?  I can also restore the original behavior, in case our current unknown / untested functionality is working for someone.

jorisvandenbossche: In theory `/vsizip` should have write capabilities (https://gdal.org/user/virtual_file_systems.html#write-capabilities), but indeed I also don't get that working (maybe the way we handle that path internally isn't fit for working with such a vsi path)

But something like writing to S3 with `vsis3` should in theory also be supported? (although that is indeed something we don't test)

pyogrio/tests/test_arrow.py
```diff
@@ -582,7 +588,7 @@ def test_write_append_unsupported(tmpdir, naturalearth_lowres, driver, ext):
 def test_write_gdalclose_error(naturalearth_lowres):
     meta, table = read_arrow(naturalearth_lowres)
 
-    filename = "s3://non-existing-bucket/test.geojson"
+    filename = "/vsis3/non-existing-bucket/test.geojson"
```
jorisvandenbossche, line 591: What was the reason for this change? (we should at least also ensure writing to s3:// works as well, so maybe parametrize the test for both options?)

pyogrio/tests/test_core.py
```diff
@@ -460,15 +460,19 @@ def test_read_info_force_feature_count(data_dir, layer, force, expected):
...
-    "force_total_bounds, expected_total_bounds",
-    [(True, (-180.0, -90.0, 180.0, 83.64513)), (False, None)],
-)
-def test_read_info_force_total_bounds(
-    tmpdir, naturalearth_lowres, force_total_bounds, expected_total_bounds
-):
-    # Geojson files don't hava a fast way to determine total_bounds
+@pytest.mark.parametrize("force_total_bounds", [True, False])
+def test_read_info_force_total_bounds(tmpdir, naturalearth_lowres, force_total_bounds):
+    # Geojson files don't have a fast way to determine total_bounds for GDAL <3.9.0
```
jorisvandenbossche, line 465: Ideally for later we should find a new file format to test with that doesn't support fast total bounds?

pyogrio/tests/test_raw_io.py
```diff
@@ -722,7 +723,7 @@ def test_write_unsupported(tmpdir, naturalearth_lowres):
 def test_write_gdalclose_error(naturalearth_lowres):
     meta, _, geometry, field_data = read(naturalearth_lowres)
 
-    filename = "s3://non-existing-bucket/test.geojson"
+    filename = "/vsis3/non-existing-bucket/test.geojson"
```
jorisvandenbossche, line 726: Same comment here

pyogrio/raw.py
jorisvandenbossche: I quickly tested this with a public test bucket, and I can write to an "s3://" path (which will get converted to /vsis3 by our preprocessing)

jorisvandenbossche (approved):
Looking good!

pyogrio/raw.py
brendan-ward: Thanks for testing!  That confirms we should indeed leave it in place for now.

pyogrio/tests/test_arrow.py
brendan-ward: These were producing failures when I had disabled handling the vsi paths; since that has been reverted I'm also reverting these.

pyogrio/tests/test_core.py
brendan-ward: It looks like GeoJSONSeq still doesn't support fast total bounds, so I reverted back to using that directly.

PR #399, opened 2024-04-30
RLS/BLD: Ensure VCPKG brings in iconv library

### LINKED ISSUE #398: BUG: encoding not handled properly for wheels
See failing tests: https://github.com/geopandas/pyogrio/actions/runs/8848749467/job/24299655017

Fails on Ubuntu / MacOS when creating non-UTF-8 encoded shapefiles.  Fewer failures on Python 3.12 than previous versions.

First guess is that there is variation in the underlying libraries GDAL uses to convert from UTF-8 to an alternative encoding (libiconv?) under VCPKG than in our other CI environments.

brendan-ward:
Resolves #398

### DIFF
--- a/ci/vcpkg.json
+++ b/ci/vcpkg.json
@@ -1,11 +1,11 @@
 {
     "name": "pyogrio",
-    "version": "0.4.0",
+    "version": "0.8.0",
     "dependencies": [
         {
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Perfect!

PR #407, opened 2024-05-03
ENH: refactor handling of reading from in-memory dataset

### LINKED ISSUE #401: BUG: read_arrow does cannot read from bytes / buffer / file-like
When reading from a file-like, we first extract out the bytes and set them into a VSI memory file, and then clean that up after reading the datasource.  When reading from Arrow using read_arrow (ultimately affects open_arrow), we inadvertently cleanup the memory file before actually reading from it because we do the following in `open_arrow`:

```Python
path, buffer = get_vsi_path(path_or_buffer)

try:
    return ogr_open_arrow(...)

finally:
    if buffer:
        remove_virtual_file(path)
```

The `finally` block gets called before `ogr_open_arrow` has read from the memory file.

The cleanest fix is likely to move all handling of the memory file to `ogr_open_arrow`, and let it cleanup the memory file in its finally block.

brendan-ward:
Resolves #401 

This moves all handling of creation / destruction of the in-memory dataset used for read operations to the Cython tier, and now passes in either a string or bytes to `read_info()`, `list_layers()`, `read_bounds()`, `read()`, `read_arrow()` functions.  I think this also sets us to for later refactors where we may pass file-like / filesystem-like objects directly down to Cython in order to use the GDAL virtual filesystem plugin.

To keep things organized, I split VSI related Cython functions into the new `_vsi.pyx` / `_vsi.pxd` files.

Because the bytes buffer is passed as a parameter, it remains in scope during the Cython function and we don't need to hold an extra handle on it like before. (the handle was required to prevent Python from deallocating it while GDAL still may use it to represent the in-memory dataset).

This expands the tests to verify that core functions still work correctly when passed bytes or file-like objects.  To avoid the various GPKG related errors when working with this type in memory, I instead added new test fixtures to create a GeoJSON file from the first 3 records of the naturalearth_lowres dataset.

This also adds a specific check that the incoming path does not already contain `/vsimem/`, since we have to handle that internally and cannot allow it to be passed in.

### DIFF
--- a/pyogrio/tests/test_arrow.py
+++ b/pyogrio/tests/test_arrow.py
@@ -161,33 +161,18 @@ def test_read_arrow_vsi(naturalearth_lowres_vsi):
     assert len(table) == 177
 
 
-@requires_arrow_write_api
-@pytest.mark.filterwarnings("ignore:A geometry of type POLYGON is inserted")
-@pytest.mark.filterwarnings("ignore:File /vsimem:RuntimeWarning")
-@pytest.mark.parametrize("driver,ext", [("GeoJSON", "geojson"), ("GPKG", "gpkg")])
-def test_read_from_bytes(tmp_path, naturalearth_lowres, driver, ext):
-    meta, table = read_arrow(naturalearth_lowres)
-    meta["geometry_type"] = "MultiPolygon"
+def test_read_arrow_bytes(geojson_bytes):
+    meta, table = read_arrow(geojson_bytes)
 
-    filename = tmp_path / f"test.{ext}"
+    assert meta["fields"].shape == (5,)
+    assert len(table) == 3
 
-    write_arrow(
-        table,
-        filename,
-        driver=driver,
-        layer="test",
-        crs=meta["crs"],
-        geometry_type=meta["geometry_type"],
-        geometry_name=meta["geometry_name"] or "wkb_geometry",
-    )
 
-    with open(filename, "rb") as f:
-        buffer = f.read()
+def test_read_arrow_filelike(geojson_filelike):
+    meta, table = read_arrow(geojson_filelike)
 
-    actual_meta, actual_table = read_arrow(buffer)
-
-    assert np.array_equal(actual_meta["fields"], meta["fields"])
-    assert len(actual_table) == len(table)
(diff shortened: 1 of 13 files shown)


pyogrio/tests/test_arrow.py
```diff
@@ -161,6 +161,35 @@ def test_read_arrow_vsi(naturalearth_lowres_vsi):
...
+        filename,
+        driver=driver,
+        layer="test",
+        crs=meta["crs"],
+        geometry_type=meta["geometry_type"],
+        geometry_name=meta["geometry_name"] or "wkb_geometry",
+    )
+
+    with open(filename, "rb") as f:
+        buffer = f.read()
```
jorisvandenbossche, line 185: Is there a need to write this first, instead of directly reading the buffer from `naturalearth_lowres` ?

jorisvandenbossche: (and then this read test also doesn't require `requires_arrow_write_api`)

pyogrio/tests/test_arrow.py
```diff
@@ -817,7 +847,7 @@ def test_write_memory(naturalearth_lowres, driver):
 
     buffer = BytesIO()
     write_arrow(
-        table.slice(0, 1),
+        table,
```
jorisvandenbossche, line 850: The reason I did this was to avoid the filterwarning you added .. ;) Ideally we would add a test dataset without mixed geometries that we can use for most tests .. (or just edit the naturalearth fixture to give all MultiPolygons), but that's for another issue

pyogrio/tests/test_arrow.py
jorisvandenbossche: Or, could also use the new fixtures `geojson_bytes` and `geojson_filelike` here?

jorisvandenbossche (commented):
This looks great! Thanks a lot

pyogrio/tests/test_arrow.py
brendan-ward: I switched the places where we were doing `.slice(0, 1)` to only read 1 feature from naturalearth, and sidestep the issue (and also still allow comparison of first table to the written & reread table).

jorisvandenbossche (approved):

PR #411, opened 2024-05-04
TST/CLN: Replace all tmpdir / os.path operations in tests with pathlib.Path

### LINKED ISSUE #400: TST: replace pytest.tmpdir fixture with pytest.tmp_path
`tmpdir` is a [legacy interface](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures), and we're actually detecting that they have a `read` method in some cases and reading their contents to bytes in `get_vsi_path`, which is not what we should be doing.

Among other things, using `tmp_path` instead allows us to pass these in as `pathlib.Path` objects, which we can specifically check for and convert to strings before subsequent processing.

brendan-ward:
Resolves #400 

This removes all instances of the legacy pytest `tmpdir` fixture, and uses the `tmp_path` (a pathlib.Path object) fixture instead.

Modernized the various path operations in the tests to use methods on the pathlib.Path object.

Removes unused `win32.py` test file; we have actual tests against windows and this is no longer used.

Note: I updated the benchmarks to use `tmp_path` and verified that most of them still run (some fail with missing test files), but we're not actively maintaining these.


pyogrio/tests/test_geopandas_io.py
```diff
@@ -1799,8 +1800,8 @@ def test_read_dataframe_arrow_dtypes(tmp_path):
     __gdal_version__ < (3, 8, 3), reason="Arrow bool value bug fixed in GDAL >= 3.8.3"
 )
 @pytest.mark.parametrize("ext", ALL_EXTS)
-def test_arrow_bool_roundtrip(tmpdir, ext):
-    filename = os.path.join(str(tmpdir), f"test{ext}")
+def test_arrow_bool_roundtrip(tmp_path, ext):
+    filename = tmp_path / f"test.{ext}"
```
theroggy, line 1804:
```suggestion
    filename = tmp_path / f"test{ext}"
```

theroggy (approved):
Nice cleanup!

jorisvandenbossche (approved):
Looks good!

There are a few cases where we still pass `str(filename)` to some read function and where the `str(..)` can be removed now (after your previous PR to uniformize the path handling)

PR #412, opened 2024-05-08
BUG: Fix hash url parsing

m-richards:
This is an upstream equivalent of https://github.com/geopandas/geopandas/pull/3280 to fix https://github.com/geopandas/geopandas/issues/3279 which is an edge case to do with using urlparse in detecting file schemes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## 0.8.1 (yyyy-mm-dd)
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


brendan-ward (commented):
Thanks @m-richards - this was definitely not something we were watching for!

Can you please add a changelog entry for this fix, for 0.8.1?

theroggy (approved):

brendan-ward (approved):
Thanks @m-richards !

PR #415, opened 2024-05-13
ENH: enable more ruff rules + make appropriate changes

theroggy:
Same rules as applied in geopandas + pyupgrade (python 3.8) + pydocstyle


pyogrio/raw.py
```diff
@@ -425,7 +422,8 @@ def open_arrow(
             "geometry_type": "<geometry_type>",
             "geometry_name": "<name of geometry column in arrow table>",
         }
-    """
+
+    """  # noqa: E501
```
brendan-ward, line 426: Or just add line breaks to docstring where needed?

pyogrio/util.py
```diff
@@ -159,15 +154,15 @@ def _construct_vsi_path(path, archive, scheme) -> str:
...
     return path
 
 
 def _preprocess_options_key_value(options):
-    """
-    Preprocess options, eg `spatial_index=True` gets converted
-    to `SPATIAL_INDEX="YES"`.
+    """Preprocess options.
+
+    Eg. `spatial_index=True` gets converted to `SPATIAL_INDEX="YES"`.
```
brendan-ward, line 165:
```suggestion
    For example, `spatial_index=True` gets converted to `SPATIAL_INDEX="YES"`.
```

pyproject.toml
```diff
@@ -62,4 +62,132 @@ tag_prefix = "v"
...
+    # pairwise-over-zipped (>=PY310 only)
+    "RUF007",
+    # explicit-f-string-type-conversion
+    "RUF010",
+]
+
 exclude = ["docs/*", "benchmarks/*", "pyogrio/_version.py"]
+
+[tool.ruff.per-file-ignores]
+"**/tests/*" = ["D"]
```
brendan-ward, line 193:
```suggestion
# ignore pydocstyle errors in tests
"**/tests/*" = ["D"]
```

brendan-ward (commented):
Thanks for the cleanup @theroggy , just a few minor comments.

pyogrio/raw.py
theroggy: The lines that are too long are in the code samples, not sure how to deal with those?

brendan-ward: Could just introduce a variable which would make those shorter:

```Python
    >>> with open_arrow(path, use_pyarrow=True) as source:
    >>>     meta, reader = source
    >>>     geom_col = meta["geometry_name"] or "wkb_geometry"
    >>>     for batch in reader:
    >>>         geometries = shapely.from_wkb(batch[geom_col])
```

pyogrio/geopandas.py
```diff
@@ -34,11 +33,11 @@ def _try_parse_datetime(ser):
     import pandas as pd  # only called when pandas is known to be installed
 
     if PANDAS_GE_22:
-        datetime_kwargs = dict(format="ISO8601")
+        datetime_kwargs = {"format": "ISO8601"}
```
jorisvandenbossche, line 36: FWIW, I am not going to disagree merging this PR for this ;), but this is a change I dislike. When specifically creating keyword arguments, I like that `dict(..)` is closer in syntax to directly passing those to the function

jorisvandenbossche (approved):

pyogrio/geopandas.py
theroggy: Yes, I remember the same remark being made in geopandas :-):
https://github.com/geopandas/geopandas/pull/2921#discussion_r1226338303

jorisvandenbossche: Ah, it seems I somewhat consistent in my preference ;) 
It's also just adding one additional rule to the ignore list if we wouldn't want this change (C408, https://docs.astral.sh/ruff/rules/unnecessary-collection-call/)

pyproject.toml
```diff
@@ -62,4 +62,133 @@ tag_prefix = "v"
...
+    # line break before binary operator
+    # "W503",  # not yet implemented
+    # line break after binary operator
+    # "W504",  # not yet implemented
+    # controversial
+    "B006",
+    # controversial
+    "B007",
+    # controversial
+    "B008",
```
jorisvandenbossche, line 123: We might want to keep this consistent with geopandas, but I _think_ those could probably all be removed (they don't seem that controversial, and I also doubt we have occurrences of them)

pyogrio/geopandas.py
brendan-ward: I'll defer to @jorisvandenbossche ; I don't personally have a strong opinion on this.  Given that it looks like this was accepted on the GeoPandas side more for tests, and this isn't strictly for tests here, it seems like ignoring `C408` might be the better option here.

theroggy: I don't have an opinion on the syntax being used. Just a mild opinion on the principle of using an oppiniated linter to avoid this kind of dicussions and then still have them :-), but I avoid having principles in general, so no problem at all to ignore `C408`.

I don't know what the basis of the remark in the geopandas comment is that it would only be for tests in geopandas. That is not the case: ruff runs on the entire geopandas codebase, so the PR also changed this in the rest of the code base.

pyproject.toml
theroggy: I agree that they sound like useful checks, and some af them can even avoid weird situations that are hard to debug.

Maybe activate them in geopandas as well, then it is consistent between geopandas and pyogrio?

pyogrio/geopandas.py
brendan-ward: I was referring to this comment via the above GeoPandas link from @jorisvandenbossche re: tests:

> (now, I am fine here for our tests to just follow ruff, I just wouldn't like that for my personal scripts ;))

theroggy: Yes, I know, but the PR in geopandas isn't only targetting the tests.... Obviously it didn't change @jorisvandenbossche s personal scripts though ;-)

pyproject.toml
jorisvandenbossche: @theroggy to follow-up on this, I did clean up those settings on the geopandas side (https://github.com/geopandas/geopandas/pull/3291, and also adding isort support in https://github.com/geopandas/geopandas/pull/3297). So maybe we could just update here to simply follow the settings from geopandas now?

theroggy: @jorisvandenbossche Yes, I saw... I was busy looking that indeed (syncing settings with geopandas again), but real life got in the way ;-)... I'll try to finish it...

I added some additional things here as well, so it would be best these additions would be synced back to geopandas to get the settings completely in sync.

theroggy:
I synced most settings with geopandas, including the custom isort config.

Also updated ruff version to 0.5.2.

pyogrio/core.py
```diff
@@ -1,27 +1,38 @@
...
     from pyogrio._ogr import (
+        _get_drivers_for_path,
+        _register_drivers,
+        get_gdal_geos_version,
         get_gdal_version,
         get_gdal_version_string,
-        get_gdal_geos_version,
         ogr_list_drivers,
-        set_gdal_config_options as _set_gdal_config_options,
+    )
```
brendan-ward, line 20: any idea why these got split up into multiple blocks?  Seems like all imports from `pyogrio._ogr` should be in a single block.

pyogrio/raw.py
```diff
@@ -472,29 +474,27 @@ def _parse_options_names(xml):
...
     if metadata is not None:
         if layer_metadata is not None:
             raise ValueError("Cannot pass both metadata and layer_metadata")
         layer_metadata = metadata
 
     # validate metadata types
-    for metadata in [dataset_metadata, layer_metadata]:
-        if metadata is not None:
-            for k, v in metadata.items():
+    for meta in [dataset_metadata, layer_metadata]:
```
brendan-ward, line 484: Any particular reason for the change of `metadata` to `meta`?  Either is fine, but `metadata` was slightly more clear.

brendan-ward (commented):
Thanks for the update @theroggy , this PR is good cleanup.  Just a couple minor suggestions from me.

pyogrio/raw.py
theroggy: There is already an input parameter called `metadata` in the function definition. The local variable of the for loop redefines this variable, which is something ruff doesn't like because it can be confusing...

pyogrio/core.py
theroggy: The default behaviour is apparently to keep import with an alias on a seperate line rather than combining them. This can be changed with an extra configuration setting: `combine_as_imports = true`.

I'll add it in the config...

brendan-ward (approved):

jorisvandenbossche (approved):
Thanks!

PR #421, opened 2024-06-09
BUG: fix write of kml lon/lat transpose

### LINKED ISSUE #420: BUG:  long, lat geometry swapped ! geopandas.to_file( "out.kml", driver="KML",engine='pyogrio' )
Hello, opening an issue here that was raised in geopandas by u/[phisan-chula](https://github.com/phisan-chula)

https://github.com/geopandas/geopandas/issues/3332

because I wanted to have a go at submitting a PR on this. I assume pyogrio is the right place to put in a fix.

This seems to be an issue that came up in GDAL 3.0 when it started using the convention set by the CRS - for EPSG:4326 - that was lat/lon - in conflict with the lon/lat convention by WKT/WKB (to my understanding).

This resulted in the KML's exported inferring a lat/lon convention when writing despite the explicit lon/lat input by the geodataframe.

This was resolved by GDAL allowing the `AxisMappingStrategy` to be set as `OAMS_TRADITIONAL_GIS_ORDER` (i.e. lon/lat)

Apologies if I got any details above wrong, I am probably operating at the limit of my geospatial knowledge at the moment.

nicholas-ys-tan:
Closes #420

My cython is atrocious.

Followed what was done in fiona, qgis and gdal documentation to find this is probably what needs to be done. But I am probably doing it too broadly and may have unintended consequences in other filetypes.

~Still need to write tests.~

Need to investigate how this may impact other file types.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,7 +49,7 @@
 -   Fixed bug preventing reading from bytes or file-like in `read_arrow` /
     `open_arrow` (#407).
 -   Fixed bug transposing longitude and latitude when writing files with
-    co-ordinate transfomration to EPSG:4326 (#421).
+    coordinate transformation from EPSG:4326 (#421).
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,8 @@
     `write` and `write_dataframe` (#384).
 -   Fixed bug preventing reading from bytes or file-like in `read_arrow` /
     `open_arrow` (#407).
+-   Fixed bug transposing longitude and latitude when writing files with
+    co-ordinate transfomration to EPSG:4326 (#421).
 
 ### Packaging
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2134,8 +2134,8 @@ cdef create_ogr_dataset_layer(
         if crs is not None:
             try:
                 ogr_crs = create_crs(crs)
-                 # force geographic CRS to use lon, lat order and ignore axis order specified by CRS, in order
-                 # to correctly write KML and GeoJSON coordinates in correct order
+                # force geographic CRS to use lon, lat order and ignore axis order specified by CRS, in order
+                # to correctly write KML and GeoJSON coordinates in correct order
                 OSRSetAxisMappingStrategy(ogr_crs, OAMS_TRADITIONAL_GIS_ORDER)
 
 
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2740,6 +2740,3 @@ cdef create_fields_from_arrow_schema(
                 f"'{get_string(child.name)}' and type {get_string(child.format)}"
                 f"{gdal_msg}."
             )
-
-cdef void osr_set_traditional_axis_mapping_strategy(OGRSpatialReferenceH srs):
-    OSRSetAxisMappingStrategy(srs, OAMS_TRADITIONAL_GIS_ORDER)
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2134,7 +2134,9 @@ cdef create_ogr_dataset_layer(
         if crs is not None:
             try:
                 ogr_crs = create_crs(crs)
-                osr_set_traditional_axis_mapping_strategy(ogr_crs)
+                 # force geographic CRS to use lon, lat order and ignore axis order specified by CRS, in order
+                 # to correctly write KML and GeoJSON coordinates in correct order
+                OSRSetAxisMappingStrategy(ogr_crs, OAMS_TRADITIONAL_GIS_ORDER)
 
 
             except Exception as exc:
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2134,6 +2134,8 @@ cdef create_ogr_dataset_layer(
         if crs is not None:
             try:
                 ogr_crs = create_crs(crs)
+                osr_set_traditional_axis_mapping_strategy(ogr_crs)
+
 
             except Exception as exc:
                 if dataset_options != NULL:
@@ -2736,3 +2738,6 @@ cdef create_fields_from_arrow_schema(
                 f"'{get_string(child.name)}' and type {get_string(child.format)}"
                 f"{gdal_msg}."
             )
+
+cdef void osr_set_traditional_axis_mapping_strategy(OGRSpatialReferenceH srs):
+    OSRSetAxisMappingStrategy(srs, OAMS_TRADITIONAL_GIS_ORDER)
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -2071,7 +2071,8 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
 
 def test_write_kml_file_coordinates(tmp_path, use_arrow):
     # confirm reading with kml stores as lon/lat convention
-    gdf = gp.GeoDataFrame(
+    points = [Point(10, 20), Point(30, 40), Point(50, 60)]
+    gdf = gp.GeoDataFrame(geometry=points, crs="EPSG:4326")
         data={
             "label": ["A", "B", "C"],
             "geometry": [Point(10, 20), Point(30, 40), Point(50, 60)],
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -6,7 +6,7 @@ import locale
 import numpy as np
 import pytest
 
-from pyogrio import list_layers, read_info, __gdal_version__
+from pyogrio import list_layers, list_drivers, read_info, __gdal_version__
 from pyogrio.errors import DataLayerError, DataSourceError, FeatureError, GeometryError
 from pyogrio.geopandas import read_dataframe, write_dataframe, PANDAS_GE_20
 from pyogrio.raw import (
@@ -2083,23 +2083,27 @@ def test_write_kml_file_coordinate_order(tmp_path, use_arrow):
 
     assert np.array_equal(gdf_in.geometry.values, points)
 
-    # test appending to the existing file
-    points_append = [Point(70, 80), Point(90, 100), Point(110, 120)]
-    gdf_append = gp.GeoDataFrame(geometry=points_append, crs="EPSG:4326")
+    if "LIBKML" in list_drivers():
+        # test appending to the existing file only if LIBKML is available
+        # as it appears to fall back on LIBKML driver when appending.
+        points_append = [Point(70, 80), Point(90, 100), Point(110, 120)]
+        gdf_append = gp.GeoDataFrame(geometry=points_append, crs="EPSG:4326")
 
-    write_dataframe(
-        gdf_append,
-        output_path,
-        layer="tmp_layer",
-        driver="KML",
-        use_arrow=use_arrow,
-        append=True,
-    )
-    # force_2d used to only compare xy geometry as z-dimension is undesirably
-    # introduced when the kml file is over-written.
-    gdf_in_appended = read_dataframe(output_path, use_arrow=use_arrow, force_2d=True)
+        write_dataframe(
+            gdf_append,
+            output_path,
+            layer="tmp_layer",
+            driver="KML",
+            use_arrow=use_arrow,
+            append=True,
+        )
+        # force_2d used to only compare xy geometry as z-dimension is undesirably
+        # introduced when the kml file is over-written.
+        gdf_in_appended = read_dataframe(
+            output_path, use_arrow=use_arrow, force_2d=True
+        )
 
-    assert np.array_equal(gdf_in_appended.geometry.values, points + points_append)
+        assert np.array_equal(gdf_in_appended.geometry.values, points + points_append)
 
 
 @pytest.mark.requires_arrow_write_api
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -2079,7 +2079,7 @@ def test_write_kml_file_coordinates(tmp_path, use_arrow):
     )
     gdf = gdf.set_crs(4326)
     output_path = tmp_path / "test.kml"
-    write_dataframe(gdf, output_path, layer="tmp_layer", driver="KML")
+    write_dataframe(gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=use_arrow)
 
     gdf_in = read_dataframe(output_path)
 
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -2067,3 +2067,21 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
     )
     assert actual.columns[0] == mandarin
     assert actual[mandarin].values[0] == mandarin
+
+def test_write_kml_file_coordinates(tmp_path):
+    # confirm reading with kml stores as lon/lat convention
+    gdf = gp.GeoDataFrame(data={'label':['A', 'B', 'C'],
+                                'geometry': [Point(10, 20),
+                                             Point(30, 40),
+                                             Point(50, 60)],
+                                }
+                        )
+    gdf = gdf.set_crs(4326)
+    output_path = tmp_path / f"test.kml"
+    write_dataframe(gdf, output_path, layer='tmp_layer', driver="KML")
+
+    gdf_in = read_dataframe(output_path)
+
+    assert (gdf_in.geometry.values == [Point(10, 20), 
+                                       Point(30, 40), 
+                                       Point(50, 60)]).all()
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -2070,7 +2070,7 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
 
 
 def test_write_kml_file_coordinates(tmp_path, use_arrow):
-    # confirm reading with kml stores as lon/lat convention
+    # confirm KML coordinates are written in lon, lat order even if CRS axis specifies otherwise
     points = [Point(10, 20), Point(30, 40), Point(50, 60)]
     gdf = gp.GeoDataFrame(geometry=points, crs="EPSG:4326")
         data={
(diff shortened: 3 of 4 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,22 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
     )
     assert actual.columns[0] == mandarin
     assert actual[mandarin].values[0] == mandarin
+
+
+def test_write_kml_file_coordinates(tmp_path):
+    # confirm reading with kml stores as lon/lat convention
+    gdf = gp.GeoDataFrame(
```
brendan-ward, line 2074:
```suggestion
    points = [Point(10, 20), Point(30, 40), Point(50, 60)]
    gdf = gp.GeoDataFrame(geometry=points, crs="EPSG:4326")
```
(no need for attributes or `set_crs()` later)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,22 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
...
+            "geometry": [Point(10, 20), Point(30, 40), Point(50, 60)],
+        }
+    )
+    gdf = gdf.set_crs(4326)
+    output_path = tmp_path / "test.kml"
+    write_dataframe(gdf, output_path, layer="tmp_layer", driver="KML")
+
+    gdf_in = read_dataframe(output_path)
+
+    assert (
```
brendan-ward, line 2086:
```suggestion
    assert np.array_equal(gdf_in.geometry.values, points)
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,22 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
     )
     assert actual.columns[0] == mandarin
     assert actual[mandarin].values[0] == mandarin
+
+
+def test_write_kml_file_coordinates(tmp_path):
```
brendan-ward, line 2072:
```suggestion
def test_write_kml_coordinate_order(tmp_path):
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,22 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
...
+    # confirm reading with kml stores as lon/lat convention
```
brendan-ward, line 2073:
```suggestion
    # confirm KML coordinates are written in lon, lat order even if CRS axis specifies otherwise
```

brendan-ward (commented):
Thanks for working on this @nicholas-ys-tan !

Given that Fiona is setting traditional order after instantiating the CRS object, it seems reasonable for us to do the same, though I'm not aware of what the negative side effects of doing so would be.

A few comments below.

Would you mind adding a test case for GeoJSON written with RFC7946 as [described here](https://github.com/geopandas/pyogrio/issues/420#issuecomment-2157556937)

Something like this should work:

```Python
def test_write_geojson_rfc7946_coordinates(tmp_path, use_arrow=use_arrow):
    points = [Point(10, 20), Point(30, 40), Point(50, 60)]
    gdf = gp.GeoDataFrame(geometry=points, crs="EPSG:4326")
    output_path = tmp_path / "test.geojson"
    write_dataframe(
        gdf,
        output_path,
        layer="tmp_layer",
        driver="GeoJSON",
        RFC7946=True,
        use_arrow=use_arrow,
    )

    gdf_in = read_dataframe(output_path, use_arrow=use_arrow)

    assert np.array_equal(gdf_in.geometry.values, points)
```

jorisvandenbossche:
We should probably also test this for appending to an existing file (not entirely sure how that will work in that case, because then we don't create the layer object with constructed CRS object ?)

pyogrio/_io.pyx
```diff
@@ -2134,6 +2134,8 @@ cdef create_ogr_dataset_layer(
         if crs is not None:
             try:
                 ogr_crs = create_crs(crs)
+                osr_set_traditional_axis_mapping_strategy(ogr_crs)
```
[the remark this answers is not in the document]
nicholas-ys-tan: makes sense, thanks!

pyogrio/tests/test_geopandas_io.py
nicholas-ys-tan: updated as suggested (manually due to conflict)

nicholas-ys-tan:
Thanks @brendan-ward ,

Have added your changes and test, will just need to do this one tomorrow:

> We should probably also test this for appending to an existing file (not entirely sure how that will work in that case, because then we don't create the layer object with constructed CRS object ?)

pyogrio/_io.pyx
```diff
@@ -2134,6 +2134,10 @@ cdef create_ogr_dataset_layer(
         if crs is not None:
             try:
                 ogr_crs = create_crs(crs)
+                 # force geographic CRS to use lon, lat order and ignore axis order specified by CRS, in order
+                 # to correctly write KML and GeoJSON coordinates in correct order
```
theroggy, line 2138:
```suggestion
                # force geographic CRS to use lon, lat order and ignore axis order specified by CRS, in order
                # to correctly write KML and GeoJSON coordinates in correct order
```
Small indentation issue...

theroggy (commented):
Small detail: can you add an entry to CHANGES.md for you PR.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,35 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
     )
     assert actual.columns[0] == mandarin
     assert actual[mandarin].values[0] == mandarin
+
+
+def test_write_kml_file_coordinate_order(tmp_path, use_arrow):
```
brendan-ward, line 2072:
```suggestion
@pytest.mark.requires_arrow_write_api
def test_write_kml_file_coordinate_order(tmp_path, use_arrow):
```

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,35 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
...
+    write_dataframe(
+        gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=use_arrow
+    )
+
+    gdf_in = read_dataframe(output_path, use_arrow=use_arrow)
+
+    assert np.array_equal(gdf_in.geometry.values, points)
+
+
+def test_write_geojson_rfc7946_coordinates(tmp_path, use_arrow=use_arrow):
```
brendan-ward, line 2086:
```suggestion
@pytest.mark.requires_arrow_write_api
def test_write_geojson_rfc7946_coordinates(tmp_path, use_arrow=use_arrow):
```

brendan-ward (commented):
Sorry, I forgot to mention that the tests needed `@pytest.mark.requires_arrow_write_api` since they are writing using the Arrow API

This should fix the test failures on CI.

nicholas-ys-tan:
I am currently investigating this:

> We should probably also test this for appending to an existing file (not entirely sure how that will work in that case, because then we don't create the layer object with constructed CRS object ?)

There seems to be an issue unrelated to this ticket - I was testing the appending to a .kml file, and it introduces a Z-dimension. I am not sure if it is known behaviour but I couldn't find an existing ticket about it. Interestingly, it has nothing to do with `append` but  happens every time a `.kml` file is over-written.

```python
points = [Point(10, 20), Point(30, 40), Point(50, 60)]
gdf = gpd.GeoDataFrame(geometry=points, crs="EPSG:4326")
output_path = r'/home/nicholas/dev/data/ogr_test/temporary_kml_file.kml'
write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=True
)

gdf_read = read_dataframe(output_path, use_arrow=True)
print(gdf_read.geometry)
```
```
0    POINT (10 20)
1    POINT (30 40)
2    POINT (50 60)
Name: geometry, dtype: geometry
```
```python
points_append = [Point(70, 80), Point(90, 100), Point(110, 120)]
gdf_append = gpd.GeoDataFrame(geometry=points_append, crs="EPSG:4326")
write_dataframe(
    gdf_append, output_path, layer='tmp_layer', driver="KML", use_arrow=True, append=True
)
gdf_read_appended = read_dataframe(output_path, use_arrow=True)

print(gdf_read_appended.geometry)
```

```
0      POINT Z (10 20 0)
1      POINT Z (30 40 0)
2      POINT Z (50 60 0)
3      POINT Z (70 80 0)
4     POINT Z (90 100 0)
5    POINT Z (110 120 0)
Name: geometry, dtype: geometry
```

This doesn't seem to be related to the `append` feature, but to over-writing a kml file.

```python
points = [Point(10, 20), Point(30, 40), Point(50, 60)]
gdf = gpd.GeoDataFrame(geometry=points, crs="EPSG:4326")
output_path = r'/home/nicholas/dev/data/ogr_test/temporary_kml_file2.kml'
write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=True
)

write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=True
)

print(gdf_read.geometry)
```
```
0    POINT Z (10 20 0)
1    POINT Z (30 40 0)
2    POINT Z (50 60 0)
Name: geometry, dtype: geometry
```

On looking into the KML, can confirm it gets restructured

Before:
```kml
<?xml version="1.0" encoding="utf-8" ?>
<kml xmlns="http://www.opengis.net/kml/2.2">
<Document id="root_doc">
<Folder><name>tmp_layer</name>
  <Placemark>
      <Point><coordinates>10,20</coordinates></Point>
  </Placemark>
  <Placemark>
      <Point><coordinates>30,40</coordinates></Point>
  </Placemark>
  <Placemark>
      <Point><coordinates>50,60</coordinates></Point>
  </Placemark>
</Folder>
</Document></kml>
```
After over-writing with same data:
```kml
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2">
  <Document id="root_doc">
    <Document id="tmp_layer">
      <name>tmp_layer</name>
      <Placemark id="tmp_layer.1">
        <Point>
          <coordinates>
            10,20,0
          </coordinates>
        </Point>
      </Placemark>
      <Placemark id="tmp_layer.2">
        <Point>
          <coordinates>
            30,40,0
          </coordinates>
        </Point>
      </Placemark>
      <Placemark id="tmp_layer.3">
        <Point>
          <coordinates>
            50,60,0
          </coordinates>
        </Point>
      </Placemark>
    </Document>
  </Document>
</kml>
```

This happens in the main branch too, but with the reverse co-ordinates.

I assume this is not something to be addressed within this ticket. I haven't yet confirmed if this is an upstream bug but I am currently assuming it is. As a work-around, I've modified the test for the appended KML file to only compare the xy-coordinates by reading with `force_2d=True`.

If this isn't already a known issue, I'll do some digging as to why this is happening.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2067,3 +2067,73 @@ def test_non_utf8_encoding_shapefile_sql(tmp_path, use_arrow):
...
+        gdf_append,
+        output_path,
+        layer="tmp_layer",
+        driver="KML",
+        use_arrow=use_arrow,
+        append=True,
+    )
+    # force_2d used to only compare xy geometry as z-dimension is undesirably
+    # introduced when the kml file is over-written.
+    gdf_in_appended = read_dataframe(output_path, use_arrow=use_arrow, force_2d=True)
```
nicholas-ys-tan, line 2100: Wanted to bring your attention to this workaround with `force_2d=True` due to the introduction of the Z-dimension when appending to a kml file.

Would like to confirm this is okay and that we would not be looking to address the z-dimension aspect within this ticket.

nicholas-ys-tan:
the ubuntu-small CIs don't seem to like to open the produced KML files for appending - will need to look more into it this weekend.

theroggy:
> the ubuntu-small CIs don't seem to like to open the produced KML files for appending - will need to look more into it this weekend.
Not sure why this leads to appending nog working... but FYI: apparently [ubuntu-small](https://github.com/OSGeo/gdal/tree/master/docker#small-ghcrioosgeogdalubuntu-small-latest) doesn't contain the "LIBKML" driver, so kml handling will fallback there to the "KML" driver according to the GDAL LIBKML docs.

nicholas-ys-tan:
> Not sure why this leads to appending nog working... but FYI: apparently [ubuntu-small](https://github.com/OSGeo/gdal/tree/master/docker#small-ghcrioosgeogdalubuntu-small-latest) doesn't contain the [LIBKML driver](https://gdal.org/drivers/vector/libkml.html#vector-libkml), so according to de doc kml handling will fallback to the [KML driver](https://gdal.org/drivers/vector/kml.html#vector-kml) in that case.

Oh that is interesting, one thing I did not but neglected to mention in my above text about the Z-dimension is that when I run `write_dataframe` with LIBKML, it always writes out the Z-dimension even on first write.

This seems to suggest when there is an existing KML file, LIBKML is used to over-write the KML file, hence why the Z-dimension is introduced, and why it is failing in ubuntu-small.

```python
points = [Point(10, 20), Point(30, 40), Point(50, 60)]
gdf = gpd.GeoDataFrame(geometry=points, crs="EPSG:4326")
output_path = r'/home/nicholas/dev/data/ogr_test/temporary_kml_file7.kml'
write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="LIBKML", use_arrow=True
)
gdf_read = read_dataframe(output_path, use_arrow=True)

print(gdf_read.geometry)
```
```
0    POINT Z (10 20 0)
1    POINT Z (30 40 0)
2    POINT Z (50 60 0)
Name: geometry, dtype: geometry
```

I'll investigate further to confirm tomorrow morning.

pyogrio/tests/test_geopandas_io.py
brendan-ward: Yes, this is fine.  Adding the Z dimension happens in GDAL / underlying driver and outside our control, so it is fine to drop it again before verifying results since we are just focused on order of x,y coords.

brendan-ward:
It looks like always adding the Z value when using `LIBKML` is an [artifact of that particular driver](https://github.com/qgis/QGIS/issues/53362#issuecomment-1577026496) and also possibly because Z value is expected by the KML spec?  Whereas the `KML` driver does not automatically add Z values.

Can you log a separate issue that appending a KML file also adds Z value when using the `KML` driver?  Then we can look into that separately from this PR.  I'm wondering if in the case of append GDAL switches to using the `LIBKML` driver if available.

CHANGES.md
```diff
@@ -48,6 +48,8 @@
     `write` and `write_dataframe` (#384).
 -   Fixed bug preventing reading from bytes or file-like in `read_arrow` /
     `open_arrow` (#407).
+-   Fixed bug transposing longitude and latitude when writing files with
+    co-ordinate transfomration to EPSG:4326 (#421).
```
jorisvandenbossche, line 52:
```suggestion
-   Fixed bug transposing longitude and latitude when writing files with
    coordinate transformation from EPSG:4326 (#421).
```

jorisvandenbossche:
> > the ubuntu-small CIs don't seem to like to open the produced KML files for appending - will need to look more into it this weekend.
> 
> Not sure why this leads to appending not working... but FYI: apparently [ubuntu-small](https://github.com/OSGeo/gdal/tree/master/docker#small-ghcrioosgeogdalubuntu-small-latest) doesn't contain the [LIBKML driver](https://gdal.org/drivers/vector/libkml.html#vector-libkml), so according to de doc kml handling will fallback to the [KML driver](https://gdal.org/drivers/vector/kml.html#vector-kml) in that case.

The KML driver might just not support appending? In that case you will have to skip the test (or the second part of the test) when LIBKML driver is not available (you can use something like `"LIBKML" not in pyogrio.list_drivers()`)

BTW, the error message we give is a bit confusing. We are adding the suggestion _"It might help to specify the correct driver explicitly by prefixing the file path with '<DRIVER>:', e.g. 'CSV:path'"_, which is strange because when writing, the user already explicitly mentions the driver (which in practice is not used when appending I think, but that's not really clear for the user)

nicholas-ys-tan:
> The KML driver might just not support appending? In that case you will have to skip the test (or the second part of the test) when LIBKML driver is not available (you can use something like `"LIBKML" not in pyogrio.list_drivers()`)

Thanks @jorisvandenbossche , I think it's not just appending but also over-writing. It seems to use LIBKML if over-writing an existing file regardless of `append=True` which is quite interesting.

Thanks for the pointer, I was looking for how pyogrio is able to check what drivers are available. Will push the fixes shortly.

```python
points = [Point(10, 20), Point(30, 40), Point(50, 60)]
gdf = gpd.GeoDataFrame(geometry=points, crs="EPSG:4326")
output_path = r'/home/nicholas/dev/data/ogr_test/temporary_kml_file2.kml'
write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=True
)

write_dataframe(
    gdf, output_path, layer="tmp_layer", driver="KML", use_arrow=True
)

print(gdf_read.geometry)
```
```
0    POINT Z (10 20 0)
1    POINT Z (30 40 0)
2    POINT Z (50 60 0)
Name: geometry, dtype: geometry
```

jorisvandenbossche:
> I think it's not just appending but also over-writing. It seems to use LIBKML if over-writing an existing file regardless of `append=True` which is quite interesting.

Yes, but this test it's the append case that matters (so that's the reason I mentioned it explicitly). 
It seems that the KML driver essentially doesn't allow to open a file in Update mode (although I don't directly see anything in the GDAL code to support that), and when you are appending or writing to an existing file, in both cases our code in `ogr_open` will call `GDALOpenEx` with the `GDAL_OF_UPDATE` flag (when reading, we will call that with `GDAL_OF_READONLY` flag, that seems to be the main difference). Only in case of overwriting, we will then delete the layer if we could open the file and the layer exists, while in case of `append=True` we don't do that.

nicholas-ys-tan:
> Yes, but this test it's the append case that matters (so that's the reason I mentioned it explicitly). It seems that the KML driver essentially doesn't allow to open a file in Update mode (although I don't directly see anything in the GDAL code to support that), and when you are appending or writing to an existing file, in both cases our code in `ogr_open` will call `GDALOpenEx` with the `GDAL_OF_UPDATE` flag (when reading, we will call that with `GDAL_OF_READONLY` flag, that seems to be the main difference). Only in case of overwriting, we will then delete the layer if we could open the file and the layer exists, while in case of `append=True` we don't do that.

Thanks Joris, that all makes sense - I'll include this information in a separate ticket to document.

jorisvandenbossche (approved):

theroggy (approved):

brendan-ward (approved):
Thanks @nicholas-ys-tan  !

jorisvandenbossche:
With this fix in, shall we do a 0.8.1 release? (this seems like an important fix to get out for when geopandas switches to use pyogrio by default)

PR #422, opened 2024-06-13
ENH: add on_invalid parameter to read_dataframe

theroggy:
xref https://github.com/geopandas/geopandas/issues/2791


theroggy:
> Thanks a lot for this PR, it looks good (from an outsiders perspective).
> 
> Would in the future this be extendable with something like a `"repair"` or `"salvage"` option?

I was starting to answer... but it's probably better to give this topic it's own issue, so I created one: https://github.com/geopandas/pyogrio/issues/423

brendan-ward (approved):
Thanks @theroggy !

PR #429, opened 2024-06-18
Fix #421 entry in changelog from v0.8.0 to 0.9.0

nicholas-ys-tan:
Apologies, I had incorrectly entered the fix for #421 under v0.8.0 in my PR. This PR fixes that in the changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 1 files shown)


brendan-ward (approved):
Thanks @nicholas-ys-tan , apologies for not catching this during review of that PR.

PR #431, opened 2024-06-24
BUG: Enable mask & bbox filter when geometry column not read

snowman2:
Without this change, when a `bbox` or `mask` is passed in and the geometry is ignored, the filter excludes all rows and an empty response is returned.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 -   Silence warning from `write_dataframe` with `GeoSeries.notna()` (#435).
+-   BUG: Enable mask & bbox filter when geometry column not read (#431).
 
 ## 0.9.0 (2024-06-17)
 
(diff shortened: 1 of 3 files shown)


brendan-ward (commented):
Good catch @snowman2 !  Can you please add a changelog entry?  Assume the next release will be 0.9.1

snowman2:
> Can you please add a changelog entry? 

Sure - just added it.

CHANGES.md
```diff
@@ -1,5 +1,17 @@
...
+## 0.9.1 (2024-??-??)
+
+### Improvements
+
+
+### Bug fixes
+
+- BUG: Enable mask & bbox filter when geometry column not read (#431).
+
+### Packaging
```
brendan-ward, line 12:
```suggestion
```

(can remove unused headers)

brendan-ward (approved):

snowman2:
Thanks @brendan-ward 👍

PR #439, opened 2024-07-02
TST: Consolidate test marks that skip if shapely is not installed

brendan-ward:
We were already using `requires_shapely` test mark for several tests, but not tests that used the `mask` parameter.


theroggy (approved):
Looks good!

Issue #433, opened 2024-06-26
Data files license


avalentino:
I'm in the process of packaging pyogrio for Debian, I hope you are fine with it.
TO meet the Debian packaging standards I need to report the license for all files included in the package.
I would appreciate a lot if you could clarify what is the license of data files included in `pyogrio/tests/fixtures`, and in particular the license of:

* `poly_not_enough_points.shp.zip`
* `sample.osm.pbf`
* `test_fgdb.gdb.zip`
* `test_mixed_surface.gpkg`

The `pyogrio/tests/fixtures/README.md` seems to clarify what is the origin of some of the data files but the license for me is not clear.
Can I safely assume that data files are provided with the same license of the source code (MIT)?

martinfleis:
> Can I safely assume that data files are provided with the same license of the source code (MIT)?

I don't think so. At least the OSM sample retains [ODbL](https://opendatacommons.org/licenses/odbl/). I am not sure about the rest.

jorisvandenbossche:
> * poly_not_enough_points.shp.zip

This was recently added in https://github.com/geopandas/pyogrio/pull/422. @theroggy did you create this file manually? (would be good to add a note about that in the README then as well)

> * test_fgdb.gdb.zip

This is downloaded from https://trac.osgeo.org/gdal/raw-attachment/wiki/FileGDB/. @rouault do you know if this wiki falls under the general GDAL license?

> * test_mixed_surface.gpkg

This is extracted from one of the datasets from https://www.usgs.gov/national-hydrography/access-national-hydrography-products. I don't directly find anything on that page about the license of those datasets (maybe the USGS has a general license it uses for all available datasets?  but not familiar with it)

rouault:
> * test_fgdb.gdb.zip

Maybe @jmckenna remembers the provenance of this file ? Otherwise you could potentially switch to one of the GDAL autotest  suite samples: https://github.com/OSGeo/gdal/tree/master/autotest/ogr/data/filegdb

theroggy:
> > * poly_not_enough_points.shp.zip
> 
> This was recently added in #422. @theroggy did you create this file manually? (would be good to add a note about that in the README then as well)

No, I exported the polygon giving the issue from the file provided in this issue: https://github.com/geopandas/geopandas/discussions/3336 

So, not sure about licensing :-(... @EwoutH can you shed some light?

EwoutH:
[...]

PR #441, opened 2024-07-04
TST: Replace test datasets with pyogrio-generated files where possible

brendan-ward:
Per #433, we were including test data files from other parties and may not have been in great compliance with their licenses, even though we were usually using a tiny extract of the source files.

See the updates to `tests/fixtures/README.md` first for updated guidance on managing test files, and steps used to create those still present as files.

Where we were using GeoJSON files, I migrated those directly into `conftest.py` so they are generated using code, which should make the licensing of those files more clear (i.e., part of Pyogrio source code so our MIT license is clear).  I don't have a strong opinion about whether these should be files or in code, but putting them in code seemed reasonable; let me know if you disagree.

This also made it easier to create an invalid GeoJSON file for testing polygons with insufficient coordinates (replaces data file with problematic license).

Where we were using 3rd party FGDB datasets, I created new datasets to mimic how we were using those.  I used QGIS to hand-digitize (super simple) LineString ZM, Polygon ZM, Curve, CurvePolygon, and MultiSurface datasets, since we were using one of the FGDB datasets to verify that they were correctly downgraded to supported geometry types.


pyogrio/tests/test_geopandas_io.py
```diff
@@ -196,61 +194,82 @@ def test_read_no_geometry_no_columns_no_fids(naturalearth_lowres, use_arrow):
         )
 
 
-def test_read_force_2d(test_fgdb_vsi, use_arrow):
-    with pytest.warns(
```
brendan-ward, line 200: Verifying the warning was incidental to this test, and already covered by `test_core.py::test_list_layers`.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -196,61 +194,82 @@ def test_read_no_geometry_no_columns_no_fids(naturalearth_lowres, use_arrow):
...
+
+    expected2 = gp.GeoDataFrame(geometry=[Point(1, 1)], crs="EPSG:4326")
+    write_dataframe(expected2, filename, layer="layer2", append=True)
+
+    assert np.array_equal(
+        list_layers(filename), [["layer1", "Point"], ["layer2", "Point"]]
+    )
 
-@pytest.mark.filterwarnings("ignore: Measured")
-@pytest.mark.filterwarnings("ignore: More than one layer found in")
```
brendan-ward, line 217: We actually should verify that the multiple layers warning is raised.

theroggy (approved):
Looks good! Will also be easier to maintain!

pyogrio/tests/conftest.py
```diff
@@ -123,28 +124,165 @@ def naturalearth_lowres_vsi(tmp_path, naturalearth_lowres):
 
 
 @pytest.fixture(scope="session")
-def test_fgdb_vsi():
-    return f"/vsizip/{_data_dir}/test_fgdb.gdb.zip"
```
martinfleis, line 127: Does it matter that we don't have a direct replacement for a zipped FGB?

jorisvandenbossche (approved):
This looks great!

One remark: I am not sure we still have coverage for reading an FGDB file? (now that it is removed as fixture file) Probably not that important, but I see we have a `test_write_openfilegdb` that only asserts the file exists on disk, maybe we can just read the resulting file as well to test the full roundtrip.

> Where we were using GeoJSON files, I migrated those directly into conftest.py so they are generated using code, which should make the licensing of those files more clear (i.e., part of Pyogrio source code so our MIT license is clear). I don't have a strong opinion about whether these should be files or in code, but putting them in code seemed reasonable; let me know if you disagree.

Those files were generated manually, so their licensing was OK anyway I think. But also no strong opinion. IIRC I added several of them, and while developing / testing, I think it is a bit easier to have them as files (I typically created them with gdal/pyogrio or manually edited them afterwards) and sometimes tested/compared directly with `ogrinfo`. But at the end moving the text of the file into the conftest.py is of course trivial, so it's perfectly fine going that way.

pyogrio/tests/conftest.py
brendan-ward: I don't think it matters.  We have other tests that use the `/vsizip/` interface for working with a zipped shapefile, which should be a reasonable proxy for zip files containing other formats.

brendan-ward:
Added tests to verify roundtrip using OpenFileGDB driver, and test of int64 dtype handling for GDAL >= 3.9 via dataset creation option.

martinfleis (approved):

Issue #430, opened 2024-06-24
`DataLayerError` when writing using `fsspec` 


DahnJ:
## Description
pyogrio raises `DataLayerError: The layer name may not contain special characters or spaces` when passing in file object using `fsspec`'s `file.open`

<details> <summary> Traceback </summary>

```python

CPLE_AppDefinedError                      Traceback (most recent call last)
File ~/miniconda3/envs/geo312/lib/python3.12/site-packages/pyogrio/_io.pyx:2183, in pyogrio._io.create_ogr_dataset_layer()

File ~/miniconda3/envs/geo312/lib/python3.12/site-packages/pyogrio/_err.pyx:183, in pyogrio._err.exc_wrap_pointer()

CPLE_AppDefinedError: The layer name may not contain special characters or spaces

During handling of the above exception, another exception occurred:

DataLayerError                            Traceback (most recent call last)
Cell In[8], line 10
      7 gdf = gpd.read_file(path)
      9 with fs.open('test.gpkg', mode='wb') as file:
---> 10     gdf.to_file(file, driver='GPKG')

File ~/miniconda3/envs/geo312/lib/python3.12/site-packages/geopandas/geodataframe.py:1536, in GeoDataFrame.to_file(self, filename, driver, schema, index, **kwargs)
   1441 """Write the ``GeoDataFrame`` to a file.
   1442 
   1443 By default, an ESRI shapefile is written, but any OGR data source
   (...)
   1532 
   1533 """
   1534 from geopandas.io.file import _to_file
-> 1536 _to_file(self, filename, driver, schema, index, **kwargs)

File ~/miniconda3/envs/geo312/lib/python3.12/site-packages/geopandas/io/file.py:686, in _to_file(df, filename, driver, schema, index, mode, crs, engine, metadata, **kwargs)
    683     raise ValueError(f"'mode' should be one of 'w' or 'a', got '{mode}' instead")
    685 if engine == "pyogrio":
--> 686     _to_file_pyogrio(df, filename, driver, schema, crs, mode, metadata, **kwargs)
    687 elif engine == "fiona":
    688     _to_file_fiona(df, filename, driver, schema, crs, mode, metadata, **kwargs)

File ~/miniconda3/envs/geo312/lib/python3.12/site-packages/geopandas/io/file.py:748, in _to_file_pyogrio(df, filename, driver, schema, crs, mode, metadata, **kwargs)
    745 if not df.columns.is_unique:
[...]

PR #442, opened 2024-07-04
FIX: Raise NotImplementedError if user passes an open file handle to write

brendan-ward:
As indicated in #430 and [geopandas #3369](https://github.com/geopandas/geopandas/discussions/3369), if a user passed an open file handle to write, we were failing with confusing error messages.  While the docstrings in pyogrio specify the allowed inputs, we weren't strictly verifying those, and so any path that was not BytesIO was first getting cast to a string, and then failing badly from there.

These usages worked for users with the Fiona engine (previous default) in GeoPandas, so it is proving to be a point of confusion now that Pyogrio is the default engine and things are no longer working as expected.

In most cases, writing the dataset to BytesIO and then writing those bytes to the file handle (regular file handle via `open`, an fsspec opened file handle, or an open handle within a `ZipFile`), so the error message includes a brief instruction to do that instead.  That also helps defer catching issues around supported drivers and lack of append support to the BytesIO handler.

In the mid term, we may be able to detect that the user passed a writeable file handle, and wrap in BytesIO ourselves, but this is more involved than I had time to take on at the moment.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 -   Silence warning from `write_dataframe` with `GeoSeries.notna()` (#435).
 -   Enable mask & bbox filter when geometry column not read (#431).
--   Raise NotImplmentedError when user attempts to write to an open file handle.
+-   Raise NotImplmentedError when user attempts to write to an open file handle (#442).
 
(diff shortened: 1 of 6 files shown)


theroggy (approved):
Looks good.

PR #443, opened 2024-07-04
Fix reading from zstd decompression stream

ChristofKaufmann:
Since 0.8.0 reading a zstandard compressed file does not work anymore. The following code works in 0.7.2, but not in 0.8.0 and 0.9.0:

``` python
import geopandas as gpd
import zstandard
import io

# make an uncompressed.geojson and a compressed.geojson.zst (fixture)
gdf = gpd.GeoDataFrame(geometry=[Point(4, 2), Point(4, 2)])

gdf.to_file('uncompressed.geojson', driver='GeoJSON')

with open('uncompressed.geojson', 'rb') as ro:
    content = ro.read()

compressed_content = zstandard.compress(content)
with open('compressed.geojson.zst', 'wb') as w:
    w.write(compressed_content)

# try to read geojson from BytesIO
# works in 0.7.2 and 0.9.0
with open('uncompressed.geojson', 'rb') as fh:
    mem = io.BytesIO(fh.read())

gdf = gpd.read_file(mem)
print(gdf)

# try to read geojson from file handle
# works in 0.7.2 and 0.9.0
with open('uncompressed.geojson', 'rb') as fh:
    gdf = gpd.read_file(fh)

print(gdf)


# try to read geojson from zstandard decompression stream
# works in 0.7.2, but not in 0.9.0
with open('compressed.geojson.zst', 'rb') as fh:
    dctx = zstandard.ZstdDecompressor()
    with dctx.stream_reader(fh) as reader:
        gdf = gpd.read_file(reader)

print(gdf)
```

The last test results in the stack trace:
``` output
  File "<stdin>", line 4, in <module>
  File "/usr/local/lib/python3.11/dist-packages/geopandas/io/file.py", line 294, in _read_file
    return _read_file_pyogrio(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/geopandas/io/file.py", line 547, in _read_file_pyogrio
    return pyogrio.read_dataframe(path_or_bytes, bbox=bbox, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/pyogrio/geopandas.py", line 252, in read_dataframe
    result = read_func(
             ^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/pyogrio/raw.py", line 197, in read
    get_vsi_path_or_buffer(path_or_buffer),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/pyogrio/util.py", line 43, in get_vsi_path_or_buffer
    path_or_buffer.seek(0)
OSError: cannot seek zstd decompression stream backwards
```

A zstd decompression stream can be seeked [only forwards](https://python-zstandard.readthedocs.io/en/latest/decompressor.html#zstddecompressionreader), but not backwards. [Hence](https://docs.python.org/3/library/io.html#io.IOBase.seekable) its seekable method returns `False`. The issue has been introduced in #397.

This PR fixes it. Instead of only checking, whether the `seek` method exists, this PR checks whether the `seekable` method exists and returns `True`. Only then `seek(0)` is used.

Do you want me to add any test cases for this? If so, based on `zstandard`?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 -   Silence warning from `write_dataframe` with `GeoSeries.notna()` (#435).
 -   BUG: Enable mask & bbox filter when geometry column not read (#431).
+-   Prevent seek on read from compressed inputs (#443).
 
 ## 0.9.0 (2024-06-17)
 
--- a/pyogrio/tests/conftest.py
+++ b/pyogrio/tests/conftest.py
@@ -1,3 +1,4 @@
+from io import BytesIO
 from pathlib import Path
 from zipfile import ZIP_DEFLATED, ZipFile
 
@@ -176,6 +177,31 @@ def geojson_filelike(tmp_path):
 
     with open(filename, "rb") as f:
         yield f
+
+
+@pytest.fixture(scope="function")
+def nonseekable_bytes(tmp_path):
+    # mock a non-seekable byte stream, such as a zstandard handle
+    class NonSeekableBytesIO(BytesIO):
+        def seekable(self):
+            return False
+
+        def seek(self, *args, **kwargs):
+            raise OSError("cannot seek")
+
+    # wrap GeoJSON into a non-seekable BytesIO
+    geojson = """{
+        "type": "FeatureCollection",
+        "features": [
+            {
+                "type": "Feature",
+                "properties": { },
+                "geometry": { "type": "Point", "coordinates": [1, 1] }
+            }
+        ]
+    }"""
+
+    return NonSeekableBytesIO(geojson.encode("UTF-8"))
 
 
 @pytest.fixture(
--- a/pyogrio/tests/test_arrow.py
+++ b/pyogrio/tests/test_arrow.py
@@ -168,6 +168,12 @@ def test_read_arrow_bytes(geojson_bytes):
     assert len(table) == 3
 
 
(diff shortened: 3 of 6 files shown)


brendan-ward:
Good catch, and sorry for introducing the bug.

I don't want us to add `zstandard` as a dependency, even for testing, so I created a very lightweight mock class that provides a non-seekable byte stream, and added it to the tests I'd want to see.

Subsequent read operations against the non-seekable bytes (also for the zstandard stream) will fail, but I'm assuming that is well expected for this use case.

ChristofKaufmann:
Very nice unit tests! It would have been difficult for me to write these. Thank you!

I agree. It is expected that streams can be read only once.

Before merging, do you want to have the Docker GDAL Test with the latest ubuntu fixed (probably just add `--break-system-packages` to `pip install`)?  Or is there anything else left to do here where I can help with?

brendan-ward:
Fix for latest GDAL is already in #438, no need for that to be here.

theroggy (approved):
Looks fine to me!

PR #452, opened 2024-07-18
MAINT: change conda dependency to libgdal-core + remove dependency on pyproj

### LINKED ISSUE #456: When geopandas is available, pyogrio expects pyproj to be available
Since geopandas 1.0, geopandas-base doesn't imply anymore that pyproj will be installed as well.

But, pyogrio now expects it to be.

Noticed in #452.

theroggy:
Switch to `libgdal-core` as dependency instead of `libgdal`. This package is significantly smaller as it doesn't contain some large GDAL plugins. Extra plugins can be installed as seperate conda packages if needed: more info [here](https://gdal.org/download.html#conda).

This PR also removes the `pyogrio` dependency on `pyproj`, as `pyproj` is not part of `libgdal-core` nor `geopandas` since `geopandas 1.0`: closes #456

Reference https://github.com/conda-forge/pyogrio-feedstock/issues/48

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,8 +7,8 @@
 -   For the conda-forge package, change the dependency from `libgdal` to
     `libgdal-core`. This package is significantly smaller as it doesn't contain
     some large GDAL plugins. Extra plugins can be installed as seperate conda
-    packages if needed: more info [here](https://gdal.org/download.html#conda)
-    (#452).
+    packages if needed: more info [here](https://gdal.org/download.html#conda).
+    This also leads to `pyproj` becoming an optional dependency (#452).
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 ## 0.9.1 (yyyy-mm-dd)
 
+### Packaging
+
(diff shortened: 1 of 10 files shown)


jorisvandenbossche:
Looking good!

CHANGES.md
```diff
@@ -2,6 +2,14 @@
 
 ## 0.9.1 (yyyy-mm-dd)
 
+### Packaging
+
+-   For the conda-forge package, change the dependency from `libgdal` to
+    `libgdal-core`. This package is significantly smaller as it doesn't contain
+    some large GDAL plugins. Extra plugins can be installed as seperate conda
+    packages if needed: more info [here](https://gdal.org/download.html#conda).
+    This also leads to `pyproj` becoming an optional dependency (#452).
```
brendan-ward, line 11:
```suggestion
    This also leads to `pyproj` becoming an optional dependency; you will need to install `pyproj` in order to support spatial reference systems (#452).
```

brendan-ward (approved):
Thanks @theroggy ; just a minor comment to make it more explicit that folks need to install `pyproj`.  Please merge once that is in (if you agree with the changes).

PR #457, opened 2024-08-03
ENH: add support to read, write, list and remove `/vsimem/` files

theroggy:
Add support for the basic operations on /vsimem/ files: read, write, list and remove.

TODO:

- [x] maybe add some examples in the manual? -> https://github.com/geopandas/pyogrio/pull/477
- [x] remove pyproj again from ci/envs/latest.yml

Triggered by a topic in the "discussion" section of geopandas, where someone noted that pre-geopandas 1.0 this worked and now it does not anymore: https://github.com/geopandas/geopandas/discussions/3376

<-- *clarified below*
I now noticed the following statement in: https://github.com/geopandas/pyogrio/pull/407

> This also adds a specific check that the incoming path does not already contain /vsimem/, since we have to handle that internally and cannot allow it to be passed in. 

@brendan-ward I wonder... is this meant as:
- an explicit choice not ever to want to allow /vsimem/ inputs because ???
- or a temporary measure till support has been built in
- or is there some technical reason that this causes problems that I didn't think about/haven't encountered yet

-->


brendan-ward:
> a temporary measure till support has been built in

Most likely this, though probably wasn't going to prioritize fixing it unless there was a need (apparently there is).

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
             CPLFree(vsi_buffer)
 
 
-cdef delete_vsimem_file(str path):
-    """ Recursively delete in-memory path or directory containing path
+cpdef vsimem_rmtree_toplevel(str path):
+    """Remove the toplevel file or toplevel directory containing the file.
```
brendan-ward, line 129:
```suggestion
    """Remove the top-level file or toplevel directory containing the file.
```
(and below)

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+    cdef VSIStatBufL st_buf
 
     if "/vsimem/" not in path:
-        return
+        raise ValueError(f"Path is not a /vsimem/ path: '{path}'")
+
+    # Determine the toplevel directory of the file
+    mempath_parts = path.split("/vsimem/")[1].split("/")
+    if len(mempath_parts) == 0:
+        raise OSError("Removing /vsimem/ is not supported")
```
brendan-ward, line 160:
```suggestion
        raise OSError("path to in-memory file or directory is required")
```

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+
+    toplevel_path = f"/vsimem/{mempath_parts[0]}"
+
+    if not VSIStatL(toplevel_path.encode("UTF-8"), &st_buf) == 0:
+        raise FileNotFoundError(f"Path does not exist: '{path}'")
+    if VSI_ISDIR(st_buf.st_mode):
```
brendan-ward, line 166: add newline above

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
-    VSIRmdirRecursive(root.encode("UTF-8"))
+    """
+    cdef const char *path_c
+    cdef int n
+    cdef char** papszFiles
+    cdef VSIStatBufL st_buf
+
+    try:
+        path_b = path.encode("UTF-8")
+    except UnicodeDecodeError:
```
brendan-ward, line 196: I don't think we allow a fallback on `UnicodeDecodeError` elsewhere; is there a specific reason to do so here?

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+        path_b = path
+    path_c = path_b
```
brendan-ward, line 198: newline above

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+    if not VSIStatL(path_c, &st_buf) == 0:
+        raise FileNotFoundError(f"Path does not exist: '{path}'")
+    if not VSI_ISDIR(st_buf.st_mode):
```
brendan-ward, line 201: newline above

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+        raise NotADirectoryError(f"Path is not a directory: '{path}'")
+
+    papszFiles = VSIReadDirRecursive(path_c)
+    n = CSLCount(<CSLConstList>papszFiles)
+    files = []
+    for i in range(n):
+        files.append(papszFiles[i].decode("UTF-8"))
+    CSLDestroy(papszFiles)
```
brendan-ward, line 209: The above block (lines 204-208) should probably be in a `try` block and then have 

```python
finally:
    CSLDestroy(papszFiles)
```

to ensure this always gets cleaned up.

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+    # Prepend files with the base path
+    if not path.endswith("/"):
+        path = f"{path}/"
+    files = [f"{path}{file}" for file in files]
+    
+    return files
+
+
+def ogr_vsi_rmtree(str path):
+    """Recursively remove vsi directory.
```
brendan-ward, line 224:
```suggestion
    """Recursively remove VSI in-memory directory.
```

pyogrio/_vsi.pyx
```diff
@@ -118,23 +125,152 @@ cdef read_vsimem_to_buffer(str path, object out_buffer):
...
+        path_b = path.encode("UTF-8")
+    except UnicodeDecodeError:
+        path_b = path
+    path_c = path_b
+    if not VSIStatL(path_c, &st_buf) == 0:
+        raise FileNotFoundError(f"Path does not exist: '{path}'")
+    if not VSI_ISDIR(st_buf.st_mode):
+        raise NotADirectoryError(f"Path is not a directory: '{path}'")
+    if path.endswith("/vsimem") or path.endswith("/vsimem/"):
+        raise OSError("Removing /vsimem/ is not supported")
```
brendan-ward, line 245:
```suggestion
        raise OSError("path to in-memory file or directory is required")
```

pyogrio/core.py
```diff
@@ -1,5 +1,8 @@
 """Core functions to interact with OGR data sources."""
 
+from pathlib import Path
+from typing import Optional, Union
```
brendan-ward, line 4: I don't believe we've integrated typing elsewhere?  Is the idea here to start progressively adding it?

pyogrio/core.py
```diff
@@ -326,3 +335,76 @@ def get_gdal_data_path():
 
     """
     return _get_gdal_data_path()
+
+
+def vsi_listtree(path: Union[str, Path], pattern: Optional[str] = None):
+    """Recursively list the contents in a vsi directory.
```
brendan-ward, line 341:
```suggestion
    """Recursively list the contents of a VSI directory.
```

pyogrio/_vsi.pyx
brendan-ward: Should these new `ogr_vsi_*` functions be limited to `/vsimem/` paths, or are you intending that they should work for some of the other VSI interfaces as well?

pyogrio/tests/test_core.py
```diff
@@ -151,7 +158,12 @@ def test_list_drivers():
     assert len(drivers) == len(expected)
 
 
-def test_list_layers(naturalearth_lowres, naturalearth_lowres_vsi, test_fgdb_vsi):
+def test_list_layers(
+    naturalearth_lowres,
+    naturalearth_lowres_vsi,
+    naturalearth_lowres_vsimem,
+    test_fgdb_vsi,
```
brendan-ward, line 165:
```suggestion
```
(removed by another PR)

brendan-ward (commented):
Thanks for working on this @theroggy and apologies for the slow review!  I only had a chance to review up to your tests and will try to review those in the next couple of days.

pyogrio/_vsi.pyx
theroggy: I don't quite remember. I'll remove it and if the tests don't fail I suppose it should be OK.

pyogrio/_vsi.pyx
theroggy: I gave this some thought, and couldn't really find a good reason to limit it to vsimem support. Storing files in the cloud is also popular I suppose, so I can imagine it might be useful to just expose the full GDAL capabilities?

pyogrio/core.py
theroggy: Actually I don't recall giving it a conscious thought... so I suppose it was rather out of habit as I normally write typed. I think it might be a good idea to start using it, espacially for pyogrio I think it wouldn't be to hard for most parameters to type them. But I wouldn't mind removing the typing either.

pyogrio/_io.pyx
```diff
@@ -2291,12 +2304,16 @@ def ogr_write(
 
     try:
         # Setup in-memory handler if needed
-        path = get_ogr_vsimem_write_path(path_or_fp, driver)
-        is_vsi = path.startswith('/vsimem/')
+        if isinstance(path_or_fp, str):
```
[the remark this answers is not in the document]
theroggy: Oops... I was planning to do that, but apparently I forgot about it in the end.

Cleaned up.

pyogrio/tests/test_core.py
```diff
@@ -611,3 +613,137 @@ def test_error_handling_warning(capfd, naturalearth_lowres):
...
 
     assert capfd.readouterr().err == ""
+
+
+def test_vsimem_listtree(naturalearth_lowres):
+    # Prepare test data in /vsimem
+    meta, _, geometry, field_data = read(naturalearth_lowres)
+    meta["spatial_index"] = False
+    meta["geometry_type"] = "MultiPolygon"
+    test_file_path = Path(f"/vsimem/pyogrio_test_{naturalearth_lowres.stem}.gpkg")
```
brendan-ward, line 623:
```suggestion
    test_file_path = Path("/vsimem/pyogrio_test_naturalearth_lowres.gpkg")
```

(slight preference to simplify and hardcode filename name)

pyogrio/tests/test_core.py
```diff
@@ -611,3 +613,137 @@ def test_error_handling_warning(capfd, naturalearth_lowres):
...
+    vsi_rmtree(test_dir_path.parent)
+    files = vsi_listtree("/vsimem/")
+    assert test_file_path.as_posix() in files
+    assert test_dir_path.as_posix() not in files
+
+    # Remove test_file
+    vsi_unlink(test_file_path)
+
+
+def test_vsimem_unlink_error(naturalearth_lowres_vsimem):
```
brendan-ward, line 695: Also would be good to parametrize this one

pyogrio/core.py
```diff
@@ -326,3 +335,76 @@ def get_gdal_data_path():
...
+        path to vsimem file to be removed
+
+    """
+    if isinstance(path, Path):
+        path = path.as_posix()
+
+    ogr_vsi_unlink(path)
+
+
+def _vsimem_rmtree_toplevel(path: Union[str, Path]):
```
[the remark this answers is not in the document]
theroggy: OK!

pyogrio/tests/test_core.py
```diff
@@ -611,3 +613,137 @@ def test_error_handling_warning(capfd, naturalearth_lowres):
...
+    assert test_file_path.as_posix() in files
+    assert test_dir_path.as_posix() not in files
+
+    # Clean up
+    vsi_unlink(test_file_path)
+    vsi_rmtree(test_dir_path.parent)
+
+
+def test_vsimem_rmtree_error(naturalearth_lowres_vsimem):
+    with pytest.raises(NotADirectoryError, match="Path is not a directory"):
```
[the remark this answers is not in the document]
theroggy: Because in one of the tests the fixture is used, there isn't a super clean solution to do this as far as I know... but I gave it a try to parametrize anyway...

pyogrio/tests/test_core.py
theroggy: Same here. If you think applying the solution for test_vsimem_rmtree_error would make this test clearer, I'll apply it here as well.

pyogrio/tests/test_path.py
```diff
@@ -339,19 +342,17 @@ def test_uri_s3_dataframe(aws_env_setup):
...
-    assert get_vsi_path_or_buffer(path) == str(path)
+@pytest.mark.parametrize(
+    "path, expected",
+    [
+        (Path("/tmp/test.gpkg"), str(Path("/tmp/test.gpkg"))),
+        (Path("/vsimem/test.gpkg"), "/vsimem/test.gpkg"),
+    ],
+)
+def test_get_vsi_path_or_buffer_obj_to_string(path, expected):
+    assert get_vsi_path_or_buffer(path) == expected
```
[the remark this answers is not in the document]
theroggy: I added some inline doc explaining why the test is like this. Does this clarify it, or is there a way to simplify the parameters anyway without breaking the test?

pyogrio/__init__.py
```diff
@@ -21,6 +21,9 @@
     read_bounds,
     read_info,
     set_gdal_config_options,
+    vsi_listtree,
+    vsi_rmtree,
+    vsi_unlink,
```
jorisvandenbossche, line 26: Do we explicitly want to expose those functions publicly (top-level) to our users as well? (are there specific use cases we are thinking of?) 
Because if this is mostly for the tests, we can also import from `pyogrio.core` or `pyogrio.util`

pyogrio/tests/test_core.py
```diff
@@ -611,3 +612,94 @@ def test_error_handling_warning(capfd, naturalearth_lowres):
...
+        ("/vsimem", False, OSError, "path to in-memory file or directory is required"),
+        ("/vsimem/", False, OSError, "path to in-memory file or directory is required"),
+    ],
+)
+def test_vsimem_rmtree_error(path, is_fixture, exception, error_match, request):
+    if is_fixture:
+        path = request.getfixturevalue(path)
+
+    with pytest.raises(exception, match=error_match):
+        vsi_rmtree(path)
```
jorisvandenbossche, line 666: It might be that you already discussed this with Brendan (didn't go through all previous review comments), but personally I would find this easier to read if we just write out those three cases:

```
    with pytest.raises(FileNotFoundError, match="Path does not exist"):
        vsi_rmtree("/vsimem/non-existent")

    with pytest.raises(OSError, match="path to in-memory file or directory is required"):
        vsi_rmtree("/vsimem")

    with pytest.raises(OSError, match="path to in-memory file or directory is required"):
        vsi_rmtree("/vsimem/")
```

I would say that is not too much repetition and the little duplication is worth the readability.

jorisvandenbossche: I see below it was indeed a question to parametrize. But maybe you could then at least split the test in two tests, one that uses the fixture, and one that is parametrized on the other paths/exception/message combos?

pyogrio/tests/test_core.py
```diff
@@ -611,3 +612,94 @@ def test_error_handling_warning(capfd, naturalearth_lowres):
...
+
+    # If path was a fixture, the temp file should still exists.
+    if is_fixture:
+        assert path.as_posix() in vsi_listtree("/vsimem")
+
+
+def test_vsimem_rmtree_unlink(naturalearth_lowres):
```
jorisvandenbossche, line 673: This one seems for a large part the same as `test_vsimem_listtree` above? Maybe those could be combined (because to test listtree, you also need to rmtree/unlink the result, so then maybe could just test them together)

pyogrio/tests/test_core.py
jorisvandenbossche: As mentioned above, my personal preference would be to keep this test as is.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -1568,6 +1584,21 @@ def test_write_read_null(tmp_path, use_arrow):
     assert result_gdf["object_str"][2] is None
 
 
+def test_write_read_vsimem(naturalearth_lowres_vsi, use_arrow_write):
```
jorisvandenbossche, line 1587: Can you use `use_arrow` and then decorate with `@pytest.mark.requires_arrow_write_api`, which seems to be how the other tests are set up here? (then you don't need to define the new `use_arrow_write` fixture)

pyogrio/tests/test_path.py
```diff
@@ -85,6 +86,8 @@ def change_cwd(path):
             "s3://testing/test.zip!a/b/item.shp",
             "/vsizip/vsis3/testing/test.zip/a/b/item.shp",
         ),
+        ("/vsimem/data.gpkg", "/vsimem/data.gpkg"),
+        (Path("/vsimem/data.gpkg"), "/vsimem/data.gpkg"),
```
jorisvandenbossche, line 90: I am not entirely sure we should allow `Path` objects with VSI file paths (that's not what they are meant for), but I suppose this is just supported because we all Paths in general as input, and not because we support this use case explicitly?

jorisvandenbossche: Although seeing the test `test_get_vsi_path_or_buffer_obj_to_string` just below, indicating that we explicitly ensure forward slashes are preserved, this support is actually quite explicit?

jorisvandenbossche (commented):
Great work!

pyogrio/__init__.py
brendan-ward: I struggled with this as well.  Some of them may be useful in order to leverage GDAL's VSI interface.  That said, if you are using a different package to create VSI resources (e.g., the `gdal` package used directly), maybe you should use that other package to handle any related cleanup.

If we do want to keep this public, want to put these in the top-level interface?  Maybe we should move the VSI related functions to `vsi.py` and expose these there, so that they are available, but compartmentalized for advanced usage a bit better?

pyogrio/tests/test_core.py
brendan-ward: I don't have strong opinions to parametrize if it makes things less readable.  Splitting the tests (having `is_fixture` parametrized feels a bit awkward) also is OK.

pyogrio/tests/test_path.py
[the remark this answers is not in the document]
brendan-ward: This is good, I wasn't thinking about Windows needing alternative paths.

brendan-ward:
I think it can be done in a later PR (but preferably for 0.10.0), but we should probably document how the `/vsimem` interface can be used more directly (e.g., alongside `gdal` package) as well as the VSI functions that we want to expose publicly here (which is why I think it is should be follow-up, so that this PR focuses on the API).

.github/workflows/lint.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+    types: [opened, edited]
```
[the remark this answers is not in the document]
theroggy: I don't remember anymore. It's weird being in this PR anyway, so probably just something I tried? I rolled back the change.

pyogrio/__init__.py
theroggy: I think it is practical/logical when it is possible to create files in e.g. /vsimem/ that the functions to remove them again are available as well without having to install another packages...

Putting them in a seperate interface is obviously possible, but I don't really see the point. pyogrio doesn't have that many functions, and I was quite happy that the hierarchy was removed in shapely so introducing it here now...

pyogrio/tests/test_core.py
theroggy: It's a pity that using fixtures combined with `@pytest.mark.parametrize` isn't better supported. However, artificially splitting the test in two tests doesn't make it clearer for e.g. new contributers (and for me in a few months ;-)) either I think... So I'll just rollback the changes if that's OK.

brendan-ward: Rolling back is just fine; apologies for the poor suggestion.

pyogrio/tests/test_path.py
theroggy: I don't quite know what you mean. But, anyway, we cannot control how a user passes in the path (as a Path or as a str)...

pyogrio/tests/test_core.py
theroggy: They are indeed very similar... I'll merge them

pyogrio/tests/test_core.py
theroggy: No problem... I took me a while to remember the complications of using a fixture with parametrize and I only ran into them a some weeks ago when I created the PR ;-)...

pyogrio/tests/test_path.py
jorisvandenbossche: We cannot control what the user passes, but that doesn't mean we have to accept everything the user passes, and so we _could_ not accept `Path("/vsimem/data.gpkg")`, because that is mixing two concepts (Python file path objects and GDAL VSI paths)

pyogrio/tests/test_geopandas_io.py
theroggy: OK, changed

pyogrio/_io.pyx
```diff
@@ -2112,7 +2125,7 @@ cdef create_ogr_dataset_layer(
                 raise exc
 
             # otherwise create from scratch
-            if is_vsi:
+            if use_tmp_vsimem:
                 VSIUnlink(path_c)
```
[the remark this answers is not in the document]
theroggy: I indeed don't think so... I'll remove it.

pyogrio/tests/test_path.py
theroggy: Ah, ok, now I understand: throw an error if a Path("/vsimem/") is passed in. Yes, that is an option as well I suppose, even though I personally viewed "/vsimem/..." conceptionally as a "path" myself, hence my confusion.

As it is about the same effort to support it I don't really see the advantage in doing that?

pyogrio/tests/test_core.py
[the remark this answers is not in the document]
theroggy: Discussed further in another comment... conclusion was to keep the original code without parametrize.

pyogrio/_vsi.pyx
theroggy: Further discussed in another comment...

pyogrio/tests/test_path.py
jorisvandenbossche: They are indeed also file paths, but not _OS_ file _system_ paths. You can indeed still use them for the nice concatenation functionality etc (like `Path(..) / ".."`), but even on Windows that essentially already breaks down when converting that naively into the string representation. So I would say it is a bit misuse of the pathlib library.

(in theory, pathlib also provides a `PurePath` that just does the path manipulation operations without any actual file system or IO operations, so that would be better fitting here. But only allowing that is probably only more confusing for users)

Now, if it is almost as much work to disallow as to allow, I won't argue strongly about disallowing them.

brendan-ward (approved):
Thanks @theroggy !  @jorisvandenbossche any final comments or disagreement with the plan to add the `vsi_*` functions at the top-level for now?

PR #459, opened 2024-08-10
MAINT: avoid tests to run if linting fails

theroggy:
I noticed the tests run even if the linting fails, which not needed.

This PR changes this so the lint workflow triggers both test workflows if the linting is ready.

### DIFF
--- a/.github/workflows/docker-gdal.yml
+++ b/.github/workflows/docker-gdal.yml
@@ -1,11 +1,6 @@
 name: Docker GDAL Test
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
-  workflow_dispatch:
+on: [workflow_call]  # will be called from lint workflow.
 
 # cancel running jobs on new commit to PR
 concurrency:
@@ -13,11 +8,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
-    uses: ./.github/workflows/lint.yml  # use the callable lint job to lint
-
   TestLinux:
-    needs: [lint]  # require linting to pass before tests run
     name: GDAL ${{ matrix.container }}
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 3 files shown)


brendan-ward (approved):
Thanks @theroggy

brendan-ward:
One potential downside of this change is that when workflows fail now, they refer to the linting workflow in the email subject, which is a bit misleading when they pass linting but fail on a downstream job.

theroggy:
Hmm... I don't get mails for failed actions... but that's indeed a bit misleading.

theroggy:
It seems under the "actions" tab, the real tests are now also under the "Linting" title: https://github.com/geopandas/pyogrio/actions/workflows/lint.yml.

theroggy:
Maybe rename it to "Lint and test"?

jorisvandenbossche:
> It seems under the "actions" tab, the real tests are now also under the "Linting" title: https://github.com/geopandas/pyogrio/actions/workflows/lint.yml.

Yeah, also the pop-up overview for a commit in a PR or history, the title now has "Linting", and also includes the "call-tests-conda":

image

While renaming "Linting" to "Lint and test" is certainly clearer, it also makes it even longer, which in this case would make that you no longer see the python version at all (actually in the above you already don't see it anymore)

This is not super important, but I would say it is a "nice to have" that the CI jobs have a clear name so you can easily see which one is failing (and the "linting" and "call-to.." are adding some noise)

theroggy:
I don't have a strong opinion on this either... and wasn't aware it would have these unwanted side-effects, so I wouldn't mind to revert to the old way of running them...

brendan-ward:
After seeing how this has worked out in practice, I think we should revert this.  Apologies for not catching the downsides earlier (it sounded good in theory!).

theroggy:
I agree. It indeed sounded good in theory :-)... I'll make a PR to revert it...

PR #460, opened 2024-08-18
Remove unnecessary build requirement

QuLogic:
I recall `wheel` being listed in instructions somewhere (maybe for `setuptools`), but it wasn't, and isn't, necessary.


QuLogic:
Ah, see the second note in this section: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use

brendan-ward (approved):
Thanks @QuLogic !

PR #461, opened 2024-08-20
CI: disable caching of conda env so most recent versions of dependencies are used

theroggy:
The caching results in older depencencies being used for long times, so incompatibilities with new versions not raised asap.

Reference https://github.com/geopandas/pyogrio/pull/452, where conda tests keep failing as the newest version of shapely, 2.0.6 , is not picked up due to the caching. In this PR shapely 2.0.6 is immediately picked up.


theroggy:
At least it is useful to know the option is available to cache for only one day...

PR #463, opened 2024-08-20
Exclude development files from wheels

QuLogic:
The Cython files are private (`_`-prefixed), so probably shouldn't be used by downstreams.


jorisvandenbossche (approved):
Yes, agreed that those files shouldn't be needed for the wheels. Thanks!

PR #465, opened 2024-08-27
MAINT: revert PR to avoid tests to run if linting fails

theroggy:
As discussed in #459, there are some unpractical side effects, so it is better to revert.


brendan-ward (approved):
Thanks for reverting, @theroggy

Issue #447, opened 2024-07-08
RLS: add LIBKML to GDAL vcpkg config for release


brendan-ward:
Per #444, we should probably include `LIBKML` in our VCPKG config, it is available as an [option](https://vcpkg.link/ports/gdal).

brendan-ward:
Closing this, as the decision was to leave LIBKML out of the release, in part due to challenges related to building it with VCPKG across the versions of manylinux that we support.  Added #472 as a replacement to this issue.

Issue #453, opened 2024-07-18
BLD/RLS: manylinux2014 at end of life


brendan-ward:
Per [manylinux EOL table](https://github.com/mayeut/pep600_compliance/blob/master/EOL.rst), manylinux2014 is at end of life.  We are currently using this for building x86_64 wheels, but using manylinux_2_28 for Arm64 wheels.

It appears the end of life date is set based on the Centos 7 EOL date.  The Docker image for manylinux2014 has been updated recently to resolve issues with Centos reaching EOL (fetching packages from Centos mirrors was failing), so it may continue to work for some time.

We have to use a custom install of zlib via VCPKG in manylinux2014 to get around compatibility issues (see #57), but don't have to do those same workarounds in manylinux_2_28.  This presents further issues as we try to add support for LIBKML in #589, which requires minizip, which depends on zlib - and probably requires a custom port of minizip for that to work.

It looks like manylinux_2_28 is the [recommended distro](https://github.com/mayeut/pep600_compliance?tab=readme-ov-file#acceptable-distros-to-build-wheels) for building wheels now, so we should plan to switch to this unless there is a reason not to.

jorisvandenbossche:
Some related discussion on the cibuildwheel issue tracker about changing their default manylinux image: https://github.com/pypa/cibuildwheel/issues/1772

jorisvandenbossche:
I am a little bit hesitant to already remove it, if that would mean around 20% of users would not longer be able to install the latest pyogrio (and a large majority of package is still providing manylinux2014)

Being able to clean-up the custom zlib port is certainly nice, but it once it was figured out, it also doesn't given much problems to keep a around a bit longer.

jorisvandenbossche:
> if that would mean around 20% of users would not longer be able to install the latest pyogrio

Although looking a bit more in detail at https://mayeut.github.io/manylinux-timeline/, it is also likely that a big part of this 20% is using Python 3.8 or older, and if we would be discussing that, I would say that it's time / fine to drop Python 3.8 support. 

When looking at readiness for glibc for non-EOL Python versions (which at this point still includes Python 3.8, but excludes older), the percentage of downloads that would not be ready for manylinux_2_28 drops to around 7-8% (and without Python 3.8 it will again be significantly lower)

One example distribution that will still require manylinux2014 mentioned in the linked issue above (https://github.com/pypa/cibuildwheel/issues/1772#issuecomment-2119982342) is AmazonLinux 2.

PR #466, opened 2024-08-29
BLD/RLS: Update wheels to include GDAL 3.9.2 and add manylinux_2_28 wheels

### LINKED ISSUE #454: BUG: wheel tests on macos-latest use the latest Arm64 version of pyogrio from PyPI
See logs from 0.9.0 release action [here](https://github.com/geopandas/pyogrio/actions/runs/9553600977/job/26334086179).  

The `macos-latest` runner is now MacOS 14 / Arm64 only, whereas previously it was x86_64.

Based on how release tests are setup, we are manually downloading the x86_64 wheel from the action artifacts to a `macos-latest` runner, but then installing that wheel fails because of architecture mismatch, so pip falls back to installing the latest Arm64 wheel from PyPI - which obviates the test altogether.

Instead, we need to have the `macos-latest` runner download the Arm64 wheel from the release artifacts instead and test with that.  We also should try to make sure that pip doesn't automatically & silently fall back to installing pyogrio from PyPI.

brendan-ward:
supersedes #451 , resolves #454

Uses GDAL 3.9.2 from VCPKG, which requires bumping this to the next minor release (0.10.0), ~and enables the libkml features (https://github.com/geopandas/pyogrio/issues/447)~. Also adds tests for GDAL 3.9.2 to CI.

This switches to pulling the Arm64 MacOS wheel when testing on the macos-latest runner, which is MacOS / Arm64 (resolves https://github.com/geopandas/pyogrio/issues/454) and uses --no-index flag to uv pip install to ensure it doesn't silently fall back to using an Arm64 wheel from PyPI.

This switches to uv for installing Python dependencies, in line with https://github.com/geopandas/pyogrio/pull/438, which also gets around issues with Python 3.12 requiring a virtual environment (needed for testing the sdist using the gdal:ubuntu-small-3.9.2 base image which has Python 3.12).

(unlike #451 this does not drop manylinux2014)


brendan-ward:
The failure on wheel building on Linux appears to be version conflicts with minizip (used by libkml) and zlib (installed via custom port).  Looks like we need to add a custom port here for minizip at the same version as zlib.

brendan-ward:
Unfortunately, libkml 1.3.0 [requires](https://github.com/libkml/libkml/blob/master/DEPENDENCIES) zlib / minizip >= 1.2.8.  1.3.0 was released in 2015, so already very old, and going for previous versions means not using cmake to build, which will likely be a pain to get working.

So - we have a couple of paths forward:
- drop attempt to include libkml in wheels and direct users to conda-forge to use the libkml driver (status quo); this means we can also drop the custom port of minizip
- drop use of manylinux2014, which enables us to stop using the custom ports of zlib and minizip and can successfully install libkml; however as mentioned in #453, this may mean no longer supporting older OS versions

I think we should go with the first option; it allows us to upgrade GDAL in wheels, but is otherwise no worse than the wheels we're currently building.

.github/workflows/release.yml
```diff
@@ -54,36 +54,60 @@ jobs:
...
+          uv pip install --no-cache --no-deps geopandas
+          uv pip list
 
       - name: Run tests
         shell: bash
+        # virtual environment is automatically activated
         run: |
           cd ..
-          python3 -m pytest --pyargs pyogrio.tests -v
+          python -c "import pyogrio; print(f'GDAL version: {pyogrio.__gdal_version__}\nGEOS version: {pyogrio.__gdal_geos_version__}')"
```
jorisvandenbossche, line 94: Not too important (this way is fine as well), but I _think_ the `uv run python -m pytest ..` should work here, because we are not in the root directory with a pyproject.toml file because of the `cd ..` (in contrast to the docker tests, I suppose)

.github/workflows/release.yml
```diff
@@ -248,15 +302,16 @@ jobs:
       - name: Install dependencies and pyogrio wheel
         shell: bash
         run: |
-          python -m pip install -r ci/requirements-wheel-test.txt
+          uv pip install --no-cache -r ci/requirements-wheel-test.txt
```
jorisvandenbossche, line 305: Is the `--no-cache` needed here?

jorisvandenbossche (commented):
Looking good!

.github/workflows/release.yml
brendan-ward: I think I was running into issues previously with using cached dependencies, but it might only be necessary when installing pyogrio itself.  Will test.

.github/workflows/release.yml
brendan-ward: This worked fine, but we still need to set the environment variables to automatically activate `uv`, so it doesn't obviate the need for those.

.github/workflows/release.yml
```diff
@@ -248,19 +302,20 @@ jobs:
       - name: Install dependencies and pyogrio wheel
         shell: bash
         run: |
-          python -m pip install -r ci/requirements-wheel-test.txt
+          uv pip install -r ci/requirements-wheel-test.txt
           if [ ${{ matrix.python-version }} != "3.12" ]; then
-            python -m pip install --no-deps geopandas
+            uv pip install geopandas
```
jorisvandenbossche, line 307: It might not be strictly needed, but I would keep the `--no-deps` here for geopandas, just to avoid we install a released pyogrio that is then overwritten below

.github/workflows/release.yml
jorisvandenbossche: Not that it matters too much, but just because I wanted to try to understand `uv` better (new to it), I _think_ what is going on:

- the reason we need to set the env variables (the reason uv does not find the venv automatically) is probably because of the `cd ..`. From the [`uv run` docs](https://docs.astral.sh/uv/reference/cli/#uv-run):

  > When used outside a project, if a virtual environment can be found in the current directory or a parent directory, the command will be run in that environment. Otherwise, the command will be run in the environment of the discovered interpreter.

  so because of the `cd ..` the venv directory is actually in a directory _down_, and not in the current or parent directory.

- We need the `cd ..` in this case to have the tests run properly on the installed package, but in case we didn't and wanted to run `uv` from the root without considering it as a project, that might work by putting a `[tool.uv] managed = false` in the pyproject.toml

jorisvandenbossche (commented):
One small last comment

.github/workflows/release.yml
jorisvandenbossche: Actually, it might definitely be needed, because from the logs it seems it then skips installing the local wheel because pyogrio is already installed? 

We could also swap the order to prevent this (first install the local wheel before installing geopandas, that's also the order we use for the sdist tests). But would also still use `--no-deps` to be sure.

brendan-ward: Whoops, yes, that was a mistake to remove `--no-deps`.  Switched order.

jorisvandenbossche (approved):

PR #352, opened 2024-01-26
BLD/RLS: fix aarch64 wheel building

jorisvandenbossche:

PR #468, opened 2024-09-05
BLD/RLS: fix aarch64 wheel building

jorisvandenbossche:
We have a patch to fix a vcpkg issue when compiling on a aarch64 system on linux (https://github.com/geopandas/pyogrio/pull/352). After updating the checked out vcpkg version in https://github.com/geopandas/pyogrio/pull/466/, the patch doesn't apply cleanly anymore. See https://app.circleci.com/pipelines/github/geopandas/pyogrio/415/workflows/c2e012fb-7cd9-47b2-b16b-9b0c0503607e/jobs/178

First testing if it now works out of the box.


jorisvandenbossche:
OK, that seems to work now!

brendan-ward:
Thanks @jorisvandenbossche !  I can't wait until arm64 runners are available on Github actions (looks like it will be [around the end of the year](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/)).

jorisvandenbossche:
Indeed!

Issue #378, opened 2024-03-30
To include spatialite in wheels or not to?


theroggy:
As mentioned here, https://github.com/geopandas/geopandas/pull/3223, think through wheter it it would be better to include it or not...

jorisvandenbossche:
Moving my comment from https://github.com/geopandas/pyogrio/issues/330#issuecomment-2028208074 here:

At this point the wheel build still doesn't include libspatialite support, i.e. the libspatialite feature is not enabled in the vcpkg GDAL build. 

Looking at it again, it seems we _could_ do that . @brendan-ward you commented about that in https://github.com/geopandas/pyogrio/issues/86#issuecomment-1109012589:

> Not sure if this matters, but noting for the record that our vcpkg build does not include `libspatialite` (we don't enable that as an add-on feature); it is LGPL licensed, which likely has license implications for the wheels we create.  I believe Fiona also does not build against `libspatialite`.

In https://github.com/geopandas/pyogrio/pull/161 I first added `libspatialite` feature, but then removed it again before the PR was merged, and I don't directly see any discussion or explanation that explains why I did that .. But there is a bit of discussion later at https://github.com/geopandas/pyogrio/pull/161#issuecomment-1269044332, which again mentions that fiona also doesn't have this.

Now, I don't think licensing is a problem, given that GEOS also uses LGPL, and we do already include that in the wheels.

Given that vcpkg has the feature to build GDAL with libspatialite, we should probably just try that.

PR #474, opened 2024-09-16
DOC: Add install note re: advanced drivers on conda-forge

### LINKED ISSUE #472: DOC: Add documentation to direct users to install pyogrio via conda-forge in order to access additional drivers (LIBKML, Spatialite)

brendan-ward:
Resolves #472, obviates #378.

It is potentially a bit confusing while we're still using `libgdal` on conda-forge, but I think these may need to be installed once we switch to `libgdal-core`.

### DIFF
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -24,6 +24,11 @@ This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
 raw I/O support and `geopandas` and their dependencies for GeoDataFrame
 I/O support.
 
+To enable additional GDAL divers, you may need to install them directly:
+
+-   LIBKML: [libkml](https://anaconda.org/conda-forge/libkml)
+-   Spatialite: [libspatialite](https://anaconda.org/conda-forge/libspatialite)
+
 ### PyPI
 
 This package is available on [PyPI](https://pypi.org/project/pyogrio/) for Linux,
@@ -39,6 +44,11 @@ If you get installation errors about Cython or GDAL not being available, this is
 most likely due to the installation process falling back to installing from the
 source distribution because the available wheels are not compatible with your
 platform.
+
+The binary wheels available on PyPI include the core GDAL drivers (GeoJSON,
+ESRI Shapefile, GPKG, FGB, OpenFileGDB, etc) but do not include more advanced
+drivers such as LIBKML and Spatialite. Use conda-forge to install pyogrio and
+then install these additional drivers directly.
 
 ### Troubleshooting installation errors
 
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -22,15 +22,15 @@ conda install -c conda-forge pyogrio
 
 This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
 raw I/O support and `geopandas` and their dependencies for GeoDataFrame
-I/O support.
+I/O support. By default, the `GDAL` package on conda-forge already supports a
+wide range of vector formats. If needed, you can install additional drivers by
+installing the associated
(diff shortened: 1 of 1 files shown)


docs/source/install.md
```diff
@@ -24,6 +24,11 @@ This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
 raw I/O support and `geopandas` and their dependencies for GeoDataFrame
 I/O support.
 
+To enable additional GDAL divers, you may need to install them directly:
+
+-   LIBKML: [libkml](https://anaconda.org/conda-forge/libkml)
```
jorisvandenbossche, line 29: Based on the list of available plugins as conda packages (https://gdal.org/en/latest/download.html#conda) I would assume that KML is still built-in with libgdal-core?

jorisvandenbossche: And the same for spatialite I assume

jorisvandenbossche: It would still be good to mention this (the fact that some drivers need an additional install), but looking at that list, the only three vector drivers that get installed separately as plugin are `libgdal-arrow-parquet`, `libgdal-pg` and `libgdal-xls`, I think

docs/source/install.md
```diff
@@ -40,6 +48,11 @@ most likely due to the installation process falling back to installing from the
 source distribution because the available wheels are not compatible with your
 platform.
 
+The binary wheels available on PyPI include the core GDAL drivers (GeoJSON,
+ESRI Shapefile, GPKG, FGB, OpenFileGDB, etc) but do not include more advanced
+drivers such as LIBKML and Spatialite. Use conda-forge to install pyogrio and
+then install these additional drivers directly if needed.
```
theroggy, line 54: I would still understand from this that LIBKML and spatialite would need installing extra things when using conda-forge?

brendan-ward: Should we include a "see above" to make that a bit more clear?  I was trying to get at that with the "if needed", but maybe it is still confusing as written?

theroggy: Maybe something like this?

```suggestion
drivers such as LIBKML and Spatialite. If you need such drivers, it is recommended to use conda-forge to install pyogrio as explained above.
```

docs/source/install.md
```diff
@@ -24,6 +24,14 @@ This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
 raw I/O support and `geopandas` and their dependencies for GeoDataFrame
 I/O support.
 
+By default, the `GDAL` package on conda-forge currently includes LIBKML and
+Spatialite. However, you may need to install additional drivers by installing
+the associated [conda-forge package](https://gdal.org/en/latest/download.html#conda):
```
theroggy, line 29:
```suggestion
By default, the `GDAL` package on conda-forge already supports a wide range of vector formats. If needed, you can install additional drivers by installing
the associated [conda-forge package](https://gdal.org/en/latest/download.html#conda). Currently, the following packages to install extra vector drivers are available:
```

docs/source/install.md
```diff
@@ -24,6 +24,14 @@ This requires compatible versions of `GDAL` and `numpy` from `conda-forge` for
...
+
+-   `libgdal-arrow-parquet`
+-   `libgdal-pg`
+-   `libgdal-xls`
```
jorisvandenbossche, line 33:
```suggestion
-   `libgdal-arrow-parquet` ((Geo)Parquet and (Geo)Arrow IPC)
-   `libgdal-pg` (PostgreSQL / PostGIS)
-   `libgdal-xls` (XLS - MS Excel format)
```

(for `pg` I think it is certainly valuable to add the full name for clarity, the others might be clear already)

theroggy (approved):

PR #477, opened 2024-09-26
DOC: Add docs for working with in-memory datasets

brendan-ward:
Depends on #457

Adds examples using the automatically-managed in-memory dataset as well as externally managed `/vsimem/` datasets.

### DIFF
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -590,9 +590,6 @@ gdal.Rasterize("test.tif", '/vsimem/test.gpkg', outputType=gdal.GDT_Byte, noData
 from pyogrio import vsi_unlink
 
 vsi_unlink('/vsimem/test.gpkg')
-
-# or use gdal to release it
-gdal.Unlink('/vsimem/test.gpkg')
 ```
 
 Pyogrio can also read from a valid `/vsimem/` file created using a different
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -533,9 +533,6 @@ You can also use a dictionary to specify either `dataset_options` or
 
 ## Reading from and writing to in-memory datasets
 
-Pyogrio can automatically create and manage in-memory virtual files (`/vsimem`)
-to enable reading and writing from in-memory datasets.
-
 It is possible to read from a dataset stored as bytes:
 
 ```python
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -581,15 +581,15 @@ packages that support the `/vsimem/` interface, such as the `gdal` package:
 ```python
 from osgeo import gdal
 
-write_dataframe(df, '/vsimem/test.gpkg', layer="my_layer", driver="GPKG")
+write_dataframe(df, "/vsimem/test.gpkg", layer="my_layer", driver="GPKG")
 
 # perform some operation using it
-gdal.Rasterize("test.tif", '/vsimem/test.gpkg', outputType=gdal.GDT_Byte, noData=255, initValues=255, xRes=0.1, yRes=-0.1, allTouched=True, burnValues=1)
+gdal.Rasterize("test.tif", "/vsimem/test.gpkg", outputType=gdal.GDT_Byte, noData=255, initValues=255, xRes=0.1, yRes=-0.1, allTouched=True, burnValues=1)
 
 # release the memory using pyogrio
 from pyogrio import vsi_unlink
 
-vsi_unlink('/vsimem/test.gpkg')
+vsi_unlink("/vsimem/test.gpkg")
 ```
 
 Pyogrio can also read from a valid `/vsimem/` file created using a different
--- a/docs/source/introduction.md
+++ b/docs/source/introduction.md
@@ -531,6 +531,79 @@ You can also use a dictionary to specify either `dataset_options` or
 >>> write_dataframe(df, '/tmp/test.gpkg', dataset_options={"version": "1.0"}, layer_options={"geometry_name": "the_geom"})
 ```
 
+## Reading from and writing to in-memory datasets
+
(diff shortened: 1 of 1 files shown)


docs/source/introduction.md
```diff
@@ -531,6 +531,79 @@ You can also use a dictionary to specify either `dataset_options` or
...
+
+geojson_bytes = BytesIO(geojson.encode("UTF-8"))
+
+df = read_dataframe(geojson_bytes)
+```
+
+Note: this may emit a `RuntimeWarning` where the in-memory dataset is detected
+to be a particular format but lacks the expected file extension (e.g., `.gpkg`)
+because the in-memory path automatically created by pyogrio does not include the
+extension.
```
theroggy, line 564: Is there a specific reason why the file extension is not added to the filename (to avoid this warning)?

docs/source/introduction.md
```diff
@@ -531,6 +531,79 @@ You can also use a dictionary to specify either `dataset_options` or
 >>> write_dataframe(df, '/tmp/test.gpkg', dataset_options={"version": "1.0"}, layer_options={"geometry_name": "the_geom"})
 ```
 
+## Reading from and writing to in-memory datasets
+
+Pyogrio can automatically create and manage in-memory virtual files (`/vsimem`)
+to enable reading and writing from in-memory datasets.
```
theroggy, line 537: Is it relevant that "under the hood" /vsimem is being used to read/write from/to byte arrays? To me this information is rather confusing?

docs/source/introduction.md
```diff
@@ -531,6 +531,79 @@ You can also use a dictionary to specify either `dataset_options` or
...
+write_dataframe(df, buffer, layer="my_layer", driver="GPKG")
+
+out_bytes = buffer.getvalue()
+```
+
+Note: this is limited to single-file data formats (e.g., GPKG) and does not
+support formats with multiple files (e.g., ESRI Shapefile).
+
+It is also possible to use a `/vsimem/` in-memory dataset with other GDAL-based
+packages that support the `/vsimem/` interface, such as the `gdal` package:
```
theroggy, line 582: I think it can also be useful to just use /vsimem without using other gdal-based packages... So would start with that (simpler) scenario... and then expand to combining with other gdal based packages?

theroggy (commented):
Thanks! Some comments...

docs/source/introduction.md
brendan-ward: because we get only bytes, not a path - so we'd have to inspect the contents of the bytes to try to deduce the file type.  Since GDAL already has the logic in place for that, it seemed better to just leave that to GDAL and not try to look for file type identifiers in the first few bytes and potentially get it wrong.

docs/source/introduction.md
```diff
@@ -531,6 +531,79 @@ You can also use a dictionary to specify either `dataset_options` or
...
+# perform some operation using it
+gdal.Rasterize("test.tif", '/vsimem/test.gpkg', outputType=gdal.GDT_Byte, noData=255, initValues=255, xRes=0.1, yRes=-0.1, allTouched=True, burnValues=1)
+
+# release the memory using pyogrio
+from pyogrio import vsi_unlink
+
+vsi_unlink('/vsimem/test.gpkg')
+
+# or use gdal to release it
+gdal.Unlink('/vsimem/test.gpkg')
```
theroggy, line 595: Not sure if there is any added value to mention this here? Most other gdal-based packages don't have this functionality anyway, and the `gdal` python bindings are not the most typical/popular combo with `pyogrio` I think?

docs/source/introduction.md
theroggy: Hmm... true.

docs/source/introduction.md
brendan-ward: Removed

docs/source/introduction.md
```diff
@@ -531,6 +531,73 @@ You can also use a dictionary to specify either `dataset_options` or
...
+Note: this is limited to single-file data formats (e.g., GPKG) and does not
+support formats with multiple files (e.g., ESRI Shapefile).
+
+It is also possible to use a `/vsimem/` in-memory dataset with other GDAL-based
+packages that support the `/vsimem/` interface, such as the `gdal` package:
+
+```python
+from osgeo import gdal
+
+write_dataframe(df, '/vsimem/test.gpkg', layer="my_layer", driver="GPKG")
```
theroggy, line 584:
```suggestion
write_dataframe(df, "/vsimem/test.gpkg", layer="my_layer", driver="GPKG")
```

docs/source/introduction.md
```diff
@@ -531,6 +531,73 @@ You can also use a dictionary to specify either `dataset_options` or
...
+
+# perform some operation using it
+gdal.Rasterize("test.tif", '/vsimem/test.gpkg', outputType=gdal.GDT_Byte, noData=255, initValues=255, xRes=0.1, yRes=-0.1, allTouched=True, burnValues=1)
```
theroggy, line 587:
```suggestion
gdal.Rasterize("test.tif", "/vsimem/test.gpkg", outputType=gdal.GDT_Byte, noData=255, initValues=255, xRes=0.1, yRes=-0.1, allTouched=True, burnValues=1)
```

docs/source/introduction.md
```diff
@@ -531,6 +531,73 @@ You can also use a dictionary to specify either `dataset_options` or
...
+
+# release the memory using pyogrio
+from pyogrio import vsi_unlink
+
+vsi_unlink('/vsimem/test.gpkg')
```
theroggy, line 592:
```suggestion
vsi_unlink("/vsimem/test.gpkg")
```

theroggy (approved):
Stupid detail I didn't notice before: some strings are defined with single quotes instead of double quotes.

Issue #478, opened 2024-09-27
GEOS 3.13 is causing curve geometry tests to segfault


brendan-ward:
First observed in the conda tests, which are pulling in GEOS 3.13 which [added support for curve geometry types](https://github.com/libgeos/geos/blob/main/NEWS.md#changes-in-3130).  I can reproduce the segfault locally when installing deps from conda-forge and building latest pyogrio.

It looks like the latest GDAL Docker versions we're testing against still use GEOS 3.12, so we aren't yet seeing this issue in the other CI tests - but it may be lurking around the corner.  Until there is a version of GDAL built against GEOS 3.13 it is harder to test against this configuration outside of the conda packages.

brendan-ward:
Looks like homebrew GDAL 3.9.2 includes GEOS, so I should be able to compile against that locally and see if this segfaults when using GDAL outside of conda-forge.

brendan-ward:
Narrowed it down a bit: this happens where `use_arrow=True` when reading certain curve geometries available in our test suite:
- `CurvePolygon`
- `MultiSurface`

(we don't have `Surface` at the moment)

`Curve` doesn't segfault but raises another error.

I can also reproduce the issue locally when compiling against homebrew GDAL and compiling shapely against GEOS 3.13, which makes this easier to investigate.

When we are reading geometries using without using Arrow, we can detect if a geometry is nonlinear and have GDAL force it to a linear type before we read it.  However, when using Arrow, we read the raw WKB with GDAL, and count on shapely to be able to parse it.  The segfault is occurring in `shapely.from_wkb` with these nonlinear types.

However, GEOS 3.13 appears to work fine:

```bash
❯ geosop -a "010C00000001000000010A00000001000000010200000004000000D2936CD801D357C01F7CE5EED1AE4F402EDD365958A657C065B26029F3884A4056D
1DE08466D52C0ED7574EB21134E40D2936CD801D357C01F7CE5EED1AE4F40"
MULTISURFACE (CURVEPOLYGON ((-95.29698763468994 63.36578165250125, -94.5991423640996 53.06992070409323, -73.7073995758013 60.149472648492385, -95.29698763468994 63.36578165250125)))
```

theroggy:
Shapely with GEOS 3.13 also seems to crash on the wkt:

``` python
import shapely

[...]

PR #479, opened 2024-09-28
TST: skip multisurface test with arrow to avoid crash

theroggy:
I wonder if there is any use in delaying the release for this. It will have to be solved upstream... and pyogrio 0.9 with GEOS 3.13 has the same problem...

So, just skipping the test might be OK for now?

Reference: https://github.com/geopandas/pyogrio/issues/478

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -1765,6 +1765,10 @@ def test_read_invalid_poly_ring(tmp_path, use_arrow, on_invalid, message):
 
 def test_read_multisurface(multisurface_file, use_arrow):
     if use_arrow:
+        # TODO: revisit once https://github.com/geopandas/pyogrio/issues/478
+        # is resolved.
+        pytest.skip("Shapely + GEOS 3.13 crashes in from_wkb for this case")
+
         with pytest.raises(shapely.errors.GEOSException):
             # TODO(Arrow)
             # shapely fails parsing the WKB


brendan-ward:
Thanks for working on this @theroggy .  I'm thinking we need a deeper fix here than skipping the tests, since occasionally users do try to work with nonlinear types and we are trying to make use of the Arrow interface more common.

I'm thinking we might be able to use [OGRSetNonLinearGeometriesEnabledFlag](https://gdal.org/en/latest/api/vector_c_api.html#_CPPv436OGRSetNonLinearGeometriesEnabledFlagi) when reading using Arrow to have a similar effect as what we are already doing when not using Arrow, and then we sidestep the issue in Shapely.  Let me try a PR with that.

brendan-ward:
Nevermind, `OGRSetNonLinearGeometriesEnabledFlag` has no effect when reading from GDAL using Arrow, likely by design.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -1765,6 +1765,10 @@ def test_read_invalid_poly_ring(tmp_path, use_arrow, on_invalid, message):
 
 def test_read_multisurface(multisurface_file, use_arrow):
     if use_arrow:
+        # TODO: revisit once https://github.com/geopandas/pyogrio/issues/478
+        # is resolved.
+        pytest.skip("Shapely + GEOS 3.13 crashes in from_wkb for this case")
```
brendan-ward, line 1770: Can you move this to an annotation instead of inline in the test function?

brendan-ward (approved):
I'm not seeing any way via a GDAL config option that would let us force geometries to linear types when reading from Arrow, so unfortunately this can't really be fixed until handled upstream in Shapely.

Until then skipping this test keeps us green on CI.

pyogrio/tests/test_geopandas_io.py
theroggy: I'm not sure how to use a fixture in an annotation, and I'm not sure if it is worth the effort to try to get this working for a temporary hack that will only be in use for a few days or weeks at max?

If you know how, feel free...

brendan-ward: You're right, I wasn't looking at `use_arrow` when I said that.  All good as is.

PR #480, opened 2024-09-28
CI: Try to resolve mamba install issue on windows when using with pip

brendan-ward:
The windows conda-forge test runner fails for the `libgdal3.5.1` env when using `pip` to install shapely.  It looks like an issue with where the mamba executables are installed.

### DIFF
--- a/.github/workflows/tests-conda.yml
+++ b/.github/workflows/tests-conda.yml
@@ -49,10 +49,21 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


brendan-ward:
No idea why the mamba install is failing on Windows when we use pip, first attempts didn't fix it.  However, we're really only using the `libgdal3.5.1` env to proof that building against GDAL with files stored in a particular location works properly; we don't need geopandas in this case.  So I limited this to the minimal dependencies and now it runs.

theroggy (approved):

PR #481, opened 2024-10-01
DOC: update pyogrio introduction

theroggy:
The pyogrio introduction was slightly dated in some locations, so I started off with doing some minor updates.

However, I got a bit carried away with making small changes here and there... I also changed the order of some blocks of text, which doesn't help to see the exact changes applied :-(.

For the "vectorized", not sure why this term was used, but I don't think it is really relevant for what happens in pyogrio, or possibly I'm just wrong on that?

Anyway, if it's not an improvement, just say so, a PR is easy to delete again ;-) (or rather, to close without merging).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # pyogrio - bulk-oriented spatial vector file I/O using GDAL/OGR
 
 Pyogrio provides fast, bulk-oriented read and write access to 
-[GDAL/OGR vector data sources](https://gdal.org/en/latest/drivers/vector/index.html),
-such as ESRI Shapefile, GeoPackage, GeoJSON,... Vector data sources typically
-have geometries, such as points, lines, or polygons, and associated records
-with potentially many columns worth of data.
+[GDAL/OGR](https://gdal.org/en/latest/drivers/vector/index.html) vector data
+sources, such as ESRI Shapefile, GeoPackage, GeoJSON,... Vector data sources
+typically have geometries, such as points, lines, or polygons, and associated
+records with potentially many columns worth of data.
 
 The typical use is to read or write these data sources to/from
 [GeoPandas](https://github.com/geopandas/geopandas) `GeoDataFrames`. Because
--- a/README.md
+++ b/README.md
@@ -1,34 +1,28 @@
-# pyogrio - Vectorized spatial vector file format I/O using GDAL/OGR
+# pyogrio - bulk-oriented spatial vector file I/O using GDAL/OGR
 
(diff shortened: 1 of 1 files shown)


brendan-ward:
Re: "vectorized" - that was originally added to distinguish it from Fiona, which one could describe as having a more iterative API: you loop over individual features in the Python API.  Whereas in pyogrio, the looping happens in compiled code, so it presents results as arrays (vectors) of features.  But it was a shorthand reference and not particularly clear, so it is fine for that to go away.

README.md
```diff
@@ -1,35 +1,29 @@
...
+Pyogrio provides fast, bulk-oriented read and write access to 
+[GDAL/OGR](https://gdal.org/en/latest/drivers/vector/index.html) vector data
+sources, such as ESRI Shapefile, GeoPackage, GeoJSON,... Vector data sources
+typically have geometries, such as points, lines, or polygons, and associated
+records with potentially many columns worth of data.
+
+The typical use is to read or write these data sources to/from
+[GeoPandas](https://github.com/geopandas/geopandas) `GeoDataFrames`. Because
+the geometry column is optional, reading or writing only non-spatial data is
+also possible. Hence, GeoPackage attribute tables, DBF files or CSV files are
```
brendan-ward, line 12:
```suggestion
also possible. Hence, GeoPackage attribute tables, DBF files, or CSV files are
```

README.md
```diff
@@ -1,35 +1,29 @@
...
-Pyogrio uses a vectorized approach for reading and writing GeoDataFrames to and
-from OGR vector data sources in order to give you faster interoperability. It
-uses pre-compiled bindings for GDAL/OGR so that the performance is primarily
-limited by the underlying I/O speed of data source drivers in GDAL/OGR rather
-than multiple steps of converting to and from Python data types within Python.
+# pyogrio - bulk-oriented spatial vector file I/O using GDAL/OGR
+
+Pyogrio provides fast, bulk-oriented read and write access to 
+[GDAL/OGR](https://gdal.org/en/latest/drivers/vector/index.html) vector data
+sources, such as ESRI Shapefile, GeoPackage, GeoJSON,... Vector data sources
```
brendan-ward, line 5:
```suggestion
sources, such as ESRI Shapefile, GeoPackage, GeoJSON, and several others. Vector data sources
```

brendan-ward (commented):
Thanks @theroggy - this is looking better!  A couple minor comments.

brendan-ward (approved):

PR #486, opened 2024-10-17
ENH: deal properly with naive datetimes with arrow

theroggy:
When arrow is used, naive datetimes are interpreted as being in UTC. Is this logical? Isn't it logical to treat them as local dates?

This PR at the moment adds a test showing this behaviour (and failing on it).

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -1,5 +1,6 @@
 import contextlib
 import locale
+import time
 import warnings
 from datetime import datetime
 from io import BytesIO
@@ -348,6 +349,29 @@ def test_read_write_datetime_tz_with_nulls(tmp_path, use_arrow):
     if use_arrow:
         # with Arrow, the datetimes are always read as UTC
         df["dates"] = df["dates"].dt.tz_convert("UTC")
+    assert_geodataframe_equal(df, result)
+
+
+@pytest.mark.filterwarnings(
+    "ignore: Non-conformant content for record 1 in column dates"
+)
+@pytest.mark.requires_arrow_write_api
+def test_read_write_datetime_no_tz(tmp_path, use_arrow):
+    dates_raw = ["2020-01-01 09:00:00.123", "2020-01-01 10:00:00"]
+    if PANDAS_GE_20:
+        dates = pd.to_datetime(dates_raw, format="ISO8601").as_unit("ms")
+    else:
+        dates = pd.to_datetime(dates_raw)
+    df = gp.GeoDataFrame(
+        {"dates": dates, "geometry": [Point(1, 1), Point(1, 1)]},
+        crs="EPSG:4326",
+    )
+    fpath = tmp_path / "test.gpkg"
+    write_dataframe(df, fpath, use_arrow=use_arrow)
+    result = read_dataframe(fpath, use_arrow=use_arrow)
+    if use_arrow:
+        # with Arrow, the datetimes are always read as UTC
+        df["dates"] = df["dates"].dt.tz_localize(time.timezone).dt.tz_convert("UTC")
     assert_geodataframe_equal(df, result)
 
 


pyogrio/tests/test_geopandas_io.py
```diff
@@ -328,11 +330,38 @@ def test_write_datetime_mixed_offset(tmp_path, use_arrow):
...
+        crs="EPSG:4326",
+    )
+    fpath = tmp_path / f"test{ext}"
+    write_dataframe(df, fpath, use_arrow=use_arrow)
+    result = read_dataframe(fpath, use_arrow=use_arrow)
+    if use_arrow and ext == ".gpkg":
+        # for GPKG with Arrow, the datetime is written as naive datetime with the
+        # correct times, but when read the naive time is assumed to be UTC, which
+        # changes the effective time so this seems wrong.
+        df["dates"] = df["dates"].dt.tz_localize(time.timezone).dt.tz_convert("UTC")
```
jorisvandenbossche, line 355:
```suggestion
        df["dates"] = df["dates"].dt.tz_localize("UTC")
```

I would expect this to be just interpreted as UTC (not depend on the locale timezone) ?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -328,11 +330,38 @@ def test_write_datetime_mixed_offset(tmp_path, use_arrow):
     assert_series_equal(result["dates"], utc_col)
 
 
+@pytest.mark.parametrize("ext", [ext for ext in ALL_EXTS if ext != ".shp"])
 @pytest.mark.filterwarnings(
     "ignore: Non-conformant content for record 1 in column dates"
 )
```
jorisvandenbossche, line 336: For the non-tz case, this filterwarnings might not be needed? 
(don't remember exactly what was the reason for the non-conformity, but potentially the `-05:00` was causing that)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -342,12 +371,31 @@ def test_read_write_datetime_tz_with_nulls(tmp_path, use_arrow):
...
-    fpath = tmp_path / "test.gpkg"
+    fpath = tmp_path / f"test{ext}"
     write_dataframe(df, fpath, use_arrow=use_arrow)
     result = read_dataframe(fpath, use_arrow=use_arrow)
     if use_arrow:
-        # with Arrow, the datetimes are always read as UTC
-        df["dates"] = df["dates"].dt.tz_convert("UTC")
+        if ext == ".fgb":
+            # when FlatGeoBuffer is read with Arrow, for datetimes with equal timezone,
+            # a column type with the appropriate minutes offset is returned.
```
jorisvandenbossche, line 380: I don't really understand this line? (or this case is not actually being tested?)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -342,12 +371,31 @@ def test_read_write_datetime_tz_with_nulls(tmp_path, use_arrow):
...
+            # REMARK: For .fgb, the timezone is just dropped when reading or writing!!!
+            #     -> 2020-01-01T09:00:00.123-05:00 becomes 2020-01-01T09:00:00.123
+            df["dates"] = df["dates"].dt.tz_localize(tz=None)
+        elif ext in (".geojson", ".geojsonl"):
+            # when GeoJSON is read with Arrow, for datetimes with equal timezone, a
+            # column type with the appropriate minutes offset is returned.
+            # REMARK: for .geojson, the data is written fine, but when reading it goes
+            # wrong: 2020-01-01T09:00:00.123-05:00 becomes 2020-01-01T04:00:00.123-05:00
```
jorisvandenbossche, line 388: I think this was a bug that is fixed in the meantime? (so will have to update the expected result for that depending on the GDAL version)

jorisvandenbossche (commented):
Thanks for diving into this and improving the test coverage!

pyogrio/tests/test_geopandas_io.py
jorisvandenbossche: (indeed the offset: https://github.com/OSGeo/gdal/issues/8460#issuecomment-1732382430)

theroggy: Has indeed become redundant... removed.

pyogrio/tests/test_geopandas_io.py
theroggy: This was "temporary" testing code to show the old (not ideal) behaviour. This PR now includes changes to improve the handling of dates (if GDAL >= 3.11), so code has been removed now.

pyogrio/tests/test_geopandas_io.py
theroggy: This was just code to show/explain what was going wrong, so can be removed now.

pyogrio/tests/test_geopandas_io.py
theroggy: Indeed.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -297,7 +297,10 @@ def test_read_datetime(datetime_file, use_arrow):
 
 @pytest.mark.filterwarnings("ignore: Non-conformant content for record 1 in column ")
 @pytest.mark.requires_arrow_write_api
-def test_read_datetime_tz(datetime_tz_file, tmp_path, use_arrow):
+def test_write_datetime_tz(datetime_tz_file, tmp_path, use_arrow):
+    if use_arrow and __gdal_version__ < (3, 11, 0):
+        pytest.skip("Arrow datetime handling improved in GDAL >= 3.11")
```
jorisvandenbossche, line 302: Can we still test what behaviour you get with the current GDAL?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -340,32 +342,104 @@ def test_write_datetime_mixed_offset(tmp_path, use_arrow):
...
+
+    assert_series_equal(result["dates"], expected)
+
+
+@pytest.mark.parametrize("ext", [ext for ext in ALL_EXTS if ext != ".shp"])
+@pytest.mark.requires_arrow_write_api
+def test_write_datetime_no_tz(tmp_path, ext, use_arrow):
+    """Test writing/reading a datetime column without timezone information."""
+    if use_arrow and ext == ".gpkg" and __gdal_version__ < (3, 11, 0):
+        pytest.skip("Arrow datetime handling improved in GDAL >= 3.11")
```
jorisvandenbossche, line 383: What is not yet working for the case with no tz for GPKG?

jorisvandenbossche (commented):
@theroggy thanks for further looking into this!

I do have some doubts about how much effort we should do to cover corner cases and what the desired default behaviour should be, see my comments below.

pyogrio/geopandas.py
```diff
@@ -51,12 +53,6 @@ def _try_parse_datetime(ser):
             res = pd.to_datetime(ser, **datetime_kwargs)
         except Exception:
             res = ser
-    # if object dtype, try parse as utc instead
-    if res.dtype == "object":
-        try:
-            res = pd.to_datetime(ser, utc=True, **datetime_kwargs)
-        except Exception:
-            pass
```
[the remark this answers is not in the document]
theroggy: > From your top post explanation:
> 
> > Add support to read mixed timezone datetimes. These are returned in an object column with Timestamps.
> 
> First, I don't think this will work with upcoming pandas 3.x (we are suppressing the warning above about mixed timezones going to raise unless passing `utc=True`, and that you have to use `apply` and `datetime.datetime.strptime` instead to get mixed offset objects) (but the tests are also passing, so maybe I am missing something)

Yes, I saw, I also thought that that would have to be fixed for pandas 3. 

> Second, a column of mixed offset objects is in general not that particularly useful .. So changing this behaviour feels like a regression to me. I understand that we might want to provide the user the option to get this, but by default, I am not sure.

For starters, this is only relevant for mixed timezone data, data saved in naive or UTC timestamps should just stay "regular" pandas datetime columns.

For the case of mixed timezone data, it depends on what you want to do with the datetime data. If it is just to look at it/show/keep it as it is part of the table data, the Timestamps look just fine to me. If you really want to do "fancy stuff" with the datetimes it will in pandas indeed be more convenient for some things to transform them into e.g. UTC datetimes to get a datetime column instead of an object column.

Regarding default behaviour, it feels quite odd to me to transform data by default to a form where information (the original time zone) is lost. Also because when you save the data again, it will then be saved as UTC as well, so also: the timezone information will be lost.

To me, the other way around is more logical: by default you don't loose data. If you want to do "fancy stuff" with a datetime column that contains mixed timezone data, you convert it to e.g. UTC, typically in an extra column, because most likely you will want to keep the timezone information again when saving.

pyogrio/tests/test_geopandas_io.py
theroggy: I replaced all skip's by a test on the result of GDAL <3.11

pyogrio/geopandas.py
```diff
@@ -620,8 +644,35 @@ def write_dataframe(
...
+                ):
+                    df[name] = col.apply(
+                        lambda x: None if pd.isna(x) else x.isoformat()
+                    )
+                    datetime_cols.append(name)
+            elif isinstance(dtype, pd.DatetimeTZDtype):
+                # Also for regular datetime columns with timezone mixed timezones are
+                # possible when thera is a difference between summer and winter time.
+                df[name] = col.apply(lambda x: None if pd.isna(x) else x.isoformat())
+                datetime_cols.append(name)
```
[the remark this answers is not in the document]
theroggy: > Is this needed for properly typed datetime64 columns? What does GDAL do with those values? Write as the UTC value?

It is actually quite weird. Base on what I tested, as long as the column only contains one single value as offset, the offset is written correctly to the file. Once there are more than 1 different offsets, the offset is simply dropped, so then naive times are written. This is the case for both GPKG and e.g. .geojson. But, note that if a timezone uses daylight saving time (difference between summer and winter time)... datetimes from different dates will have another offset...

Hence, to get correct results, in almost all cases where a column has a timezone the only path to have correct time zone information written is to convert to string. An exception is UTC... as it doesn't apply daylight saving changes, so I just made that exception.

> And with this change it will write it, still as a datetime (because of adding the metadata?), but with offset?

With this change offsets will be correctly written as timestamp, indeed because of the custom arrow metadata being added and interpreted by GDAL from GDAL 3.11 (https://github.com/OSGeo/gdal/pull/11213) 

> FWIW, if we need to do this, you can do `df[name].astype(str)` to avoid the `apply`

Interesting! Note however that `df[name].astype(str)` outputs "... ..." instead of "...T...", so no strictly valid ISO... strings. I think GDAL will interprete it correctly as well, but for GDAL version < 3.11 this leads to the strings being written in not-strictly correct ISO strings. Not sure how important that is...

pyogrio/geopandas.py
```diff
@@ -496,8 +500,28 @@ def write_dataframe(
             # Convert each row offset to a signed multiple of 15m and add to GMT value
             gdal_offset_representation = tz_offset // pd.Timedelta("15m") + 100
             gdal_tz_offsets[name] = gdal_offset_representation.values
-        else:
+
+        elif col.dtype == "object":
+            # Column of Timestamp objects, also split in naive datetime and tz offset
+            col_na = df[col.notna()][name]
+            if len(col_na) and all(isinstance(x, pd.Timestamp) for x in col_na):
```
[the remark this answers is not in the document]
theroggy: If they are written as strings to the output files without the proper metadata, it depends on format to format if they will be recognized as datetimes when read. For text files they will typically be recognized as datetime as the data types are "guessed" when the file is read (e.g. geojson), for files like .fgb and .gpkg they won't be recognized as the file metadata will be wrong.

That's not very clean, and as it is very easy to solve I don't quite see the point of not supporting it properly?

jorisvandenbossche:
@theroggy thanks a lot for the updates here! The implementation is generally looking good, I mostly have some remaining concerns about the exact user-facing API. 

- `"UTC"` for the default seems a bit misleading, because it is only UTC for tz-aware columns. If you have a tz-naive column, that roundtrips as tz-naive. Also for tz-aware columns, if there is only one offset, you will get the fixed offset, not UTC (unless that behaviour changed with this PR?)
- `"DATETIME"`: the important aspect here is that it are datetime/timestamp _objects_, because "datetime" could also mean the dtype, and then it could be interpreted as "use a datetime dtype" (which is datetime64). Also the current PR returns `pd.Timestamp` and not `datetime.datetime`, so that is also a bit confusing (although we should maybe return datetime.datetime?)

We already have the `datetime_as_string=True/False` keyword (in the raw IO, just not exposed in the dataframe IO), so wondering if we could also expose that in `read_dataframe` and then add a `datetime_as_object=True/False` keyword. 
Of course, two separate boolean keywords that then cannot both be True at the same time because they essentially control the same choice is also not great design API, but from a user point of view, for writing code I think I would find that easier to read in my code (`read_dataframe(..., datetime_as_string=True)` vs `read_dataframe(..., datetimes="string")`)

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 0.12.0 (????-??-??)
+
+### Improvements
+
+-   Add `datetimes` parameter to `read_dateframe` to choose the way datetime columns are
```
jorisvandenbossche, line 7:
```suggestion
-   Add `datetimes` parameter to `read_dataframe` to choose the way datetime columns are
```

pyogrio/geopandas.py
```diff
@@ -267,11 +308,10 @@ def read_dataframe(
...
     gdal_force_2d = False if use_arrow else force_2d
-    if not use_arrow:
-        # For arrow, datetimes are read as is.
-        # For numpy IO, datetimes are read as string values to preserve timezone info
-        # as numpy does not directly support timezones.
-        kwargs["datetime_as_string"] = True
+
+    # Always read datetimes as string values to preserve (mixed) timezone info
+    # as numpy does not directly support timezones and arrow datetime columns
+    # don't support mixed timezones.
```
jorisvandenbossche, line 314: Do you know what the performance impact is of also setting `datetime_as_string` to True for the Arrow code path? 
I would think that this is going to be quite a bit slower, and in that case we should maybe only set it if the user asked for getting datetime objects or strings, and not in the default case?

pyogrio/geopandas.py
```diff
@@ -37,33 +39,56 @@ def _stringify_path(path):
...
-def _try_parse_datetime(ser):
+def _try_parse_datetime(ser, datetimes="UTC"):
     import pandas as pd  # only called when pandas is known to be installed
+    from pandas.api.types import is_string_dtype
+
+    datetimes = datetimes.upper()
+    datetime_kwargs = {}
+    if datetimes == "STRING":
+        if not is_string_dtype(ser.dtype):
+            res = ser.astype("string").str.replace(" ", "T")
```
jorisvandenbossche, line 50: When is this case used? For when we use a GDAL version that does not support returning as string (for the arrow code path?)

pyogrio/geopandas.py
```diff
@@ -37,33 +39,56 @@ def _stringify_path(path):
...
+            return res
+        if __gdal_version__ < (3, 7, 0):
+            # GDAL < 3.7 doesn't return datetimes in ISO8601 format, so fix that
+            return ser.str.replace(" ", "T").str.replace("/", "-")
+        return ser
+    elif datetimes == "UTC":
+        datetime_kwargs["utc"] = True
```
jorisvandenbossche, line 57: So by setting this always in the default case, that means we actually no longer preserve tz-naive data as being tz-naive? Since GDAL has that concept somewhat (and depending on the file format), I would prefer that keep that roundtrip functionality.

jorisvandenbossche: It's a bit annoying that pandas does not give an easy way to get this, and you essentially need to do this try/except calling `to_datetime` a second time ..

pyogrio/geopandas.py
```diff
@@ -37,33 +39,56 @@ def _stringify_path(path):
...
+    elif datetimes == "DATETIME":
+        datetime_kwargs["utc"] = False
+    else:
+        raise ValueError(
+            f"Invalid value for 'datetimes': {datetimes!r}. "
+            "Must be 'UTC' or 'DATETIME'."
```
jorisvandenbossche, line 63: or 'STRING' ?

pyogrio/geopandas.py
```diff
@@ -619,8 +666,33 @@ def write_dataframe(
...
+                first_non_na_index = col.first_valid_index()
+                if first_non_na_index is not None:
+                    if isinstance(col[first_non_na_index], (pd.Timestamp, datetime)):
+                        df[name] = col.astype("string")
+                        datetime_cols.append(name)
+            elif isinstance(dtype, pd.DatetimeTZDtype) and str(dtype.tz) != "UTC":
+                # When it is a datetime column with a timezone different than UTC, it
+                # needs to be converted to string, otherwise the timezone info is lost.
+                df[name] = col.astype("string")
+                datetime_cols.append(name)
```
jorisvandenbossche, line 686: Do you know why GDAL preserves the tz-awareness for UTC, but not for other offsets (even though the values written to the file are in UTC) in the Arrow write path?

pyogrio/geopandas.py
```diff
@@ -223,8 +250,22 @@ def read_dataframe(
...
+    datetimes : str, optional (default: "UTC")
+        The way datetime columns should be returned. Possible values:
+
+        - **"UTC"**: return all datetime columns as pandas datetime64 columns
+          converted to UTC. Naive datetimes (without timezone information) will
+          be assumed to be in UTC timezone.
+        - **"DATETIME"**: return datetimes in the timezone as they were read
+          from the data source. Columns with values in a single timezone or
+          without timezone information will be returned as pandas datetime64
+          columns. Columns with mixed timezone data are returned as object
```
jorisvandenbossche, line 262: Small nitpick on the wording, but I think we should be clear about that it are "timezone-aware columns with mixed offsets", because the general case of mixed offsets (due to DST changes throughout the year) is still a single time _zone_. 
The issue is that GDAL does not actually support the concept of time zones, only offsets.

jorisvandenbossche:
> It is mainly creating the pd.Timestamp objects that is slow: datetime.fromisoformat() just takes 1 second, but Timestamp() takes 4-11 seconds. So an option is also to return datetime.datetime objects, but they are not as "rich" as pd.Timestamp feature-wise which is a bit of a pity...

Given that, I think it would also be fine to return the `datetime.datetime` objects (that's also what pandas did in the past (< 2.0) when it could not parse to datetime64). It's true they have less attributes/methods, but the user can still do that conversion themselves if they need those.

I am not sure if pandas still has that functionality for vectorized conversion to pd.Timestamp under the hood (because it no longer needs it itself)

pyogrio/geopandas.py
theroggy: Yes indeed, GDAL < 3.11 didn't support returning strings for arrow...

pyogrio/geopandas.py
theroggy: Sorry, I reinstated the "old" behaviour incorrectly... should be fixed now.

pyogrio/geopandas.py
theroggy: True, I tried to clarify this.

pyogrio/geopandas.py
```diff
@@ -619,8 +666,33 @@ def write_dataframe(
...
+        for name, dtype in df.dtypes.items():
+            col = df[name]
+            if dtype == "object":
+                # If first non-NA value is a datetime-like object, treat as datetime
+                # column.
+                first_non_na_index = col.first_valid_index()
+                if first_non_na_index is not None:
+                    if isinstance(col[first_non_na_index], (pd.Timestamp, datetime)):
+                        df[name] = col.astype("string")
+                        datetime_cols.append(name)
```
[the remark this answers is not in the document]
theroggy: Thanks for pointing me there! That's indeed cleaner... and faster!

theroggy:
> > I mostly have some remaining concerns about the exact user-facing API.
> > 
> > * `"UTC"` for the default seems a bit misleading, because it is only UTC for tz-aware columns. If you have a tz-naive column, that roundtrips as tz-naive. Also for tz-aware columns, if there is only one offset, you will get the fixed offset, not UTC (unless that behaviour changed with this PR?)
> > * `"DATETIME"`: the important aspect here is that it are datetime/timestamp _objects_, because "datetime" could also mean the dtype, and then it could be interpreted as "use a datetime dtype" (which is datetime64). Also the current PR returns `pd.Timestamp` and not `datetime.datetime`, so that is also a bit confusing (although we should maybe return datetime.datetime?)
> 
> I clearly hadn't yet looked in detail at the code changes ànd the docstring explanation of the keyword .. because I see now that the names of the keyword options indeed more or less match with the behaviour, I just wasn't expecting that behaviour.
> 
> First, as also mentioned in the inline comments, I would like to preserve the current default behaviour of roundtripping tz-naive datetimes. The default of "UTC" no longer does that .. 
Yes, that was unintentional changed behaviour. Fixed.

> For "DATETIME", I had naively interpreted it as always returning objects, but the your docstring explanation of the keyword is clear about still returning datetime64 when that does not loose mixed offset information.
I just want to expose the default behaviour of `pandas.to_datetime` (before it was crippled in pandas 3): return the "richest" format possible depending on the data, without changing the data... Hence also a small preference for pd.Timestamp... An option is also to have both a DATETIME and TIMESTAMP option... to be able to choose between "fastest" option and "most convenient" option?

Name-wise, if we would keep strings an option is also to use "MIXED_TO_UTC", "MIXED_AS_DATETIME" (and possibly "MIXED_AS_TIMESTAMP" to make it clearer what is actually to be expected?

pyogrio/geopandas.py
theroggy: I didn't test it again explicitly, but this follows the general logic that for most timezones daylight saving time is used, leading to (potential) different offsets in a column... and with different offsets in one column the timezone information gets lost... I updates the inline comment to clarify this.

pyogrio/geopandas.py
```diff
@@ -268,11 +362,13 @@ def read_dataframe(
...
-        # For numpy IO, datetimes are read as string values to preserve timezone info
-        # as numpy does not directly support timezones.
-        kwargs["datetime_as_string"] = True
+
+    # Always read datetimes as string values to preserve (mixed) timezone info
+    # correctly. If arrow is not used, it is needed because numpy does not
+    # directly support timezones. If arrow is used, needed because datetime
+    # columns don't support mixed timezone offsets + e.g. for .fgb files
+    # timezone info isn't handled correctly even for unique timezone offsets
+    # if datetimes are not read as string.
```
jorisvandenbossche: This is what was discussed in https://github.com/OSGeo/gdal/issues/11212, i.e. it is not considered a bug on the GDAL side, but just a limitation of the combo of fgb/gpkg (which can have mixed offsets, information not known at the column level) and arrow (needs tz information at the column level, and we don't know the tz before parsing the full file, thus always normalize to UTC)?

pyogrio/geopandas.py
```diff
@@ -627,8 +729,33 @@ def write_dataframe(
...
+                df[name] = df[name].astype("string")
+                datetime_cols.append(name)
+
         table = pa.Table.from_pandas(df, preserve_index=False)
 
+        # Add metadata to datetime columns so GDAL knows they are datetimes.
+        for datetime_col in datetime_cols:
+            table = _add_column_metadata(
+                table, column_metadata={datetime_col: {"GDAL:OGR:type": "DateTime"}}
+            )
```
```suggestion
        table = _add_column_metadata(
            table, column_metadata={col: {"GDAL:OGR:type": "DateTime"} for col in datetime_cols}
        )
```

?

The helper function seems to support multiple keys at once

pyogrio/raw.py
```diff
@@ -392,6 +395,9 @@ def open_arrow(
         ArrowStream object. In the default case, this stream object needs
         to be passed to another library supporting the Arrow PyCapsule
         Protocol to consume the stream of data.
+    datetime_as_string : bool, optional (default: False)
+        If True, will return datetime dtypes as detected by GDAL as strings,
+        as arrow doesn't support e.g. mixed timezones.
```
jorisvandenbossche:
```suggestion
        as Arrow doesn't support e.g. mixed timezones.
```

pyogrio/raw.py
```diff
@@ -429,6 +435,7 @@ def open_arrow(
         Meta is: {
             "crs": "<crs>",
             "fields": <ndarray of field names>,
+            "dtypes": <ndarray of numpy dtypes corresponding to fields>
```
jorisvandenbossche: (definitely not for this PR ;) but was just thinking that we should maybe also return here some (string) repr of the actual GDAL types, instead of this lossy conversion to numpy dtypes)

jorisvandenbossche (commented):
Did another pass over this! Looking good, just went through the code changes, will also take a final look at the test_geopandas_io.py changes. Comments are mostly minor code suggestions.

I have some (final?) thoughts on the API, but will put that in a next comment (and we can maybe discuss at the meeting if you are joining)

jorisvandenbossche:
Regarding the user facing API, i.e. how to expose the currently implemented options, I think I am coming back to my earlier comment about whether it might be clearer to separate it into two options. 

I know that gives two mutually exclusive keywords, which in general is not a sign of a good API design, but some issues with the current combo:

- While the keyword is specific to datetime columns, _when_ the keyword is used still depends on the passed value. I.e. the choice whether you want strings or not is something that applies always. While the choice to use datetime.datetime objects vs cast to UTC _only_ applies in the case you have mixed timezone offsets. This is somewhat addressed by naming those two options `MIXED_TO_<..>`, but it still feels a bit wrong to me.
- The `datetime_as_string` option also exists for the raw numpy and arrow APIs. While those are certainly less used (although the arrow API might get used more), I think there is _some_ value in using a consistent keyword for all different APIs.

If we use `datetime_as_string` for `read_dataframe` as well, then the other two options could be specific about mixed offsets (i.e. essentially how pandas to_datetime has the `utc=True/False` option, although we can make it more explicit). For example, `mixed_timezone_offsets="utc"/"datetime"` or `mixed_offsets_as_utc=True/False`.

pyogrio/geopandas.py
theroggy: The problem described here was solved in `ogr2ogr` the same way as in `pyogrio`: by reading datetimes as string via arrow.

The PR that fixes this in `ogr2ogr` is referenced by rouault in this comment: https://github.com/OSGeo/gdal/issues/11212#issuecomment-2462091204

pyogrio/geopandas.py
```diff
@@ -38,33 +47,93 @@ def _stringify_path(path):
...
+        if __gdal_version__ < (3, 7, 0):
+            # GDAL < 3.7 doesn't return datetimes in ISO8601 format, so fix that
+            return ser.str.replace(" ", "T").str.replace("/", "-")
+        return ser
+    elif datetimes in get_args(DateTimeParseTypes):
+        pass
+    else:
+        raise ValueError(
+            f"Invalid value for 'datetimes': {datetimes!r}. "
+            f"Must be one of {DateTimeParseTypes!r}."
```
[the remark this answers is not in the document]
theroggy: Solved in the change to the API.

pyogrio/geopandas.py
```diff
@@ -38,33 +47,93 @@ def _stringify_path(path):
...
+                elif datetimes == "MIXED_TO_UTC":
+                    # Convert mixed timezone datetimes to UTC.
+                    try:
+                        res = pd.to_datetime(ser, utc=True, **datetime_kwargs)
+                    except Exception as ex:
+                        warnings.warn(warning.format(message=str(ex)), stacklevel=1)
+                        return ser
+                else:
+                    warnings.warn(warning.format(message=str(ex)), stacklevel=1)
+                    return ser
```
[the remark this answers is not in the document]
theroggy: Solved in the change to the API.

pyogrio/raw.py
theroggy: This information was already added to meta in #556, but I forgot to update this docstring apparently.

As it is just adding a few lines of docstrings I included it in this PR.

theroggy:
> If we use `datetime_as_string` for `read_dataframe` as well, then the other two options could be specific about mixed offsets (i.e. essentially how pandas to_datetime has the `utc=True/False` option, although we can make it more explicit). For example, `mixed_timezone_offsets="utc"/"datetime"` or `mixed_offsets_as_utc=True/False`.

As agreed in the community meeting, implementation has been changed to using two booleans as API: `datetime_as_string` and `mixed_offsets_as_utc`.

pyogrio/tests/test_geopandas_io.py
```diff
@@ -451,39 +432,471 @@ def test_write_datetime_mixed_offset(tmp_path, use_arrow):
...
+        # https://github.com/geopandas/pyogrio/issues/487#issuecomment-2423762807
+        exp_dates = df.dates.dt.tz_localize("UTC")
+        if datetime_as_string:
+            exp_dates = exp_dates.astype("string").str.replace(" ", "T")
+            assert_series_equal(result.dates, exp_dates)
+        elif not mixed_offsets_as_utc:
+            assert_series_equal(result.dates, exp_dates)
+        pytest.xfail("naive datetimes read wrong in GPKG with GDAL < 3.11 via arrow")
+
+    if datetime_as_string:
```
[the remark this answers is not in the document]
theroggy: I think it is useful to keep the xfail here so it is clear that this scenario results in an unexpected result for the user?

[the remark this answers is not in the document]
jorisvandenbossche: > I think it is useful to keep the xfail here so it is clear that this scenario results in an unexpected result for the user?

I certainly understand your point, but my preference would be to not keep the xfails, and use xfails for what they meant for: for marking tests that _fail_. And currently, this test, _as written_, does not actually fail. We know that there can be different (unwanted) behaviour depending on versions, but so the test has been written to check those behaviours explicitly.

I would like to keep xfail for tests that are actually failing, and where the xfail can be strict (i.e. result in an xpass if it no longer fails). 
The thing is that right now we only have a few xfails, and you can run the tests with reporting about this enabled, allowing you to easily inspect if those are still up to date (and I see we actually have a few xpassing tests). With the xfails here, it is adding a whole bunch of xfails, making that much less useful IMO. 
If the goal is to make it clear to the reader of the test that this is kind of testing failed behaviour, then I would just add your current xfail message as a comment (`# XFAIL ....`)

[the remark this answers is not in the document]
theroggy: OK, I replaced the xfails where the tests explicitly test the wrong results to `# XFAIL: ` comments so it is at least clear when someone reads the tests.

docs/source/introduction.md
```diff
@@ -481,13 +481,17 @@ Not all file formats have dedicated support to store datetime data, like ESRI
...
 
-Timezone information is preserved where possible, however GDAL only represents
-time zones as UTC offsets, whilst pandas uses IANA time zones (via `pytz` or
-`zoneinfo`). This means that dataframes with columns containing multiple offsets
-(e.g. when switching from standard time to summer time) will be written correctly,
-but when read via `pyogrio.read_dataframe()` will be returned as a UTC datetime
-column, as there is no way to reconstruct the original timezone from the individual
-offsets present.
+Be careful when reading columns with mixed timezone offsets using
+`pyogrio.read_dataframe()`. By default such columns will be returned as a UTC datetime
```
jorisvandenbossche: I would keep the original intro here, i.e. start with explaining how timezone information is supported in general, and only then warn to be careful with mixed offsets.

(can do a push with some proposed rewording)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -550,39 +527,493 @@ def test_write_datetime_mixed_offset(tmp_path, use_arrow):
...
+
+            if overflow_occured:
+                # Strings are returned because of an overflow.
+                assert is_string_dtype(df.datetime_col.dtype)
+                assert_series_equal(df.datetime_col, exp_dates_str, check_dtype=False)
+            else:
+                # With use_arrow or pandas >= 3.0, old datetimes are parsed correctly.
+                assert is_datetime64_dtype(df.datetime_col)
+                assert df.datetime_col.iloc[0] == pd.Timestamp(1670, 1, 1, 9, 0, 0)
+                assert df.datetime_col.iloc[0].unit == "ms"
```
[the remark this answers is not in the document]
theroggy: I suppose I didn't find an easy solution to create the `exp_dates` series when pandas<3.0 and use_arrow=True, and so I replaced it with just checking if the value and unit is correct instead.

theroggy:
> All green! I want to do some more clarifications in the tests, but that can wait for later ;)

Great!

> I also made some small changes to the docstrings. The diff makes it look like it all changes because of reflowing the text, but it were only minor edits I think

OK! Minor edits indeed. As you changed "timezone" to "time zone" in some places, I suppose you have a preference for that... so I made the change consistent through the code base.

PR #495, opened 2024-11-10
ENH: capture all errors logged by gdal when opening a file fails

### LINKED ISSUE #491: ENH: keep full gdal error when error is thrown by gdal while reading a file
When an error occurs in gdal while reading a file, detail in the error seems to be lost when the error is rethrown in pyogrio.

Sample case here:
https://stackoverflow.com/questions/79162109/geopandas-fails-reading-geojson/79165236#79165236

theroggy:
This was present in Rasterio (https://github.com/rasterio/rasterio/pull/2526) and has been included in fiona::_err recently (https://github.com/Toblerity/Fiona/pull/1408)... so this PR is based on the change there.

Closes #491

### DIFF
--- a/pyogrio/_err.pyx
+++ b/pyogrio/_err.pyx
@@ -1,11 +1,25 @@
-# ported from fiona::_err.pyx
-from enum import IntEnum
+"""Error handling code for GDAL/OGR.
+
+Ported from fiona::_err.pyx
+"""
+
+import contextlib
+import logging
 import warnings
+from contextvars import ContextVar
+from enum import IntEnum
+from itertools import zip_longest
 
 from pyogrio._ogr cimport (
     CE_None, CE_Debug, CE_Warning, CE_Failure, CE_Fatal, CPLErrorReset,
     CPLGetLastErrorType, CPLGetLastErrorNo, CPLGetLastErrorMsg, OGRErr,
-    CPLErr, CPLErrorHandler, CPLDefaultErrorHandler, CPLPushErrorHandler)
+    CPLErr, CPLErrorHandler, CPLDefaultErrorHandler, CPLPopErrorHandler,
+    CPLPushErrorHandler, CPLPushErrorHandlerEx)
+
+log = logging.getLogger(__name__)
+
+_ERROR_STACK = ContextVar("error_stack")
+_ERROR_STACK.set([])
 
 
 # CPL Error types as an enum.
@@ -17,9 +31,9 @@ class GDALError(IntEnum):
     fatal = CE_Fatal
 
 
-
 class CPLE_BaseError(Exception):
     """Base CPL error class.
+
     For internal use within Cython only.
     """
 
@@ -103,6 +117,10 @@ class CPLE_AWSSignatureDoesNotMatchError(CPLE_BaseError):
     pass
 
 
+class CPLE_AWSError(CPLE_BaseError):
+    pass
+
+
 class NullPointerError(CPLE_BaseError):
     """
     Returned from exc_wrap_pointer when a NULL pointer is passed, but no GDAL
@@ -111,6 +129,21 @@ class NullPointerError(CPLE_BaseError):
     pass
 
 
+class CPLError(CPLE_BaseError):
+    """
+    Returned from exc_wrap_int when a error code is returned, but no GDAL
+    error was set.
+    """
+    pass
+
+
+cdef dict _LEVEL_MAP = {
+    0: 0,
+    1: logging.DEBUG,
+    2: logging.WARNING,
+    3: logging.ERROR,
+    4: logging.CRITICAL
+}
 
 # Map of GDAL error numbers to the Python exceptions.
 exception_map = {
@@ -132,12 +165,51 @@ exception_map = {
     13: CPLE_AWSObjectNotFoundError,
     14: CPLE_AWSAccessDeniedError,
     15: CPLE_AWSInvalidCredentialsError,
-    16: CPLE_AWSSignatureDoesNotMatchError
+    16: CPLE_AWSSignatureDoesNotMatchError,
+    17: CPLE_AWSError
+}
+
+cdef dict _CODE_MAP = {
+    0: 'CPLE_None',
+    1: 'CPLE_AppDefined',
+    2: 'CPLE_OutOfMemory',
+    3: 'CPLE_FileIO',
+    4: 'CPLE_OpenFailed',
+    5: 'CPLE_IllegalArg',
+    6: 'CPLE_NotSupported',
+    7: 'CPLE_AssertionFailed',
+    8: 'CPLE_NoWriteAccess',
+    9: 'CPLE_UserInterrupt',
+    10: 'ObjectNull',
+    11: 'CPLE_HttpResponse',
+    12: 'CPLE_AWSBucketNotFound',
+    13: 'CPLE_AWSObjectNotFound',
+    14: 'CPLE_AWSAccessDenied',
+    15: 'CPLE_AWSInvalidCredentials',
+    16: 'CPLE_AWSSignatureDoesNotMatch',
+    17: 'CPLE_AWSError'
 }
 
 
+cdef class GDALErrCtxManager:
+    """A manager for GDAL error handling contexts."""
+
+    def __enter__(self):
+        CPLErrorReset()
+        return self
+
+    def __exit__(self, exc_type=None, exc_val=None, exc_tb=None):
+        cdef int err_type = CPLGetLastErrorType()
+        cdef int err_no = CPLGetLastErrorNo()
+        cdef const char *msg = CPLGetLastErrorMsg()
+        # TODO: warn for err_type 2?
+        if err_type >= 2:
+            raise exception_map[err_no](err_type, err_no, msg)
+
+
 cdef inline object exc_check():
-    """Checks GDAL error stack for fatal or non-fatal errors
+    """Checks GDAL error stack for fatal or non-fatal errors.
+
     Returns
     -------
     An Exception, SystemExit, or None
@@ -173,8 +245,31 @@ cdef inline object exc_check():
         return
 
 
+cdef get_last_error_msg():
+    """Checks GDAL error stack for the latest error message.
+
+    Returns
+    -------
+    An error message or empty string
+
+    """
+    err_msg = CPLGetLastErrorMsg()
+
+    if err_msg != NULL:
+        # Reformat messages.
+        msg_b = err_msg
+        msg = msg_b.decode('utf-8')
+        msg = msg.replace("`", "'")
+        msg = msg.replace("\n", " ")
+    else:
+        msg = ""
+
+    return msg
+
+
 cdef void *exc_wrap_pointer(void *ptr) except NULL:
-    """Wrap a GDAL/OGR function that returns GDALDatasetH etc (void *)
+    """Wrap a GDAL/OGR function that returns GDALDatasetH etc (void *).
+
     Raises an exception if a non-fatal error has be set or if pointer is NULL.
     """
     if ptr == NULL:
@@ -188,10 +283,9 @@ cdef void *exc_wrap_pointer(void *ptr) except NULL:
 
 
 cdef int exc_wrap_int(int err) except -1:
-    """Wrap a GDAL/OGR function that returns CPLErr or OGRErr (int)
-    Raises an exception if a non-fatal error has be set.
+    """Wrap a GDAL/OGR function that returns CPLErr or OGRErr (int).
 
-    Copied from Fiona (_err.pyx).
+    Raises an exception if a non-fatal error has be set.
     """
     if err:
         exc = exc_check()
@@ -218,9 +312,11 @@ cdef int exc_wrap_ogrerr(int err) except -1:
 cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil:
     """Custom CPL error handler to match the Python behaviour.
 
-    Generally we want to suppress error printing to stderr (behaviour of the
-    default GDAL error handler) because we already raise a Python exception
-    that includes the error message.
+    For non-fatal errors (CE_Failure), error printing to stderr (behaviour of
+    the default GDAL error handler) is suppressed, because we already raise a
+    Python exception that includes the error message.
+
+    Warnings are converted to Python warnings.
     """
     if err_class == CE_Fatal:
         # If the error class is CE_Fatal, we want to have a message issued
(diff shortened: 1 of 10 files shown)


pyogrio/_err.pyx
```diff
@@ -1,11 +1,25 @@
...
+from itertools import zip_longest
 
 from pyogrio._ogr cimport (
     CE_None, CE_Debug, CE_Warning, CE_Failure, CE_Fatal, CPLErrorReset,
     CPLGetLastErrorType, CPLGetLastErrorNo, CPLGetLastErrorMsg, OGRErr,
-    CPLErr, CPLErrorHandler, CPLDefaultErrorHandler, CPLPushErrorHandler)
+    CPLErr, CPLErrorHandler, CPLDefaultErrorHandler, CPLPopErrorHandler,
+    CPLPushErrorHandler, CPLPushErrorHandlerEx)
+
+log = logging.getLogger(__name__)
```
jorisvandenbossche, line 19: Do we need to logging part of this? (in contrast to fiona, we in general don't enable logging)

pyogrio/_err.pyx
```diff
@@ -173,8 +245,31 @@ cdef inline object exc_check():
         return
 
 
+cdef get_last_error_msg():
```
jorisvandenbossche, line 248: This is not actually used yet? Above in the new code you directly use `CPLGetLastErrorMsg()`, but I suppose that could use this helper?

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+        elif err_no == 0:
+            log.log(_LEVEL_MAP[err_class], "%s", msg.decode("utf-8"))
+        else:
+            log.log(_LEVEL_MAP[err_class], "%s:%s", _CODE_MAP[err_no], msg.decode("utf-8"))
+    else:
+        log.info("Unknown error number %r", err_no)
+
+
+IF UNAME_SYSNAME == "Windows":
+    cdef void __stdcall stacking_error_handler(
```
jorisvandenbossche, line 426: Is the only difference this `__stdcall`? And any idea why his is needed? (we don't have it for the `error_handler` defined above)

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+cdef class StackChecker:
+
+    def __init__(self, error_stack=None):
+        self.error_stack = error_stack or {}
+
+    cdef int exc_wrap_int(self, int err) except -1:
+        """Wrap a GDAL/OGR function that returns CPLErr (int).
+
+        Raises an exception if a non-fatal error has been added to the
+        exception stack.
```
brendan-ward, line 360:
```suggestion
        exception stack.  Only checks for errors if `err` is set to a nonzero value.
```

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+                if last is not None:
+                    raise last
+
+        return err
+
+    cdef void *exc_wrap_pointer(self, void *ptr) except NULL:
+        """Wrap a GDAL/OGR function that returns a pointer.
+
+        Raises an exception if a non-fatal error has been added to the
+        exception stack.
```
brendan-ward, line 379:
```suggestion
        exception stack.  Only checks for errors if `ptr` is `NULL`.
```

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
 
 def _register_error_handler():
     CPLPushErrorHandler(<CPLErrorHandler>error_handler)
+
+cpl_errs = GDALErrCtxManager()
+
+
+cdef class StackChecker:
```
brendan-ward, line 351:
```suggestion
cdef class ErrorHandler:
```

The stack part seems an implementation detail, I think the part we want to focus here on the name for usage below is that it checks for errors.

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+
+    def __init__(self, error_stack=None):
+        self.error_stack = error_stack or {}
+
+    cdef int exc_wrap_int(self, int err) except -1:
```
brendan-ward, line 356:
```suggestion
    cdef int check_int(self, int err) except -1:
```

I wonder if it is time for us to diverge from Fiona's original naming here and try to make this a bit more clear.
(also good to consider doing this for the top-level function names too.)

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+                    error.__cause__ = cause
+
+            if stack:
+                last = stack.pop()
+                if last is not None:
+                    raise last
+
+        return err
+
+    cdef void *exc_wrap_pointer(self, void *ptr) except NULL:
```
brendan-ward, line 375:
```suggestion
    cdef void *check_pointer(self, void *ptr) except NULL:
```

pyogrio/_err.pyx
```diff
@@ -248,3 +344,197 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+
+            return
+
+        # Fall back to the default handler for non-failure messages since
+        # they won't be translated into exceptions.
+        CPLDefaultErrorHandler(err_class, err_no, err_msg)
+
+
+@contextlib.contextmanager
+def stack_errors():
```
brendan-ward, line 519:
```suggestion
def get_error_handler():
```

The purpose of the context manager is to get the error handler.

pyogrio/_io.pyx
```diff
@@ -196,28 +205,33 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
-
     try:
         # WARNING: GDAL logs warnings about invalid open options to stderr
         # instead of raising an error
-        ogr_dataset = exc_wrap_pointer(
-            GDALOpenEx(path_c, flags, NULL, <const char *const *>options, NULL)
-        )
-
-        return ogr_dataset
+        with stack_errors() as error_stack_checker:
```
brendan-ward, line 211:
```suggestion
        with get_error_handler() as error_handler:
```

(to demonstrate increased clarity of above suggestions)

pyogrio/_io.pyx
```diff
@@ -196,28 +205,33 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
+            ogr_dataset = GDALOpenEx(path_c, flags, NULL, <const char *const *>options, NULL)
+            return error_stack_checker.exc_wrap_pointer(ogr_dataset)
```
brendan-ward, line 213:
```suggestion
            return error_handler.check_pointer(ogr_dataset)
```
(ditto above)

brendan-ward (commented):
Thanks for working on this @theroggy !  A few suggestions re: naming things.

theroggy:
> Thanks for looking into this! Added a few quick questions
> 
> Do you have a link to the PRs that added this to rasterio/fiona (that might have some background)?

I added links to them in the [PR description](https://github.com/geopandas/pyogrio/pull/495#issue-2647800422)...

pyogrio/_err.pyx
theroggy: True. I extracted the error message cleaning code in a new function `clean_error_message` and now use it where applicable.

pyogrio/_err.pyx
theroggy: Not sure how interesting the logging is in general, but as it was apparently consciously removed when porting from fiona I suppose it's not that interesting, so I'll remove it.

pyogrio/_err.pyx
theroggy: Yes, that's the only difference. I had a quick further look, and apparently `__stdcall` is even ignored by ARM and x64 compilers, so there is probably not a lot of use in keeping it, so I removed it.

pyogrio/_err.pyx
```diff
@@ -1,25 +1,41 @@
-# ported from fiona::_err.pyx
-from enum import IntEnum
+"""Error handling code for GDAL/OGR.
+
+Ported from fiona::_err.pyx
+"""
+
+import contextlib
+import logging
```
jorisvandenbossche, line 7:
```suggestion
```

pyogrio/_err.pyx
```diff
@@ -132,49 +159,114 @@ exception_map = {
...
+    """A manager for GDAL error handling contexts."""
+
+    def __enter__(self):
+        CPLErrorReset()
+        return self
+
+    def __exit__(self, exc_type=None, exc_val=None, exc_tb=None):
+        cdef int err_type = CPLGetLastErrorType()
+        cdef int err_no = CPLGetLastErrorNo()
+        cdef const char *msg = CPLGetLastErrorMsg()
```
jorisvandenbossche, line 198: Should this also use the `get_last_error_msg()` helper?

pyogrio/_err.pyx
```diff
@@ -132,49 +159,114 @@ exception_map = {
...
+    16: 'CPLE_AWSSignatureDoesNotMatch',
+    17: 'CPLE_AWSError'
 }
 
 
+cdef class GDALErrCtxManager:
+    """A manager for GDAL error handling contexts."""
+
+    def __enter__(self):
+        CPLErrorReset()
```
jorisvandenbossche, line 192: Should this also do a `CPLErrorReset()` at the end of `__exit__`?

pyogrio/_err.pyx
```diff
@@ -132,49 +159,114 @@ exception_map = {
...
+        return self
+
+    def __exit__(self, exc_type=None, exc_val=None, exc_tb=None):
+        cdef int err_type = CPLGetLastErrorType()
+        cdef int err_no = CPLGetLastErrorNo()
+        cdef const char *msg = CPLGetLastErrorMsg()
+        # TODO: warn for err_type 2?
+        if err_type >= 2:
+            raise exception_map[err_no](err_type, err_no, msg)
```
jorisvandenbossche, line 201: We use `exception_map.get(err_no, CPLE_BaseError)` (for the case that there is an err_no we are not aware of, maybe also for future compatibility?), should we use that here as well?

pyogrio/_err.pyx
```diff
@@ -248,3 +339,121 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+    Yields an ErrorHandler object that can be used to handle the errors
+    if any were captured.
+    """
+    CPLErrorReset()
+    global _ERROR_STACK
+    _ERROR_STACK.set([])
+
+    # stacking_error_handler records GDAL errors in the order they occur and
+    # converts them to exceptions.
+    CPLPushErrorHandlerEx(<CPLErrorHandler>stacking_error_handler, NULL)
```
jorisvandenbossche, line 452: If we would (eventually?) handle all catching of GDAL errors using this context manager, then we don't need to each time register this error handler and pop it again?

pyogrio/_err.pyx
theroggy: Ok, renamed.

pyogrio/_err.pyx
theroggy: ok, renamed

pyogrio/_io.pyx
theroggy: I renamed the context handler to `capture_errors` instead of `get_error_handler` as I think this is a clearer description of the behaviour.

OK?

pyogrio/_err.pyx
theroggy: In the stack, they are still c const char* 's... it is only when the stack is being interpreted that the messages are converted to python objects...

theroggy: This context handler isn't used anywhere in pyogrio, and thinking further on it I'm also not sure why there a two gdal error context handlers in fiona/rasterio so I removed it.

pyogrio/_err.pyx
theroggy: True. I'm not sure why both approaches (or even three taking the second context handler in account) are being used in rasterio/fiona...

In fiona at least it is only used when files are opened, all other calls use the getlasterror approach... possibly there is a downside to capturing all errors in some cases?

pyogrio/_err.pyx
```diff
@@ -1,25 +1,40 @@
...
 class GDALError(IntEnum):
+    """GDAL error types/classes.
+    
+    GDAL doc: https://gdal.org/en/latest/doxygen/cpl__error_8h.html#a463ba7c7202a505416ff95b1aeefa2de
+    """
     none = CE_None
     debug = CE_Debug
     warning = CE_Warning
     failure = CE_Failure
     fatal = CE_Fatal
```
theroggy, line 32: This enum isn't use anywhere, should it be kept here or shall I remove it?

brendan-ward: Removing it is fine, likely a leftover from porting from Fiona in the early days.

pyogrio/_err.pyx
jorisvandenbossche: > I'm not sure why both approaches are being used in rasterio/fiona...

That might also be partly just historical reasons, since this was only added more recently to fiona (and although it could be used everywhere, there was maybe also no pressing need to change the code that much?)

pyogrio/_err.pyx
theroggy: Code removed...

pyogrio/_err.pyx
theroggy: Code removed...

pyogrio/_err.pyx
```diff
@@ -132,53 +144,88 @@ exception_map = {
...
         except UnicodeDecodeError as exc:
-            msg = f"Could not decode error message to UTF-8.  Raw error: {msg_b}"
+            msg = f"Could not decode error message to UTF-8. Raw error: {msg_b}"
 
-    if err_type == 3:
-        CPLErrorReset()
-        return exception_map.get(
-            err_no, CPLE_BaseError)(err_type, err_no, msg)
+    else:
+        msg = ""
```
jorisvandenbossche, line 214: Do we want to keep "No error message." here? Or we don't want that for the case that we are stacking multiple errors where not every error might have a message?

pyogrio/_err.pyx
```diff
@@ -187,40 +234,45 @@ cdef void *exc_wrap_pointer(void *ptr) except NULL:
...
-            raise CPLE_BaseError(-1, -1, "Unspecified OGR / GDAL error")
+            raise CPLError(-1, -1, "Unspecified OGR / GDAL error")
+
     return err
 
 
-cdef int exc_wrap_ogrerr(int err) except -1:
-    """Wrap a function that returns OGRErr (int) but does not use the
-    CPL error stack.
+cdef int check_ogrerr(int err) except -1:
```
jorisvandenbossche, line 257: Maybe for another PR, but I am wondering if we should keep this helper function. 
At the moment we use this for calling `OGR_L_GetExtent`, but then if the return value is non-zero, that indicates the bounds are not known and we catch the error to set the bounds to `None`. But so that checking for non-zero return we can do directly where we call `OGR_L_GetExtent`, instead of raising a python error and catching that again to do the same.

And the other place it is used is in `exc_wrap_ogrerr(exc_wrap_int(OSRImportFromEPSG(srs, 4326)))`, which seems a bit fishy in that it chains both exception handlers. I don't think the `exc_wrap_ogrerr` has any effect there.

pyogrio/_err.pyx
```diff
@@ -248,3 +298,127 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+    """A context manager that captures all GDAL non-fatal errors occuring.
+
+    It adds all errors to a single stack, so it assumes that no more than one
+    GDAL function is called.
+
+    Yields an ErrorHandler object that can be used to handle the errors
+    if any were captured.
+    """
+    CPLErrorReset()
+    global _ERROR_STACK
```
jorisvandenbossche, line 412: What is the reason this needs to be a global variable? Because otherwise `stacking_error_handler` cannot access it? (being a `cdef` function, we would not be able to pass a python object as argument it?)

pyogrio/_err.pyx
```diff
@@ -248,3 +298,127 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+    _ERROR_STACK.set([])
+
+    # stacking_error_handler records GDAL errors in the order they occur and
+    # converts them to exceptions.
+    CPLPushErrorHandlerEx(<CPLErrorHandler>stacking_error_handler, NULL)
```
jorisvandenbossche, line 417:
```suggestion
    CPLPushErrorHandler(<CPLErrorHandler>stacking_error_handler)
```

pyogrio/_io.pyx
```diff
@@ -196,28 +205,33 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
         ) from None
 
     except CPLE_BaseError as exc:
-        if str(exc).endswith("a supported file format."):
+        # If there are inner exceptions, append their errmsg to the errmsg
+        errmsg = str(exc)
+        inner = exc.__cause__
+        while inner is not None:
+            errmsg = f"{errmsg}; {inner}"
+            inner = inner.__cause__
```
jorisvandenbossche, line 226: Is this not something we want to do in the logic in `_err.pyx` ?

jorisvandenbossche (commented):
Thanks! A few more comments/questions

pyogrio/_err.pyx
jorisvandenbossche: Ah, and also of course the `stacking_error_handler` has a fixed signature and gets called from GDAL, so if we want to pass some custom data to that function, that's what the user data could be used for in `CPLPushErrorHandlerEx` (and retrieved with `CPLGetErrorHandlerUserData`).

I suppose we could use that to pass through the error stack, but at that point using a global variable might be simpler

pyogrio/_err.pyx
```diff
@@ -1,25 +1,26 @@
...
-class GDALError(IntEnum):
-    none = CE_None
-    debug = CE_Debug
-    warning = CE_Warning
-    failure = CE_Failure
-    fatal = CE_Fatal
+    CPLErr, CPLErrorHandler, CPLDefaultErrorHandler, CPLPopErrorHandler,
+    CPLPushErrorHandler, CPLPushErrorHandlerEx)
 
+_ERROR_STACK = ContextVar("error_stack")
```
jorisvandenbossche, line 17: Do you know the reason for using a ContextVar here? (and not a thread-local variable) from quickly reading up on it, it seems this is mostly used to have something similar to thread-local storage but in async / coroutines?

pyogrio/_err.pyx
jorisvandenbossche: But, isn't the point of a context var that you don't need to use `global` for it to work? Does it still work if you just remove this line?

pyogrio/_err.pyx
theroggy: No, I didn't really look into it in that detail...

pyogrio/_io.pyx
theroggy: I've been wondering about this as well.

I'll add a parameter "squash_errors" or something like to the `check_...` functions that to have the choice in having the errors returned as a stack of exceptions or as one exception with all error messages "squashed" to one.

theroggy: Done.

pyogrio/_err.pyx
theroggy: I had a quick read, and indeed... I'd guess as well that the `global ...` statements shouldn't be there. I remover them, and the tests still pass without issues.

pyogrio/_err.pyx
theroggy: That's why I retained the empty string. I'm not sure if having no error message actually happens... it seems there is no test for it either...

pyogrio/_err.pyx
theroggy: In both cases the added value is indeed not very obvious...

theroggy: I'll remove it.

pyogrio/_err.pyx
theroggy: I moved the check so "No error message." will be shown in the typical cases where there would be an exception without message...

brendan-ward:
This appears to be raising a lot of test errors on GDAL 3.4.3 specifically; is that expected?

theroggy:
> This appears to be raising a lot of test errors on GDAL 3.4.3 specifically; is that expected?

No... I've had a quick look at it... but I didn't immediately see why these errors occur...

theroggy:
@brendan-ward I found and fixed the issue with gdal < 3.5.2...

pyogrio/_err.pyx
```diff
@@ -132,53 +144,79 @@ exception_map = {
...
+
+    if err_type == CE_Fatal:
+        return SystemExit("Fatal error: {0}".format((err_type, err_no, err_msg)))
 
-    if err_msg == NULL:
-        msg = "No error message."
     else:
-        # Reformat messages.
-        msg_b = err_msg
+        return
```
brendan-ward, line 178: Suggest removing the `else` branch entirely

pyogrio/_err.pyx
```diff
@@ -248,3 +277,170 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+        -------
+        int
+            The `err` input parameter if it is OGRERR_NONE. Otherwise an exception is
+            raised.
+
+        """
+        if err != OGRERR_NONE:
+            stack = self.error_stack.get()
+
+            if stack:
```
brendan-ward, line 314:
```suggestion
            if self.error_stack.get():
```

(and remove line 312, and similar changes below)

pyogrio/_err.pyx
```diff
@@ -248,3 +277,170 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+    Warnings are converted to Python warnings.
+    """
+    if err_class == CE_Fatal:
+        # If the error class is CE_Fatal, we want to have a message issued
+        # because the CPL support code does an abort() before any exception
+        # can be generated
+        CPLDefaultErrorHandler(err_class, err_no, err_msg)
+        return
+
+    elif err_class == CE_Failure:
```
brendan-ward, line 401:
```suggestion
    if err_class == CE_Failure:
```

(`elif` not needed because above returns)

pyogrio/_err.pyx
```diff
@@ -248,3 +277,170 @@ cdef void error_handler(CPLErr err_class, int err_no, const char* err_msg) nogil
...
+            stack.append(
+                exception_map.get(err_no, CPLE_BaseError)(
+                    err_class, err_no, clean_error_message(err_msg)
+                ),
+            )
+            _ERROR_STACK.set(stack)
+
+        return
+
+    elif err_class == CE_Warning:
```
brendan-ward, line 414:
```suggestion
    if err_class == CE_Warning:
```

pyogrio/_io.pyx
```diff
@@ -196,27 +205,27 @@ cdef void* ogr_open(const char* path_c, int mode, char** options) except NULL:
...
-
-        return ogr_dataset
+        with capture_errors() as errors:
+            ogr_dataset = GDALOpenEx(path_c, flags, NULL, <const char *const *>options, NULL)
+            return errors.check_pointer(ogr_dataset, True)
 
     except NullPointerError:
         raise DataSourceError(
-            "Failed to open dataset (mode={}): {}".format(mode, path_c.decode("utf-8"))
+            f"Failed to open dataset ({mode=}): {path_c.decode('utf-8')}"
```
brendan-ward, line 217:
```suggestion
            f"Failed to open dataset (mode={mode}): {path_c.decode('utf-8')}"
```

brendan-ward (commented):
Thanks for updating @theroggy ; just a few minor comments.

pyogrio/_io.pyx
theroggy: Why change this, as both are equivalent, but the existing syntax seems more concise ([ref](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging))?

brendan-ward: Was not familiar with that way of using F-strings!  I thought it was a typo in converting from `.format` to F-string.  Fine to leave as is, then.

brendan-ward (approved):
Thanks @theroggy !

PR #497, opened 2024-11-17
Fix WKB writing on big-endian systems

QuLogic:
`shapely.to_wkb` (called in `pyogrio.geopandas.write_dataframe` via `geopandas.array.to_wkb`) defaults to machine byte order, so simply reading the first byte (of a 4-byte integer geometry type) is broken on big-endian machines. And since some geometry type could be 2 bytes, this also seems broken on little-endian machines, though I did not check if any 2-byte types could be produced.

The commented-out code here indicates the correct course of action (reading the first byte to determine byte order).

This fixes the 190 tests that go through this function and fail on big-endian.

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2253,7 +2253,8 @@ def ogr_write(
     cdef OGRGeometryH ogr_geometry_multi = NULL
     cdef OGRFeatureDefnH ogr_featuredef = NULL
     cdef OGRFieldDefnH ogr_fielddef = NULL
-    cdef unsigned char *wkb_buffer = NULL
+    cdef const unsigned char *wkb_buffer = NULL
+    cdef unsigned int wkbtype = 0
     cdef int supports_transactions = 0
     cdef int err = 0
     cdef int i = 0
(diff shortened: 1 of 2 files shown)


QuLogic:
Note, if you'd prefer to not use the Python `int.from_bytes`, I could write it with bytes and bitshifting. This usually compiles to a single instruction with optimizing compilers (though I'm unsure about the result through Cython.)

QuLogic:
> Note, if you'd prefer to not use the Python `int.from_bytes`, I could write it with bytes and bitshifting. This usually compiles to a single instruction with optimizing compilers (though I'm unsure about the result through Cython.)

Since Windows didn't like the `int.from_bytes`, I decided to implement this. (Probably it just needed the explicit type, but I can't check here.) I also changed `wkb_buffer` to `const` as it doesn't need to be written to, and potentially saves a copy using it directly.

brendan-ward (commented):
Thanks @QuLogic !  Can you please add an entry to the changelog?

Do you have suggestions on how we can test for this in CI?  I'm not seeing a straightforward way to do this with Github Actions, and it seems like it would require building the full stack GDAL, GEOS, Shapely, etc for something like `s390x`.

pyogrio/_io.pyx
```diff
@@ -2373,15 +2374,18 @@ def ogr_write(
...
-                wkbtype = <int>bytearray(wkb)[1]
-                # may need to consider all 4 bytes: int.from_bytes(wkb[0][1:4], byteorder="little")
-                # use "little" if the first byte == 1
+                wkb_buffer = wkb
+                if wkb_buffer[0] == 1:
+                    # Little endian WKB type.
+                    wkbtype = wkb_buffer[1] + (wkb_buffer[2] << 8) + (wkb_buffer[3] << 16) + (wkb_buffer[4] << 24)
+                else:
+                    # Big endian WKB type.
+                    wkbtype = (wkb_buffer[1] << 24) + (wkb_buffer[2] << 16) + (wkb_buffer[3] << 8) + wkb_buffer[4]
```
rouault, line 2383: Due to a subtlety of type promotion in C, a unsigned char shifted 24 position to the left is first cast to a *signed* integer, which can thus cause undefined behavior if the most-significant-bit of this unsigned char is set:

```suggestion
                    wkbtype = wkb_buffer[1] + (wkb_buffer[2] << 8) + (wkb_buffer[3] << 16) + ((unsigned)wkb_buffer[4] << 24)
                else:
                    # Big endian WKB type.
                    wkbtype = ((unsigned)wkb_buffer[1] << 24) + (wkb_buffer[2] << 16) + (wkb_buffer[3] << 8) + wkb_buffer[4]
```

QuLogic: Yes, good catch. I didn't think this was possible, but it looks like the types with Z data get the highest bit set.

brendan-ward:
@QuLogic would you mind please merging in the latest from `main`?  We recently fixed the outstanding CI issues.  Once everything shows up green here, I think we can go ahead and merge this, and save implementing specific CI tests for big endian systems for another PR.

QuLogic:
Looks good now.

brendan-ward (approved):
Thanks @QuLogic !

PR #499, opened 2024-11-28
MNT: Upgrade to GDAL 3.10.0

### LINKED ISSUE #500: BUG: failure during parsing GDAL version number on windows during setup
Per our latest CI failures with conda-forge on Windows:

```Python
...setuptools\build_meta.py", line 320, in run_setup
      exec(code, locals())
    File "<string>", line 156, in <module>
    File "<string>", line 156, in <genexpr>
  ValueError: invalid literal for int() with base 10: ''
  [end of output]
```

If I follow, it is failing on line 156 of `setup.py`:

```Python
gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
```

The GDAL version in conda-forge used here looks valid, so I'm not yet sure why this is failing to parse.

brendan-ward:
supersedes #498

Closes #500

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,11 +153,6 @@ else:
 
     ext_options, gdal_version_str = get_gdal_config()
 
-    # FIXME: remove
-    print(
-        f'GDAL version string: {gdal_version_str}; split: {gdal_version_str.strip("dev").split(".")}'
-    )
-
     gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
     if not gdal_version >= MIN_GDAL_VERSION:
         sys.exit(f"GDAL must be >= {'.'.join(map(str, MIN_GDAL_VERSION))}")
--- a/setup.py
+++ b/setup.py
@@ -153,6 +153,11 @@ else:
 
     ext_options, gdal_version_str = get_gdal_config()
 
+    # FIXME: remove
+    print(
+        f'GDAL version string: {gdal_version_str}; split: {gdal_version_str.strip("dev").split(".")}'
+    )
+
     gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
     if not gdal_version >= MIN_GDAL_VERSION:
         sys.exit(f"GDAL must be >= {'.'.join(map(str, MIN_GDAL_VERSION))}")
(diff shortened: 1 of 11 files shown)


jorisvandenbossche:
So the fix of https://github.com/geopandas/pyogrio/pull/496 for `uv` seems to not have been working for the release test builds (and they were also not tested on that PR because of the failing windows wheels .. although for the sdist test build it does work).

It seems that the ones that did work are running in a docker container, and so the `uv` set up on a plain github actions runner is still failing

jorisvandenbossche:
Hmm, it's bizarre that some environment fail doing the `source ..` (ubuntu, recent mac), but then when removing it seems that others (older mac) requires doing it for uv to work .. And Windows now also requires some activation step, which was before not needed.

We should maybe use the github action to install uv instead? (https://docs.astral.sh/uv/guides/integration/github/) 
For the other places where we were using uv with a docker image that might not have been possible, but I think for this test-wheels job we should be able to use it?

brendan-ward:
Note: `macos-latest` is currently set to `macos-14` and succeeds, whereas `macos-12` is the older version and fails.

On `macos-12`, the install command for `uv` emits instructions to source it:

```
installing to /Users/runner/.local/bin
  uv
  uvx
everything's installed!

To add $HOME/.local/bin to your PATH, either restart your shell or run:

    source $HOME/.local/bin/env (sh, bash, zsh)
    source $HOME/.local/bin/env.fish (fish)
```

where as that instruction is omitted on `macos-latest`.  Likewise, on `windows-latest`, it instructs us to set some environment variables:

```
Installing to C:\Users\runneradmin\.local\bin
  uv.exe
  uvx.exe
everything's installed!

To add C:\Users\runneradmin\.local\bin to your PATH, either restart your system or run:

    set Path=C:\Users\runneradmin\.local\bin;%Path%   (cmd)
    $env:Path = "C:\Users\runneradmin\.local\bin;$env:Path"   (powershell)
```

That said, it is certainly worth testing if the action will work better for all runners we're using here.  Just pushed up a test of that.

brendan-ward:
Using the `uv` action worked!  All was green prior to the final cleanup push.

setup.py
```diff
@@ -153,6 +153,11 @@ def get_gdal_config():
 
     ext_options, gdal_version_str = get_gdal_config()
 
+    # FIXME: remove
+    print(
+        f'GDAL version string: {gdal_version_str}; split: {gdal_version_str.strip("dev").split(".")}'
+    )
```
jorisvandenbossche, line 159: This can still be removed?

jorisvandenbossche (approved):

PR #505, opened 2024-12-06
CI: Upgrade to MacOS 13 runner

### LINKED ISSUE #504: MNT: upgrade MacOS version used for release
The `macos-12` runners are now [unsupported](https://github.com/actions/runner-images/issues/10721) and raise errors.

brendan-ward:
Resolves #504 

Note: this is the last MacOS runner that supports x86_64; MacOS >= 14 is Arm64 only.  (not that we have any control over that, just mentioning it)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -152,13 +152,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: "macos-12"
+          - os: "macos-13"
             triplet: "x64-osx-dynamic-release"
             arch: x86_64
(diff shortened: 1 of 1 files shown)


martinfleis (approved):

theroggy (approved):

jorisvandenbossche:
> Note: this is the last MacOS runner that supports x86_64; MacOS >= 14 is Arm64 only. (not that we have any control over that, just mentioning it)

I suppose the newer MacOS will still support to cross-compile, we just won't be able to test the wheels (just as initially we were also only building but not testing the arm wheels), if we want to keep providing such wheels then.

PR #509, opened 2024-12-19
CI: use pyarrow-core as dependency in tests

theroggy:
We don't need/use the full pyarrow... and since pyarrow 16 there is a "pyarrow-core" available on conda.

Number of skipped tests stays the same...


brendan-ward (approved):
Thanks @theroggy

Issue #287, opened 2023-09-06
ENH: Add index_as_fid option to write_dataframe?


theroggy:
There is a `fid_as_index` option in `read_dataframe` to be able to... read the `fid`. However, there is no explicit option to e.g. ìndex_as_fid` in `write_dataframe` to write it again.

There is a workaround: if you add the index as a column named "fid" to the dataframe you will write the `fid`, at least for Geopackages. 

For e.g. shapefiles writing fid's is not relevant as they don't save fid's.

Not sure if an explicit option is needed... an alternative could be to just document the "workaround".

jorisvandenbossche:
Do you know if GDAL exposes some information or metadata about this? For example so that we can raise an error if the file format actually does not support it?

theroggy:
> Do you know if GDAL exposes some information or metadata about this? For example so that we can raise an error if the file format actually does not support it?

Yes, if [OGR_L_GetFIDColumn()](https://gdal.org/doxygen/classOGRLayer.html#af0cb9a05310d7b17d9dfec9e83a18e61)  returns "", the file type cannot save `fid`'s. If it returns another string, that's the column name used to store the `fid`. 

Examples: 
- for shapefile "" is returned. 
- for Geopackage "fid" is returned: that's the default name of the column the fid is stored in.
- for sqlite files I never tested/used it, but according to the gdal doc the default column name for the `fid` is "OGC_FID", so I suppose that `OGR_L_GetFIDColumn()` will return that and I suppose that you'll have to call the column like that in the GeoDataFrame if you want the `fid`'s to be retained.

jorisvandenbossche:
Nice, in that case this sounds as a good option to me.

theroggy:
I just found another way to save the index of the GeoDataFrame as "fid" (primary key) in a GPKG: if the name of the index is "fid" it will also be saved as "fid" in the file.

xref: https://github.com/geopandas/geopandas/issues/3217

jorisvandenbossche:
(moved this to the next milestone)

evetion:
Additionaly, one can pass `fid="custom_index_name"` to the kwargs of write_dataframe, and make any column the primary key in a GPKG. See https://gdal.org/drivers/vector/gpkg.html#layer-creation-options.

[...]

PR #511, opened 2024-12-20
BUG: fix writing fids to e.g. GPKG file with use_arrow

### LINKED ISSUE #510: BUG: write_dataframe to GPKG with an "fid" column gives an error if use_arrow=True
For some file types it is possible to write "custom" FID's by including an appropriate column in the input dataset, e.g. a column "fid" for GeoPackage.

However, when this is applied using `write_dataframe` with `use_arrow=True` this error is thrown: `Inconsistent values of FID and field of same name`.

An issue for this was opened upstream in GDAL: https://github.com/OSGeo/gdal/issues/11527

EDIT: the conclusions of the upstream issue was that:
1. there was a bug that lead to the error, which is now fixed
2. in pyogrio, the case when there is an "fid" column present isn't handled really well. For formats that explicitly store the FID (e.g. GPKG) columns with the "correct" names (e.g. "fid" for GPKG) are automatically used by GDAL for the FID field. But, in pyogrio, these columns are also added as an ordinary column, which is redundant. The double use of the column is also what triggers the bug in GDAL mentioned in 1., so if this behaviour is changed, the bug isn't triggered anymore: #511 

Originally noticed in https://github.com/geofileops/geofileops/pull/392

theroggy:
resolves #510

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -2741,6 +2741,16 @@ cdef create_fields_from_arrow_schema(
     IF CTE_GDAL_VERSION < (3, 8, 0):
         raise RuntimeError("Need GDAL>=3.8 for Arrow write support")
 
+    # Some formats store the FID explicitly in a real column, e.g. GPKG.
+    # For those formats, OGR_L_GetFIDColumn will return the column
+    # name used for this and otherwise it returns "". GDAL typically also provides a
+    # layer creation option to overrule the column name to be used as FID column.
+    # When a column with the appropriate name is present in the data, GDAL
+    # will automatically use it for the FID. Reference:
+    # https://gdal.org/en/stable/tutorials/vector_api_tut.html#writing-to-ogr-using-the-arrow-c-data-interface
+    # Hence, the column should not be created as an ordinary field well.
+    # Doing so anyway additionally triggers a bug in GDAL < 3.10.1:
+    # https://github.com/OSGeo/gdal/issues/11527#issuecomment-2556092722
     # The schema object is a struct type where each child is a column.
     cdef ArrowSchema* child
     for i in range(schema.n_children):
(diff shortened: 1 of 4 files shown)


pyogrio/_io.pyx
```diff
@@ -2752,6 +2752,20 @@ cdef create_fields_from_arrow_schema(
...
+        # For those formats, OGR_L_GetFIDColumn will return the default column
+        # name used and otherwise it returns "". The column name can also be
+        # explicitly choosed
+        # When a column with the appropriate name is present in the data, GDAL
+        # will automatically use it for the FID. Reference:
+        # https://gdal.org/en/stable/tutorials/vector_api_tut.html#writing-to-ogr-using-the-arrow-c-data-interface
+        # Hence, the field should not be created as well. Doing so anyway
+        # additionally triggers a bug in GDAL < 3.10.1:
+        # https://github.com/OSGeo/gdal/issues/11527#issuecomment-2556092722
+        fid_column = get_string(OGR_L_GetFIDColumn(destLayer))
```
jorisvandenbossche, line 2766: The reason this works is because `destLayer` knows that it is a layer for a gpkg driver and then actually returns "fid" ?

theroggy: Yes indeed. I added .sqlite to the tests to make this "explicit": for SQLite files the default column name is "ogc_fid"

theroggy:
I put the PR as draft while waiting for feedback on the comment here: https://github.com/OSGeo/gdal/issues/11527#issuecomment-2556840022

pyogrio/tests/test_geopandas_io.py
```diff
@@ -993,6 +993,48 @@ def test_write_csv_encoding(tmp_path, encoding):
...
+        kwargs["fid"] = fid_param_value
+    path = tmp_path / f"test{ext}"
+
+    write_dataframe(input_gdf, path, use_arrow=use_arrow, **kwargs)
+
+    assert path.exists()
+    output_gdf = read_dataframe(path, fid_as_index=True, use_arrow=use_arrow)
+    output_gdf = output_gdf.reset_index()
+
+    # pyogrio always sets "fid" as index name with `fid_as_index`
```
brendan-ward, line 1033: Question: should we be preserving the original custom name when uisng `fid_as_index`?

theroggy: This has been suggested/requested in this comment: https://github.com/geopandas/pyogrio/issues/287#issuecomment-2298222604

I don't mind, but it will be a breaking change...

brendan-ward: Probably fine to leave that for a different PR to deal with that specifically.

brendan-ward (approved):
Thanks @theroggy

PR #512, opened 2024-12-21
BUG: fix error in write_dataframe when writing an empty or all-None object column with use_arrow

### LINKED ISSUE #513: BUG: writing an empty dataframe with an object column with use_arrow fails
When an empty dataframe (no rows) with an "object" dtype column is written using `use_arrow=True`, this leads to this type of error: `pyogrio.errors.FieldError: Error while creating field from Arrow for field 3 with name 'HFDTLT' and type n (Type 'n' for field HFDTLT is not supported.)`.

The same problem occurs when an object type column contains only None values.

Looking deeper into this, the issue is that in the table returned by `pyarrow.Table.from_pandas`, the `object` column is returned as having the "null" type... and this "null" type is not supported by GDAL.

https://github.com/geopandas/pyogrio/blob/ed39ad75618a3b2ca97e46b2818cad8b306eb8bc/pyogrio/geopandas.py#L623

theroggy:
When a dataframe is being written with an object column without any rows or with only None values in the object column, the object column is converted to an null type arrow column, which is not supported by gdal and leads to an error being thrown.

Fixes #513

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -1157,9 +1157,6 @@ def test_write_dataframe_index(tmp_path, naturalearth_lowres, use_arrow):
 )
 @pytest.mark.requires_arrow_write_api
 def test_write_empty_dataframe(tmp_path, ext, columns, dtype, use_arrow):
-    if use_arrow and dtype is object:
-        pytest.xfail(reason="writing an empty object column with Arrow gives an error")
-
     expected = gp.GeoDataFrame(geometry=[], columns=columns, dtype=dtype, crs=4326)
 
     filename = tmp_path / f"test{ext}"
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -1157,6 +1157,11 @@ def test_write_dataframe_index(tmp_path, naturalearth_lowres, use_arrow):
 )
 @pytest.mark.requires_arrow_write_api
 def test_write_empty_dataframe(tmp_path, ext, columns, dtype, use_arrow):
+    """Test writing dataframe with no rows.
+
+    With use_arrow, object type columns with no rows are converted to null type columns
+    by pyarrow, but null columns are not supported by GDAL. Added to test fix for #513.
+    """
     expected = gp.GeoDataFrame(geometry=[], columns=columns, dtype=dtype, crs=4326)
     filename = tmp_path / f"test{ext}"
     write_dataframe(expected, filename, use_arrow=use_arrow)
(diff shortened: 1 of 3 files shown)


brendan-ward:
I'm not clear on what the proper fix is going to be in this case.  Should we instead raise our own error when writing an empty dataframe with one or more object dtype columns present using arrow, and direct user to the non-arrow interface?  Or should we fall back to using non-arrow ourselves when detecting an empty data frame?  No benefit of using arrow for this case.

theroggy:
> I'm not clear on what the proper fix is going to be in this case. Should we instead raise our own error when writing an empty dataframe with one or more object dtype columns present using arrow, and direct user to the non-arrow interface? Or should we fall back to using non-arrow ourselves when detecting an empty data frame? No benefit of using arrow for this case.

Yes, I didn't have a clear idea yet either... The thing I was wondering about (as indeicated in #513) was if it was a very conscious  in `pyarrow.Table.from_pandas` to convert an `object` column to `null` datatype, as all other datatypes (int,...) are retained as such. `object` is obviously a very special case, so I understand it is a different case compared to int,... but the `null` datatype doesn't seem super useful to me (I might be wrong)...

It is a good point however that arrow doesn't have a lot of added value if there is no data to be written, so it could be an easy fix to detect the dataframe being empty upfront and disabling use of arrow...

theroggy:
@brendan-ward I disabled `use_arrow` for empty dataframes so this works now...

However, I then wondered what the effect would be of a column with all None values... So I added a test for that, and this triggers also an error. There are probably workarounds possible for this case as well... but those will probably cost performance... so I wonder if it is an option that the `table.from_pandas` would be changed?

jorisvandenbossche:
> The thing I was wondering about was if it was a very conscious in `pyarrow.Table.from_pandas` to convert an `object` column to `null` datatype, as all other datatypes (int,...) are retained as such. `object` is obviously a very special case, so I understand it is a different case compared to int,... but the `null` datatype doesn't seem super useful to me (I might be wrong)...

This is a conscious choice, yes, AFAIK (although it was already like that before my involvement in pyarrow). For other data types in pandas like int, there is a clear equivalent in Arrow, and so it can be retained even for empty dataframes. But "object" dtype has no equivalent in Arrow, and thus the resulting arrow type always has to be "inferred" from the data when converting from pandas to arrow. However, if the column is empty or all-None, there is no data to infer .. At that point there is no ideal choice, but the benefit of using the "null" type is that it essentially does not make a choice, and the type is then not "viral" (if you would infer it as `string` instead, but it should actually have been something else, then you cannot combine the string column anymore with an int column, while with a null column this can still work).

Now in practice, given that object dtype in pandas is often used as strings, and given that GDAL does not support null and so we have to do some conversion to make it work, I think casting to string as you did now in the PR is indeed the best solution  (in practice, if you have an object dtype column, we also convert this to their string representation in the non-arrow write path anyway, so this makes that more consistent between both write paths)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -1146,16 +1146,29 @@ def test_write_dataframe_index(tmp_path, naturalearth_lowres, use_arrow):
...
+    """
+    expected = gp.GeoDataFrame(geometry=[], columns=columns, dtype=dtype, crs=4326)
     filename = tmp_path / f"test{ext}"
     write_dataframe(expected, filename, use_arrow=use_arrow)
 
     assert filename.exists()
-    df = read_dataframe(filename)
-    assert_geodataframe_equal(df, expected)
+    df = read_dataframe(filename, use_arrow=use_arrow)
+    assert_geodataframe_equal(df, expected, check_index_type=False)
```
jorisvandenbossche, line 1171: Any idea why the `check_index_type=False` is now needed while before it was not?

theroggy: I didn't remember, but had a quick look... Apparently for the tests running with an older version of python (or the according pandas version) the dataframe index is created as dtype Object, but read again as int (a range) so the index dtype check fails. For the other python version CI environments the Dataframe is created as int as well so the test passes.

jorisvandenbossche (approved):

PR #516, opened 2025-01-16
CI: also use gdal nightly in nightly-deps tests

theroggy:
Up to now, the nightly gdal is not used in the conda environment for this test. Probably better to also use the master/nightly version of GDAL in this test?

Remark: there is also a test using the latest gdal docker image... which is also typically a master/nightly version of gdal that is used.


brendan-ward (approved):
Seems OK to have a conda test that uses latest nightly GDAL in addition to the one in the  docker test.

PR #518, opened 2025-01-18
ENH: avoid redundant processing in write_dataframe with use_arrow

theroggy:
Preparing the data in numpy arrays is not needed for `use_arrow=True`, so move a block of code to avoid the unneeded processing.


brendan-ward (approved):
Good catch, thanks @theroggy

PR #519, opened 2025-01-27
MAINT: enable cython-lint + fix errors

theroggy:


jorisvandenbossche:
The cython-lint has no autofixing (like black or ruff), right? So this is one where we have to manually ensure the formatting is following the rules?

theroggy:
> The cython-lint has no autofixing (like black or ruff), right? So this is one where we have to manually ensure the formatting is following the rules?

Yes, only the conversion of single quoted strings to double quoted strings is automatic...

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks! Let's see how this goes with the manual fixing. If it gets too annoying, we can always turn it off, but this cleanup in itself is already nice

PR #528, opened 2025-03-24
TST: changes in tests to pass GDAL 3.11 (nightlies)

theroggy:
GDAL 3.11 retires quite some drivers... and apparently the .txt ext doesn't have multiple drivers anymore, so remove it from the test.


jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #529, opened 2025-03-30
CI: fix nightly-deps env so nightlies of all relevant dependencies are used

theroggy:
Because geopandas-base is being installed via conda-forge, all it's dependencies (numpy, geopandas, shapely) are also already being installed from standard conda-forge. Because they are already installed when pip kicks in, the nightlies from scientific-python-nightly-wheels aren't actually installed...

This PR changes the env so geopandas is also installed via pip, which solves above issue.


jorisvandenbossche:
I think this is probably not needed because we get the shapely nightly wheel through the `scientific-python-nightly-wheels` pypi index?

jorisvandenbossche:
Ah, but then we should fix that, because the same is true for numpy and pandas, so we are not actually installing nightly packages for those.

(also, adding the conda-forge shapely_dev label seems to give you shapely from that channel, although right now this is then actually an older version (2.1.0rc) than the latest ..)

So I think either we have to ensure we are not yet installing those dependencies in the base conda env (e.g. remove geopandas-base, and install geopandas from pip as well), or otherwise have to ensure the pip call upgrades (adding `-U`? although that might try to upgrade too much?)

jorisvandenbossche:
Looking at the yml file again, I assume removing `geopandas-base` and adding `geopandas` to the list of pip dependencies should solve the issue

theroggy:
> Ah, but then we should fix that, because the same is true for numpy and pandas, so we are not actually installing nightly packages for those.
> 
> (also, adding the conda-forge shapely_dev label seems to give you shapely from that channel, although right now this is then actually an older version (2.1.0rc) than the latest ..)

True.

I was focussing on shapely 2.1.0 rc when making this PR as I wanted the pyogrio tests to have run before shapely 2.1 was officially released... but the current fix is indeed not a long term solution.

Moving geopandas to being installed via pip as well should indeed fix it for everything...

jorisvandenbossche (approved):

jorisvandenbossche:
Looking good now!

PR #532, opened 2025-04-10
DOC/TST: document and test read_dataframe with on_invalid=fix

### LINKED ISSUE #423: ENH: add "repair" and/or "salvage" option to `on_invalid` parameter in `read_dataframe`
The `on_invalid` functionality in `read_dataframe` always results in None geometries now.

It would be interesting to also have a "repair" and/or "salvage" option, so it becomes possible to recover data if possible.

In theory this should be possible:
- *EDITED:* in GEOS, there is apparently a function [GEOSWKBReader_setFixStructure_r](https://libgeos.org/doxygen/geos__c_8h.html#a6c6c66ac349741cf9d3a3a79c6d6bb6f) that enables fixing such issues when parsing WKB's. Support to use this could be added to shapely.
- Or via another library, but this most likely implies extra dependencies which might not be ideal for an edge case like this. 
- Or an option to get back the WKB's instead of trying to read them, so you can salvage yourself
- ...

xref https://github.com/geopandas/pyogrio/pull/422

theroggy:
Document and test the extra option "fix" for the `on_invalid` parameter of `read_dataframe` made possible in shapely 2.1.

resolves #423

### DIFF
--- a/pyogrio/_compat.py
+++ b/pyogrio/_compat.py
@@ -46,3 +46,4 @@ GDAL_GE_38 = __gdal_version__ >= (3, 8, 0)
 HAS_GDAL_GEOS = __gdal_geos_version__ is not None
 
 HAS_SHAPELY = shapely is not None and Version(shapely.__version__) >= Version("2.0.0")
+SHAPELY_GE_21 = shapely is not None and Version(shapely.__version__) >= Version("2.1.0")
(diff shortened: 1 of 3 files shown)


brendan-ward (approved):
Looks good, thanks @theroggy 

(please resolve conflict then this is ready to merge)

PR #546, opened 2025-07-05
BUILD: quick check if LibKML and LibSpatialite can be included now

theroggy:


dg0yt:
Would it help to include libspatialite without freexl?

theroggy:
> Would it help to include libspatialite without freexl?

Yes, freexl is not needed for our use, so if that's possible that would be great!

dg0yt:
A little bit different. Add a libspatialite as a top-level dependency, with default-features: false.

dg0yt:
vcpkg linux logs are still reported to be in /opt/vcpkg/buildtrees - this is not what is uploaded in the next step.

theroggy:
> vcpkg linux logs are still reported to be in /opt/vcpkg/buildtrees - this is not what is uploaded in the next step.

Thanks for having a look! I'm not a specialist at all in this, so much appreciated!

I only made the change in one of the 3 linux containers/builds (manylinux_2_28_x86_64-vcpkg-gdal), so sorry for the confusion. I now changed it in all three...

But, they log files are not found anyway...

I wondered if it has something to do with paths in the container versus local paths, so I tried copying the files first to a local path,, but I don't seem to get that working either :-(...

dg0yt:
And yes. libspatialite needs freexl needs minizip.

PR #558, opened 2025-08-09
ENH: add support to read list fields without arrow

### LINKED ISSUE #17: ENH: Support list data types
From https://github.com/geopandas/geopandas/issues/2113

GDAL has data types for IntegerList, RealList, StringList (https://gdal.org/api/vector_c_api.html#_CPPv412OGRFieldType). Currently those are not supported, but in theory that could be mapped to a column with python lists (that is never going to be super performance, since it's an object ndarray with python lists, and it might also give some complexity to convert this in the cython code)

theroggy:
Add support for all list field types without arrow: list of strings, list of integers and list of double.

closes #17

reference https://github.com/geopandas/geopandas/issues/2113

### DIFF
--- a/pyogrio/_ogr.pxd
+++ b/pyogrio/_ogr.pxd
@@ -283,6 +283,12 @@ cdef extern from "ogr_api.h":
     int             OGR_F_GetFieldAsInteger(OGRFeatureH feature, int n)
     int64_t         OGR_F_GetFieldAsInteger64(OGRFeatureH feature, int n)
     const char*     OGR_F_GetFieldAsString(OGRFeatureH feature, int n)
+    char **         OGR_F_GetFieldAsStringList(OGRFeatureH feature, int n)
+    int *           OGR_F_GetFieldAsIntegerList(
+                        OGRFeatureH feature, int n, int* pnCount)
+    int64_t *       OGR_F_GetFieldAsInteger64List(
+                        OGRFeatureH feature, int n, int* pnCount)
+
     int             OGR_F_IsFieldSetAndNotNull(OGRFeatureH feature, int n)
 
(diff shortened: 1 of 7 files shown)


theroggy:
> Can you please expand the tests to include empty lists of each type, and expand the string test to include empty strings (maybe just an entry of one empty string)?

Good idea! Tests added!

brendan-ward (approved):
Thanks @theroggy

pyogrio/_ogr.pxd
```diff
@@ -283,6 +283,14 @@ cdef extern from "ogr_api.h":
     int             OGR_F_GetFieldAsInteger(OGRFeatureH feature, int n)
     int64_t         OGR_F_GetFieldAsInteger64(OGRFeatureH feature, int n)
     const char*     OGR_F_GetFieldAsString(OGRFeatureH feature, int n)
+    char **         OGR_F_GetFieldAsStringList(OGRFeatureH feature, int n)
+    int *           OGR_F_GetFieldAsIntegerList(
+                        OGRFeatureH feature, int n, int* pnCount)
+    int64_t *       OGR_F_GetFieldAsInteger64List(
```
brendan-ward, line 289: Unfortunately, it looks like marking this as `int64_t` is failing on at some platforms:

```
error: assignment to ‘int64_t *’ {aka ‘long int *’} from incompatible pointer type ‘const GIntBig *’ {aka ‘const long long int *’} [-Wincompatible-pointer-types]
```

[error here](https://github.com/geopandas/pyogrio/actions/runs/17311694459/job/49146952701#step:5:805)

Unfortunately not caught by our tests, only when building the wheels.

theroggy: Yes, I saw it as well... I'm solving it...

PR #559, opened 2025-08-09
ENH: add whether append is supported in list_drivers

### LINKED ISSUE #375: Show append support in list_drivers
Fiona shows which drivers support appending in `fiona.supported_drivers`. Is there a way for pyogrio to do the same?

```py
In [1]: import fiona

In [2]: fiona.supported_drivers
Out[4]: 
{'DXF': 'rw',
 'CSV': 'raw',
 'OpenFileGDB': 'raw',
 'ESRIJSON': 'r',
 'ESRI Shapefile': 'raw',
 'FlatGeobuf': 'raw',
 'GeoJSON': 'raw',
 'GeoJSONSeq': 'raw',
 'GPKG': 'raw',
 'GML': 'rw',
 'OGR_GMT': 'rw',
 'GPX': 'rw',
 'Idrisi': 'r',
 'MapInfo File': 'raw',
 'DGN': 'raw',
 'PCIDSK': 'raw',
 'OGR_PDS': 'r',
 'S57': 'r',
 'SQLite': 'raw',
 'TopoJSON': 'r'}
```

theroggy:
`list_drivers` lists all drivers, and specifies for each format if is is read-only or read-write ("r" vs. "rw"). This PR adds another flag in the output to indicate whether appending is supported or not ("r" vs "raw" vs "rw"), like it is the case in fiona.

This uses features introduced in GDAL 3.11/3.12, so the "a" is also only available when minimum that version of GDAL is used: https://github.com/OSGeo/gdal/issues/12899#event-19350983841

There are a few differences compared to the hardcoded list in fiona, but it seems to me that most are improvements. Check out `test_list_drivers` for details.

**Remark**: 
- at the moment I expect that this won't be finalized/reviewed in time for version 0.12. If it does get finalized/reviewed in time... these need to be updated to 0.12, the `versionadded` and `versionchanged` directives + changelog intre need to be changed to 0.12

resolves https://github.com/geopandas/pyogrio/issues/375


pyogrio/_ogr.pyx
```diff
@@ -121,14 +139,20 @@ def ogr_list_drivers():
...
 
-        if ogr_driver_supports_write(name):
-            drivers[name] = "rw"
+        capability = "r"
 
+        if ogr_driver_supports_update(name):
+            capability += "a"
         else:
-            drivers[name] = "r"
+            if ogr_driver_supports_append(name):
```
jorisvandenbossche: Combine the `else:  if:  ` into a `elif`?

pyogrio/tests/test_core.py
```diff
@@ -137,16 +137,63 @@ def test_ogr_driver_supports_write(driver, expected):
...
+    }
+
     # verify that the core drivers are present
-    for name in ("ESRI Shapefile", "GeoJSON", "GeoJSONSeq", "GPKG", "OpenFileGDB"):
-        assert name in all_drivers
-        expected_capability = "rw"
-        assert all_drivers[name] == expected_capability
+    for name, expected_capability in expected_drivers.items():
+        if name not in all_drivers:
+            print(f"{name} not in list_drivers(), ignore")
```
jorisvandenbossche: Not sure we want to keep this `print` in here?

pyogrio/tests/test_core.py
```diff
@@ -137,16 +137,63 @@ def test_ogr_driver_supports_write(driver, expected):
...
+            continue
+
+        if not GDAL_GE_312:
+            expected_capability = expected_capability.replace("a", "")
+        if name == "OpenFileGDB" and __gdal_version__ < (3, 6, 0):
```
jorisvandenbossche: We don't support GDAL < 3.6 anymore, I think, in which case this can be removed

jorisvandenbossche (commented):
Looks good!

jorisvandenbossche:
Doesn't necessarily need to happen in this PR, but just to remember for myself: with new capability check, we can try to improve the error message as mentioned in https://github.com/geopandas/pyogrio/issues/375#issuecomment-2003468022

theroggy:
@jorisvandenbossche in the context of eg. https://github.com/geopandas/pyogrio/issues/647, I was wondering if it might be a good idea to create a `list_drivers2` (or any other name ;-)) that returns the drivers as a dict of dicts with more properties of a driver, eg.
``` python
{
    "GPKG": {
        "long_name": "Geopackage",
        "access_modes": "raw",
    }
}
```

Or a seperate field per access mode being supported or not or whatever... but a structure having the option to extend the number of driver properties being returned.

If we would go for something like this, it might be better to move adding support for the "append" access mode to there, as this avoids this non-backwards compatible PR change to `list_drivers`?

jorisvandenbossche:
That sounds indeed as a good idea. I would either go for a keyword like `details=True` in the existing function, or add a `pyogrio.list_drivers_details()` or something alike. 

For that case we could also keep it closer to the GDAL terminology when it comes to the topic of this PR, i.e. the update vs append? (or let's discuss that on the relevant issue or PR)

theroggy:
Closed because it has been replaced by #656.

pyogrio/tests/test_core.py
theroggy: True.

pyogrio/tests/test_core.py
theroggy: oops, removed.

pyogrio/tests/test_core.py
```diff
@@ -139,16 +139,65 @@ def test_ogr_driver_supports_write(driver, expected):
 def test_list_drivers():
     all_drivers = list_drivers()
 
+    # Expected capabilities based on `fiona.supported_drivers`.
```
jorisvandenbossche, line 142: FWIW, if this turns out to be too much of a hassle to test (i.e. if we need a lot of gdal version checks for to cover changes), I think we could also just test a small "known" subset as before. But let's merge now as is, and we can always go back later.

jorisvandenbossche (approved):
Thanks!

PR #561, opened 2025-08-10
BUIlD: add libkml to builds for recent platforms

theroggy:


ci/vcpkg2014.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "pyogrio",
+    "version": "0.11.0",
```
brendan-ward, line 3:
```suggestion
    "version": "0.12.0",
```

brendan-ward (commented):
Thanks for working on this @theroggy !  Only supporting on >= manylinux_2_28 seems reasonable to me.  Presumably we are going to drop manylinux2014 at some point in the not-too-distant future.

brendan-ward (approved):
Thanks @theroggy

PR #562, opened 2025-08-11
ENH: Compile free-threading wheels

benoit9126:
I propose to activate the free-threading compilation of cibuildwheel.


jorisvandenbossche:
@benoit9126 apologies for the slow follow-up? Would you like to update this with the latest main? (some changes to update cibuildwheel and docker have already been done separately in the meantime) 
Then I think we can include it for the next 0.12 release

benoit9126:
@jorisvandenbossche No problem. I have rebased my work. A single test failed but as far as I can understand, it is not related to my modifications.

jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
Checking the output, it seems we will need to install nightly pandas to test free-threaded (it gives the message the GIL has been enabled for importing pandas)

In the "Install dependencies" step, can you add `pip install --pre --extra-index https://pypi.anaconda.org/scientific-python-nightly-wheels/simple pandas` when the python version is 3.14t ?

(I think the other if cases for 3.14 can actually be simplified (removed) now)

.github/workflows/release.yml
```diff
@@ -313,15 +313,12 @@ jobs:
...
-            uv pip install -r ci/requirements-wheel-test.txt
-          else
-            uv pip install pytest numpy certifi packaging
-          fi
+          uv pip install -r ci/requirements-wheel-test.txt
           uv pip install --no-cache --pre --no-index --find-links wheelhouse pyogrio
-          if [ ${{ matrix.python-version }} != "3.14" ]; then
-            uv pip install --no-deps geopandas
+          if [ ${{ matrix.python-version }} == "3.14t" ]; then
+            uv pip install --pre --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple pandas
```
jorisvandenbossche:
```suggestion
            uv pip install --pre --upgrade --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple pandas
```

This might be needed since pandas is already installed, to ensure the nightly version actually gets installed

benoit9126:
Yes, it needs to be `--extra-index-url` because it is not a `pip` command but an `uv pip` command.

jorisvandenbossche:
Tests are looking good now, Thanks a lot @benoit9126!

PR #568, opened 2025-08-26
BUG: fix decode error reading an sqlite file on windows

### LINKED ISSUE #567: BUG: reading a sqlite file with text data with special characters gives encoding errors on Windows
The following error is being thrown when doing `read_dataframe()` on an sqlite file with text columns with special characters in them:
```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 2: character maps to <undefined>
decoding with 'cp1252' codec failed
```

If the parameter `encoding="UTF-8"` is passed the problem does not occur.

Cleanest (long term) solution would probably be that the GDAL `TestCapability()` function returns True for `OLCStringsAsUTF8`.
Issue will be opened in the GDAL issue tracker.

Script to reproduce on windows:
``` python
"""Test."""

import locale
import sqlite3
import sys
from pathlib import Path

import pandas as pd

import pyogrio

script_path = Path(__file__)


df = pd.DataFrame({"name": ["foö", "bÏr"], "value": [1, 2]})

path = script_path.parent / "test.sqlite"
if path.exists():
    path.unlink()
con = sqlite3.connect(str(path))
df.to_sql(name="test", index=True, con=con)
con.close()
print(pyogrio.read_info(path)["encoding"])

sqlite_path = script_path.with_suffix(".sqlite")
columns = pyogrio.read_info(sqlite_path)["fields"]
print(f"{locale.getpreferredencoding()=}")
print(f"{locale.getencoding()=}")
print(f"{sys.flags.utf8_mode=}")
print(pyogrio.read_info(sqlite_path)["encoding"])

df = pyogrio.read_dataframe(path)
```

theroggy:
Closes #567

### DIFF
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -609,7 +609,7 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
     if driver == "GeoJSONSeq":
         # In old gdal versions, OLCStringsAsUTF8 wasn't advertised yet.
         return "UTF-8"
-    
+
     if driver == "SQLite":
         # In gdal versions < 3.12, OLCStringsAsUTF8 wasn't advertised yet.
         return "UTF-8"
--- a/pyogrio/_io.pyx
+++ b/pyogrio/_io.pyx
@@ -612,7 +612,8 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
         return "UTF-8"
 
     if driver == "SQLite":
-        # In gdal versions < 3.12, OLCStringsAsUTF8 wasn't advertised yet.
+        # TestCapability for OLCStringsAsUTF8 returns False for SQLite in GDAL 3.11.3.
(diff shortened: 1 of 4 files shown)


pyogrio/_io.pyx
```diff
@@ -611,6 +611,10 @@ cdef detect_encoding(OGRDataSourceH ogr_dataset, OGRLayerH ogr_layer):
         # In old gdal versions, OLCStringsAsUTF8 wasn't advertised yet.
         return "UTF-8"
 
+    if driver == "SQLite":
+        # In gdal versions < 3.12, OLCStringsAsUTF8 wasn't advertised yet.
```
brendan-ward, line 615: Is this fixed in GDAL as of 3.12?   I didn't see a fix referenced in your upstream GDAL issue.  Maybe for now just put in the link to https://github.com/OSGeo/gdal/issues/12962 and later we'll update that to the correct version and add a version check around this.

theroggy: Indeed, I was a bit optimistic :-).

PR #572, opened 2025-09-04
ENH: unlock the gil during GDAL functions that can take significant time

theroggy:
Some GDAL functions (e.g. executing SQL statements,...) can take a while, so nogil-ling those can already give significant performance improvements when using multithreading.

Remark: determining exactly which GDAL functions take time is not trivial as this can depend on the file format,... For files types with a stricter data schema (e.g. shapefile,...) determining column types is fast, for some other file types (e.g. geojson) the file needs to be read entirely first to determine all data types,.... Hence, a relatively large subset of GDAL calls have been no-gilled to avoid missing significant situations.


pyogrio/_io.pyx
```diff
@@ -464,19 +483,25 @@ cdef get_feature_count(OGRLayerH ogr_layer, int force):
...
-        OGR_L_ResetReading(ogr_layer)
+        with nogil:
+            OGR_L_ResetReading(ogr_layer)
 
         feature_count = 0
         while True:
             try:
-                ogr_feature = check_pointer(OGR_L_GetNextFeature(ogr_layer))
+                with nogil:
+                    ogr_feature = OGR_L_GetNextFeature(ogr_layer)
```
jorisvandenbossche: Since we are in a loop here for counting the features, this means for a file with 10,000 rows, we are going to release/acquire the GIL 10,000 times? 
(if I am not misreading the code, that might not be optimal)

jorisvandenbossche:
> I reran my performance tests

Thanks for that script (and sorry for the slow follow-up). Using that script, I ran some tests here as well (on a Linux). 

For filter resulting in reading only a few rows, I see similar results as you (just generally a bit faster), and there the parallelization clearly gives a speedup with this branch.

But then I tried it with a different filter that would read much more data, using `where = "{filter} ST_NPOINTS(geom) < 10"` (instead of `> 500`), and then I see:

With this branch:
```
$ python bench_nogil.py 
************ with use_arrow=True *******************
len(df)=2450558, with sql, use_arrow=True took 5.245021343987901
len(df)=2450558, with where, use_arrow=True took 5.515946925967
read, lens=[644326, 587358, 585793, 633081] sequential, sql, use_arrow=True, took=5.735124277998693
read, lens=[644326, 587358, 585793, 633081] sequential, where, use_arrow=True, 4.991556770983152
read, lens=[644326, 587358, 585793, 633081], parallel, sql, use_arrow=True, took=3.546093443990685
read, lens=[644326, 587358, 585793, 633081], parallel, where, use_arrow=True, took=4.354040208039805
************ with use_arrow=False *******************
len(df)=2450558, with sql, use_arrow=False took 25.786610602983274
len(df)=2450558, with where, use_arrow=False took 26.817800754972268
read, lens=[644326, 587358, 585793, 633081] sequential, sql, use_arrow=False, took=26.75142541900277
read, lens=[644326, 587358, 585793, 633081] sequential, where, use_arrow=False, 26.243713215983007
read, lens=[644326, 587358, 585793, 633081], parallel, sql, use_arrow=False, took=33.68034533999162
read, lens=[644326, 587358, 585793, 633081], parallel, where, use_arrow=False, took=35.566779381013475
```

And on main:

```
$ python bench_nogil.py 
************ with use_arrow=True *******************
len(df)=2450558, with sql, use_arrow=True took 5.182760039984714
len(df)=2450558, with where, use_arrow=True took 5.551710308005568
read, lens=[644326, 587358, 585793, 633081] sequential, sql, use_arrow=True, took=5.625802075956017
read, lens=[644326, 587358, 585793, 633081] sequential, where, use_arrow=True, 5.042418016004376
read, lens=[644326, 587358, 585793, 633081], parallel, sql, use_arrow=True, took=3.391982380999252
read, lens=[644326, 587358, 585793, 633081], parallel, where, use_arrow=True, took=4.339610085007735
************ with use_arrow=False *******************
len(df)=2450558, with sql, use_arrow=False took 25.491990304959472
len(df)=2450558, with where, use_arrow=False took 26.391454448981676
read, lens=[644326, 587358, 585793, 633081] sequential, sql, use_arrow=False, took=26.625511340040248
read, lens=[644326, 587358, 585793, 633081] sequential, where, use_arrow=False, 25.869791070988867
read, lens=[644326, 587358, 585793, 633081], parallel, sql, use_arrow=False, took=27.369246470043436
read, lens=[644326, 587358, 585793, 633081], parallel, where, use_arrow=False, took=27.604114448011387
```

So first, parallellizing in this case using `use_arrow=False` does not give any speedup in neither case. But while on main it is only like 1s slower than sequential, with this PR it goes to 7-8s slower.

Now, this is maybe a less typical use case for using a sql/where filter, and with a filter that excludes more rows, it does give a speedup .. But so it is not entirely black and white.

Commenting out the single `nogil` around `OGR_L_GetNextFeature` in the while loop of `get_features` does make this branch go back to the similar performance characteristics of main (i.e. no parallel speedup, but also no significant slowdown):

```
************ with use_arrow=False *******************
len(df)=2450558, with sql, use_arrow=False took 26.224873373052105
len(df)=2450558, with where, use_arrow=False took 26.845924963010475
read, lens=[644326, 587358, 585793, 633081] sequential, sql, use_arrow=False, took=26.65382137801498
read, lens=[644326, 587358, 585793, 633081] sequential, where, use_arrow=False, 26.249538164993282
read, lens=[644326, 587358, 585793, 633081], parallel, sql, use_arrow=False, took=26.100831627030857
read, lens=[644326, 587358, 585793, 633081], parallel, where, use_arrow=False, took=26.970447928993963
```

And then when going back to the stricter filter, it still gives the speedup in those cases:

```
************ with use_arrow=False *******************
len(df)=16, with sql, use_arrow=False took 1.5636954689980485
len(df)=16, with where, use_arrow=False took 1.587898463010788
read, lens=[9, 0, 4, 3] sequential, sql, use_arrow=False, took=1.1720041320077144
read, lens=[9, 0, 4, 3] sequential, where, use_arrow=False, 1.1913723179604858
read, lens=[9, 0, 4, 3], parallel, sql, use_arrow=False, took=0.4037232000264339
read, lens=[9, 0, 4, 3], parallel, where, use_arrow=False, took=0.6050780219957232
```

theroggy:
> Thanks for that script (and sorry for the slow follow-up). Using that script, I ran some tests here as well (on a Linux).
For filter resulting in reading only a few rows, I see similar results as you (just generally a bit faster), and there the parallelization clearly gives a speedup with this branch.
But then I tried it with a different filter that would read much more data, using where = "{filter} ST_NPOINTS(geom) < 10" (instead of > 500), and then I see:

OK... Odd that reading without filter doesn't seem to have this impact.

Anyway, I removed the `with nogil`s in the feature loops...

jorisvandenbossche:
> Odd that reading without filter doesn't seem to have this impact.

FWIW I also see this slowdown when reading the full file (i.e. without specifying any filter) several times in parallel

jorisvandenbossche (approved):

PR #577, opened 2025-09-13
PERF: improve performance of read_dataframe if a filter is used

theroggy:
In `read_dataframe` without arrow, the number of rows of the result was counted first, and then the full data was read.

Especially when using a filter, counting the rows can take significant time.

This PR avoids doing the rowcount before reading to improve performance by reading the results in chunks till no more rows are returned from the dataset. For the new zealand building outlines geopackage (3.3 million rows) this gives these results:
  - If the filter limits the rows a lot counting the rows can even take the same time as the subsequent reading of all data... so in this case the time taken ~halves.
      - e.g. reading the test file file with `where="ST_NPOINTS(st_buffer(geom, 10)) > 2000"` (returning 9 rows) took 82 s, now 45 s.
  - When reading the entire file without filter, both implementations take 55-60 seconds on my windows laptop (plugged in), with the chunked implementation giving the same average timings. To be sure, if no filter is used and counting rows is fast for the file type involved, no chunking is used (=old way).


pyogrio/_io.pyx
```diff
@@ -1294,34 +1302,81 @@ cdef get_features(
                     "encountering this error."
                 ) from None
 
+            if i_total == 0 or i == last_chunk_size:
+                # No chunk yet or last chunk filled up: add a new chunk.
+                i = 0
+                last_chunk_index += 1
+                if num_features < 0 and last_chunk_size < max_chunk_size:
+                    # Double chunk size for next chunk, up to max_chunk_size
+                    last_chunk_size = min(last_chunk_size * 2, max_chunk_size)
```
jorisvandenbossche, line 1356: This means the first chunks essentially starts at 2000, and not 1000? (not that it matters though)

pyogrio/_io.pyx
```diff
@@ -1230,49 +1230,54 @@ cdef get_features(
...
-    else:
-        fid_data = None
-
-    if read_geometry:
-        geometries = np.empty(shape=(num_features, ), dtype="object")
-        geom_view = geometries[:]
+    # We don't (always) know beforehand how many features there will be read, so
+    # allocate output arrays as needed (in chunks), and return the concatenated results.
+    last_chunk_size = 1_000
+    max_chunk_size = 50_000
```
jorisvandenbossche: Again not really that it will matters much, but if we have to pick a somewhat random max number here, we could go with the same batch size as the arrow interface: 65,536 (2**16)

pyogrio/_io.pyx
```diff
@@ -1689,9 +1744,14 @@ def ogr_read(
...
-            skip_features, num_features = validate_feature_range(
-                ogr_layer, skip_features, max_features
-            )
+            num_features = -1
+            if max_features > 0:
+                num_features = max_features
+
+            if skip_features > 0:
+                skip_features, num_features = validate_feature_range(
+                    ogr_layer, skip_features, max_features
```
jorisvandenbossche, line 1805: The reason we actually do have to perform the count in case of `skip_features`, is that because we might otherwise get errors if `skip_features` is larger than the actual number of features? (in which case this helper method returns `0, 0` to ensure we actually don't read anything)

jorisvandenbossche (commented):
I don't know if it is worth the added complexity, but in theory the following optimization might be possible: check if the driver supports fast count, and in the case there are no filters, that count should be a cheap lookup (like GPKG without any filtering), and at that point actually populate `num_features`. That would avoid using the growing chunk size in those cases (and we could directly read the entire data or directly use the max chunk size)

pyogrio/_io.pyx
theroggy: OK, I changed it to 1000 as you seem to have a preference for that.

pyogrio/_io.pyx
theroggy: Yes indeed, setting skip_features with a value larger than the rows that can be returned gives a GDAL error.

theroggy:
> I don't know if it is worth the added complexity, but in theory the following optimization might be possible: check if the driver supports fast count, and in the case there are no filters, that count should be a cheap lookup (like GPKG without any filtering), and at that point actually populate `num_features`. That would avoid using the growing chunk size in those cases (and we could directly read the entire data or directly use the max chunk size)

According to my tests when I wrote the PR, I didn't measure a significant difference between chunking or not for the case where no filtering was done, but it is not difficult to shortcut this case, so I added it.

pyogrio/_io.pyx
```diff
@@ -1689,9 +1744,26 @@ def ogr_read(
...
             # Limit feature range to available range
-            skip_features, num_features = validate_feature_range(
-                ogr_layer, skip_features, max_features
-            )
+            num_features = -1
+            if max_features > 0:
+                num_features = max_features
+
+            if skip_features > 0:
+                skip_features, num_features = validate_feature_range(
```
jorisvandenbossche, line 1804:
```suggestion
                # When skipping features, we need to validate the value against the
                # actual count to avoid out-of-bound index in apply_skip_features
                skip_features, num_features = validate_feature_range(
```

pyogrio/_io.pyx
```diff
@@ -1689,9 +1744,26 @@ def ogr_read(
                 apply_geometry_filter(ogr_layer, mask)
 
             # Limit feature range to available range
```
jorisvandenbossche:
```suggestion
            # Limit feature range to available range, but avoid doing an actual count if
            # not needed (with num_features=-1, we will read in chunks in get_features)
```

jorisvandenbossche (approved):
Looks good! Just two suggestions for clarifying comments for the future reader

jorisvandenbossche:
(and needs to fix the merge conflicts)

Issue #594, opened 2025-11-15
Support GDAL 3.12


ljstrnadiii:
The docs say that up to 3.11 is supported, but it would be nice to stay up to date with the newest GDAL.

I haven't looked too closely, but what is the level of effort to expand support for GDAL 3.12?

ljstrnadiii:
I assumed that conda/pixi wouldn't lock based on the 3.11 note in the docs. I'll give it a go!

I actually don't need any new ogr functionality available in 3.12, but I would like to get other features for new raster capabilities outside the pyogrio context eg rasterio currently needs <3.11.

That said, are you aware of what breaks between pyogrio and the newest GDAL version?

PR #604, opened 2025-11-20
DOC: Update versions in readme/docs + small improvements

theroggy:
Update versions in readme/docs + small improvements:
- remove max version of python and GDAL: triggered by #594, I think it is easier not to mention to **_highest_** version of python/GDAL being supported in the readme/docs, but only the minimal version. It is impractical to keep them up-to-date and in general the most recent version is supported in a timely manner. Shapely and geopandas also don't mention these in their docs...
- clarify that the GDAL version in the wheels is not always the most recent version + how you can check the version
- fix so special attributes like `__version__`, `__gdal_version__`,... are now listed in API reference + can be linked to
- change pyogrio functions being mentioned in `install.md` and `introduction.md` to cross-links to the API docs of the functions
- the environments.yml file in docs pinned very old versions -> update to recent versions

### DIFF
--- a/.github/workflows/docker-gdal.yml
+++ b/.github/workflows/docker-gdal.yml
@@ -44,7 +44,7 @@ jobs:
       #     add-apt-repository -y ppa:deadsnakes/ppa
       #     apt-get update && apt-get install -y python3.9-dev
 
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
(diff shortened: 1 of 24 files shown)


jorisvandenbossche (commented):
Thanks, looks good!

jorisvandenbossche (approved):

PR #612, opened 2025-11-26
BUG: fix error in read dataframe with use arrow and columns filter

### LINKED ISSUE #611: BUG: read_dataframe with use_arrow and columns gives errors in some cases
When using `read_dataframe` with `use_arrow=True` on a file containing list-type or datetime columns and (on of) the list-type or datetime column is not specified in `columns`, an error occurs.

theroggy:
When calling `read_dataframe` with `use_arrow=True` on a file containing `list-type`, `json` or `datetime` columns and (one of) the columns of these types is not specified to be read in `columns`, an error is thrown.

The underlying reason is that the `meta` object returned by `read_arrow` contains all fields instead of only the ones requested in the `columns` parameter... Hence, when looping over all fields in `meta["fields"]`... some of those fields haven't been read to the DataFrame... so an error is thrown trying to access those columns.

This PR makes sure only fields that have been read are returned + adds appropriate tests.

While writing the tests, another, unrelated problem was encountered. An issue for this was created here: #613 

Resolves #611

### DIFF
--- a/pyogrio/tests/test_raw_io.py
+++ b/pyogrio/tests/test_raw_io.py
@@ -159,9 +159,9 @@ def test_read_no_geometry_no_columns_no_fids(naturalearth_lowres):
     "descr, columns, exp_columns",
     [
         ("all", None, ["pop_est", "continent", "name", "iso_a3", "gdp_md_est"]),
-        ("case_sensitive", ["NAME", "NAME_LONG"], []),
+        ("case_sensitive", ["NAME"], []),
         ("repeats_dropped", ["continent", "continent", "name"], ["continent", "name"]),
-        ("keep_original_order", ["pop_est", "continent"], ["pop_est", "continent"]),
+        ("keep_original_order", ["continent", "pop_est"], ["pop_est", "continent"]),
     ],
 )
 def test_read_columns(naturalearth_lowres, descr, columns, exp_columns):
--- a/pyogrio/tests/test_arrow.py
+++ b/pyogrio/tests/test_arrow.py
@@ -231,9 +231,9 @@ def test_open_arrow_batch_size(naturalearth_lowres):
     "descr, columns, exp_columns",
     [
         ("all", None, ["pop_est", "continent", "name", "iso_a3", "gdp_md_est"]),
-        ("case_sensitive", ["NAME", "NAME_LONG"], []),
+        ("case_sensitive", ["NAME"], []),
         ("repeats_dropped", ["continent", "continent", "name"], ["continent", "name"]),
-        ("keep_original_order", ["pop_est", "continent"], ["pop_est", "continent"]),
+        ("keep_original_order", ["continent", "pop_est"], ["pop_est", "continent"]),
     ],
 )
 def test_open_arrow_columns(naturalearth_lowres, descr, columns, exp_columns):


pyogrio/_io.pyx
```diff
@@ -1722,6 +1722,11 @@ def ogr_open_arrow(
         if columns is not None:
             # Fields are matched exactly by name, duplicates are dropped.
             ignored_fields = list(set(fields[:, 2]) - set(columns))
+
+            # Find index of each field in columns, and only keep those
+            idx = np.intersect1d(fields[:, 2], columns, return_indices=True)[1]
```
jorisvandenbossche: I think the intersect1d might give a sorted result. While this will not be problematic I assume in practice (we loop over it, but still using the name and not position in the list), let's maybe preserve the order in the file.

jorisvandenbossche (commented):
Looks good!

pyogrio/_io.pyx
jorisvandenbossche:
```suggestion
            idx = np.sort(np.intersect1d(fields[:, 2], columns, return_indices=True)[1])
```

might be sufficient?

pyogrio/tests/test_raw_io.py
```diff
@@ -155,19 +155,20 @@ def test_read_no_geometry_no_columns_no_fids(naturalearth_lowres):
...
-    # Repeats should be dropped
-    columns = ["NAME", "NAME_LONG", "NAME"]
-    meta, _, geometry, fields = read(
+@pytest.mark.parametrize(
+    "descr, columns, exp_columns",
+    [
+        ("all", None, ["pop_est", "continent", "name", "iso_a3", "gdp_md_est"]),
+        ("case_sensitive", ["NAME", "NAME_LONG"], []),
+        ("repeats_dropped", ["continent", "continent", "name"], ["continent", "name"]),
+        ("keep_original_order", ["pop_est", "continent"], ["pop_est", "continent"]),
```
jorisvandenbossche:
```suggestion
        ("keep_original_order", ["continent", "pop_est"], ["pop_est", "continent"]),
```

Then it tests both that it doesn't sort alphabetically and doesn't follow the order of the passed list

pyogrio/tests/test_raw_io.py
```diff
@@ -155,19 +155,20 @@ def test_read_no_geometry_no_columns_no_fids(naturalearth_lowres):
...
-    array_equal(meta["fields"], columns)
-
-    # Repeats should be dropped
-    columns = ["NAME", "NAME_LONG", "NAME"]
-    meta, _, geometry, fields = read(
+@pytest.mark.parametrize(
+    "descr, columns, exp_columns",
+    [
+        ("all", None, ["pop_est", "continent", "name", "iso_a3", "gdp_md_est"]),
+        ("case_sensitive", ["NAME", "NAME_LONG"], []),
```
jorisvandenbossche:
```suggestion
        ("case_sensitive", ["NAME"], []),
```

("name_long" is not in the columns anyway, so just a bit confusing to include here. This is a leftover from long ago when the raw naturalearth data was used which has capital names and did have a NAME_LONG column .. which wasn't catched because of the missing assert)

jorisvandenbossche (approved):
Thanks!

PR #617, opened 2025-11-27
BUG: fix Time type columns being skipped with arrow=False

### LINKED ISSUE #615: BUG: columns with datatype time are ignored if arrow=False
When reading a file with columns of the type "Time", they are ignored/skipped when reading the file.

theroggy:
resolves #615 
reference  https://github.com/geopandas/geopandas/issues/3787

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -551,8 +551,6 @@ def test_read_many_data_types_geojson_file(many_data_types_geojson_file, use_arr
         else:
             assert result["date_col"].to_list() == [pd.Timestamp("2020-01-01")]
 
-    # Time columns are ignored without arrow:
-    # Reported in https://github.com/geopandas/pyogrio/issues/615
     assert "time_col" in result.columns
     assert is_object_dtype(result["time_col"].dtype)
     assert result["time_col"].to_list() == [time(12, 0, 0)]
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -3,7 +3,7 @@ import locale
 import os
 import re
 import warnings
-from datetime import datetime
+from datetime import datetime, time
 from io import BytesIO
 from zipfile import ZipFile
 
@@ -23,6 +23,7 @@ from pyogrio._compat import (
     GDAL_GE_311,
     HAS_ARROW_WRITE_API,
     HAS_PYPROJ,
+    NUMPY_GE_20,
     PANDAS_GE_15,
     PANDAS_GE_23,
     PANDAS_GE_30,
@@ -51,7 +52,14 @@ try:
     import geopandas as gp
     import pandas as pd
     from geopandas.array import from_wkt
-    from pandas.api.types import is_datetime64_dtype, is_object_dtype, is_string_dtype
+    from pandas.api.types import (
+        is_bool_dtype,
+        is_datetime64_dtype,
+        is_float_dtype,
+        is_integer_dtype,
+        is_object_dtype,
+        is_string_dtype,
+    )
 
     import shapely  # if geopandas is present, shapely is expected to be present
     from shapely.geometry import Point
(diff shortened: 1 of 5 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -615,22 +616,21 @@ def validate_result(df: pd.DataFrame, use_arrow: bool, ignore_mixed_list_col=Fal
...
     read_gdf = read_dataframe(many_data_types_geojson_file, use_arrow=use_arrow)
-    validate_result(read_gdf, use_arrow)
+    validate_result(read_gdf, use_arrow, after_write=False)
 
     # Write the data read, read it back, and validate again
     if use_arrow:
         # Writing a column with mixed types in a list is not supported with Arrow.
-        ignore_mixed_list_col = True
         read_gdf = read_gdf.drop(columns=["list_mixed_col"])
     else:
```
jorisvandenbossche: With the other PR merged, this would now be supported but converts it to all strings?

jorisvandenbossche (commented):
Actual change looks good! The tests will need some merge conflicts solved probably

pyogrio/tests/test_geopandas_io.py
```diff
@@ -599,54 +600,55 @@ def validate_result(df: pd.DataFrame, use_arrow: bool, ignore_mixed_list_col=Fal
...
+            assert is_object_dtype(df["time_col"].dtype)
+            assert df["time_col"].to_list() == [time(12, 0, 0)]
 
         assert "datetime_col" in df.columns
         assert is_datetime64_dtype(df["datetime_col"].dtype)
         assert df["datetime_col"].to_list() == [pd.Timestamp("2020-01-01T12:00:00")]
 
         assert "list_int_col" in df.columns
-        assert is_object_dtype(df["list_int_col"].dtype)
-        assert df["list_int_col"][0].tolist() == [1, 2, 3]
```
jorisvandenbossche, line 615: The reason this was "working" before is because the non-arrow case was just getting xfailed?

theroggy: @jorisvandenbossche Yes, indeed...

By adding more detailed checks now I noticed the lists are being returned in the following format, so without comma's between the elements:
`assert df["list_int_col"][0] == "[1 2 3]"`

I need to look a bit further to be certain, but at first sight it seems that GDAL is returning the list columns like that... but possibly the conversion already happens somewhere implicitly in pyogrio.

Anyway, I think we agreed we shouldn't put too much effort anymore in the non-arrow code paths but if not too difficult to implement I think a string with comma's would be more logical, no? Or just fine to keep it like this, because I suppose changing it now could give backwards-compatibility issues for people somehow using this already?

theroggy: It looks like an implicit conversion of a numpy array by default doesn't have commas... so I'm using json.dumps now for array serialization...

jorisvandenbossche: I am inclined to leave it as is, as you said above:

>  Or just fine to keep it like this, because I suppose changing it now could give backwards-compatibility issues for people somehow using this already?

I think we want to move towards unifying on the Arrow interface code path at some point in the future (or enabling it by default), and then we will already have a change in behaviour anyhow (to actually preserve lists instead of stringifying them). 
So now still changing how the string repr looks like, I don't know if that is worth it.

jorisvandenbossche: Essentially it depends on whether you have lists or numpy arrays as the elements in the column:

```python
>>> l = [1, 2, 3]
>>> str(l)
'[1, 2, 3]'
>>> str(np.array(l))
'[1 2 3]'
```

I agree this is a bit strange that this gives a difference. But then again people could also have tuples, and that would still give yet another string ..

theroggy: It is in the context of the unification that it might be useful: if the behaviour between both options (with arrow or without) is as similar as possible, switching to arrow becomes easier. But of course these list-datatypes are not a common thing... so I agree it isn't that important either.

Regardless of the conclusion, the change definitely doesn't belong in this PR... so I reverterted it.

In another branch I had been doing some more tests and also added .gpkg as output format in the test because geojson is quite a specific case for many data types being tested here... I'll make PR out of that and then we'll see where it leads...

jorisvandenbossche (approved):
Thanks!

PR #619, opened 2025-11-30
DOC: improvements to inline docs/comments

theroggy:
General improvements to mainly docstrings + add missing docstrings.


jorisvandenbossche (commented):
@theroggy thanks for working on the docs! 
To be fair, I would personally not put too much effort in listing every parameter in the internal docs (when it is often not much more as repeating the type that is also in the signature, mostly in _io.pyx), but adding the short summary lines is definitely a nice improvement (and since you already did the work, I won't further complain about it ;))

PR #630, opened 2026-01-19
BUG: fix writing some non-string object columns with arrow

### LINKED ISSUE #631: BUG: writing a column with Path objects gives an error with use_arrow
Without use_arrow the Path objects are simply serialised to string and written, but with use_arrow this leads to an error like this:

> tests/test_predict.py::test_predict_use_cache_skip[False-True-12585] - RuntimeError: ERROR in load_images for footballfields_BEFL-2019_test: ("Could not convert PosixPath('/tmp/pytest-of-runner/pytest-0/test_predict_use_cache_skip_Fa1/sample_projects/_image_cache/BEFL-2019/1024x1024_128pxOverlap/174816/174816_176352_175136_176672_1280_1280.jpg') with type PosixPath: did not recognize Python value type when inferring an Arrow data type", 'Conversion failed for column path with type object')

Reference: https://github.com/orthoseg/orthoseg/issues/287

theroggy:
In `write_dataframe` with `use_arrow=False`, object dtype columns are implicitly serialised to string to be able to write them to the output file.

With `use_arrow=True`, the pyarrow `to_table` function supports treatment of some datatypes (e.g. lists,...), but for other cases such columns rather given an error. For datatypes that are not supported by pyarrow a good default behaviour would be to just convert them to string as is done without arrow.

This PR explicitly converts object columns to string for columns that aren't supported to be interpreted by pyarrow.

resolves #631

### DIFF
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -802,7 +802,7 @@ def write_dataframe(
                     # The arrow timestamp type doesn't support mixed time zone offsets,
                     # so convert to string to avoid data loss and pass on to GDAL
                     # that it is a datetime so GDAL can preserve the type information.
-                    df[name] = df[name].astype("str")
+                    df[name] = df[name].astype("string")
                     datetime_cols.append(name)
                 else:
                     # Check if pyarrow can handle the data in the column. If not,
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -798,17 +798,22 @@ def write_dataframe(
             if dtype == "object":
                 # An object column with datetimes can contain multiple offsets.
                 inferred_dtype = pd.api.types.infer_dtype(df[name])
-                if inferred_dtype in {"mixed", "mixed-integer", "mixed-integer-float"}:
-                    # mixed is an unknown object column, convert to string
-                    df[name] = df[name].astype("string")
+                if inferred_dtype in {"mixed-integer", "mixed-integer-float"}:
+                    # just convert these mixed columns to string
+                    df[name] = df[name].astype("str")
+                elif inferred_dtype == "mixed":
+                    # mixed is an unknown object column... if it isn't a list, convert
+                    # to string
+                    if not isinstance(df[name].dropna().iloc[0], list | np.ndarray):
+                        df[name] = df[name].astype("str")
                 elif inferred_dtype == "datetime":
-                    df[name] = df[name].astype("string")
+                    df[name] = df[name].astype("str")
                     datetime_cols.append(name)
 
             elif isinstance(dtype, pd.DatetimeTZDtype) and str(dtype.tz) != "UTC":
                 # A pd.datetime64 column with a time zone different than UTC can contain
                 # data with different offsets because of summer/winter time.
-                df[name] = df[name].astype("string")
+                df[name] = df[name].astype("str")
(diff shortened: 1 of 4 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -2541,6 +2530,87 @@ def test_write_read_null(tmp_path, use_arrow):
...
+        "geometry": [geom] * len(object_col_data),
+        "object_col": object_col_data,
+    }
+    test_gdf = gp.GeoDataFrame(test_data, crs="epsg:31370")
+
+    # Verify that object_col is actually inferred as object dtype for this test.
+    str_dtype = (
+        "str"
+        if PANDAS_GE_30 or (PANDAS_GE_23 and pd.options.future.infer_string)
+        else "object"
```
jorisvandenbossche: Instead of verifying this here, I would instead specify `dtype=object` in the construction above to ensure the input data is always using object dtype (also for purely strings). 

Strings as the `str` dtype is already covered by other tests (and also does not go through the code path you changed), so I would have this test just focus on object dtype (and we should cover the case of all strings as object dtype anyway as well)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -2606,6 +2595,83 @@ def test_write_read_null(tmp_path, use_arrow):
...
+            expected_dtype = "datetime64[ms]" if PANDAS_GE_20 else "datetime64[ns]"
+            expected_data = [pd.Timestamp(value) for value in object_col_data]
+        elif isinstance(object_col_data[0], bytes | list):
+            # These types are read back as object type with arrow
+            expected_dtype = "object"
+            expected_data = object_col_data
+        elif isinstance(object_col_data[0], Decimal):
+            # Decimal objects are read back as decimal objects with arrow
+            expected_dtype = "float64"
+            expected_data = object_col_data
```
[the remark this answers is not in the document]
theroggy: Indeen... the == just works to compare Decimal values with float values. Fixed comment.

pyogrio/geopandas.py
```diff
@@ -793,11 +794,26 @@ def write_dataframe(
...
+                    # that it is a datetime so GDAL can preserve the type information.
                     df[name] = df[name].astype("string")
                     datetime_cols.append(name)
+                else:
+                    # Check if pyarrow can handle the data in the column. If not,
+                    # convert it to string and save it like that.
+                    try:
+                        _ = pa.Schema.from_pandas(df[[name]])
+                    except pa.ArrowInvalid:
+                        df[name] = df[name].astype("str")
```
```suggestion
                        df[name] = df[name].astype("string")
```

Same as what is done above for datetime, this might fix the failure for the build with older pandas.

The details are that `astype('str')` with pandas <3 was converting every element to a string, including missing values, and then put it in an object-dtype column. With pandas >3, it now does the "correct" thing (preserve missing values) and cast to the new str dtype. 
For pandas <3, we can use `astype("string")` here, which uses the opt-in dtype using pd.NA,  but since we just convert it for then directly writing it to GDAL, that shouldn't matter, I think

theroggy: Thanks, interesting to know. It did indeed solve 2 out of 3 issues.

jorisvandenbossche (approved):
Looks good! Sorry for the delay ;)

PR #633, opened 2026-02-18
TST: fixes to tests for pandas 3.0

theroggy:
In this PR:
- remove `pandas_future_infer_string=0` environment variable used when running conda tests now pandas 3 is released
    - remark: this lead to some fails when `use_arrow=True` results are compared with `use_arrow=False`, as arrow seems to ignore the environment variable?
- some fixes in tests where combinations of old gdal versions with newer pandas versions gave some issues


jorisvandenbossche (approved):
Looks good, thanks!

PR #634, opened 2026-02-19
BUG: fix error writing a column with mix of datetimes with and without timezone without arrow

theroggy:
Fixes following errors:
- Error occured in `write_dataframe` (with `use_arrow=False`) when writing a dataframe with a datetime column where  all datetimes have no time zone offset: ``TypeError: Invalid dtype float64 for __floordiv__```
- Error occured in `write_dataframe` (with `use_arrow=False`) when writing a dataframe with a datetime column where  some datetimes have a time zone offset and some don't have an offset: ```ValueError: cannot convert float NaN to integer```

Issue noticed in https://github.com/geofileops/geofileops/issues/802

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,8 +4,9 @@
 
 ### Bug fixes
 
--   Fix error in `write_dataframe` with `use_arrow=False` when writing a column with a
-    mix of datetimes with and without time zones (#).
+-   Fix error in `write_dataframe` with `use_arrow=False` when writing a column with
+    datetimes without any time zone offsets or with a mix of offsets and no offsets
+    (#634).
 
 ## 0.12.1 (2025-11-28)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,12 +1,19 @@
 # CHANGELOG
 
+## 0.12.2 (????-??-??)
+
+### Bug fixes
+
+-   Fix error in `write_dataframe` with `use_arrow=False` when writing a column with a
+    mix of datetimes with and without time zones (#).
+
 ## 0.12.1 (2025-11-28)
 
 ### Bug fixes
 
--   Fix regression in reading date columns (#616)
+-   Fix regression in reading date columns (#616).
 -   Fix regression in `read_dataframe` when `use_arrow=True` and `columns` is used to filter
-    out columns of some specific types (#611)
+    out columns of some specific types (#611).
 
 ## 0.12.0 (2025-11-26)
 
--- a/pyogrio/geopandas.py
+++ b/pyogrio/geopandas.py
@@ -893,7 +893,13 @@ def write_dataframe(
             # Column of Timestamp/datetime objects, split in naive datetime and tz.
             if pd.api.types.infer_dtype(df[name]) == "datetime":
                 tz_offset = col.map(lambda x: x.utcoffset(), na_action="ignore")
-                gdal_offset_repr = tz_offset // pd.Timedelta("15m") + 100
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+-   Fix error in `write_dataframe` with `use_arrow=False` when writing a column with
```
jorisvandenbossche:
```suggestion
-   Fix error in `write_dataframe` with `use_arrow=False` when writing an object-dtype column with
```

(to make it clear this is not a typical "datetime" column without time zone offsets, because that would be a datetime64 column which works fine)

pyogrio/geopandas.py
```diff
@@ -906,7 +906,13 @@ def write_dataframe(
             # Column of Timestamp/datetime objects, split in naive datetime and tz.
             if pd.api.types.infer_dtype(df[name]) == "datetime":
                 tz_offset = col.map(lambda x: x.utcoffset(), na_action="ignore")
```
```suggestion
                tz_offset = col.map(lambda x: x.utcoffset(), na_action="ignore").astype("timedelta64[us]" if PANDAS_GE_30 else "timedelta64[ns]") 
```

If you actually have offsets, pandas will already have inferred it as that timedelta64 dtype (and this astype will be a no-op), and if not and you got None values, this should ensure it gets converted to a proper dtype with NaT values, and then the below division will work fine with the faster vectorized operation (so then it should not be needed to fall back to the slower `map`)

pyogrio/geopandas.py
```diff
@@ -906,7 +906,13 @@ def write_dataframe(
...
-                gdal_offset_repr = tz_offset // pd.Timedelta("15m") + 100
+                gdal_offset_repr = tz_offset.map(
+                    lambda x: x // pd.Timedelta("15m") + 100, na_action="ignore"
+                )
+                # Replace pd.NA with np.nan because the cython code expects numpy arrays
+                gdal_offset_repr = gdal_offset_repr.where(
+                    ~pd.isna(gdal_offset_repr), np.nan
```
jorisvandenbossche: I would not expect this to be needed (by default, pandas should not yet be using pd.NA (unless the user uses it explicitly), and since this path is only for object dtype, the user is not using an opt-in nullable dtype)

But maybe my suggestion above to force to use "timedelta64" might also solve it

pyogrio/tests/test_geopandas_io.py
```diff
@@ -1001,6 +1001,61 @@ def test_write_read_datetime_tz_mixed_offsets(
...
+    reason="before GDAL 3.11, datetimes weren't handled as well",
+)
+def test_write_read_datetime_tz_offsets_None(tmp_path, dates, use_arrow):
+    """Test writing a column with datetimes with and without time zone offsets.
+
+    Two types of errors occured:
+      - For datetimes without any offset, when written without arrow:
+        `TypeError: Invalid dtype float64 for __floordiv__`
+      - For datetimes with a mix of having a time zone offset and without, when written
+        without arrow: `ValueError: cannot convert float NaN to integer`
```
jorisvandenbossche: I don't think it is needed to keep this detail here in the test, because for a future reader that should not actually be anymore in the code (and the details are in the PR)

pyogrio/geopandas.py
theroggy: It indeed doesn't seem to be needed anymore, so removed.

jorisvandenbossche (approved):
Looks good!

PR #635, opened 2026-02-19
BUG: fix wrong handling of some windows network paths

### LINKED ISSUE #632: BUG: Windows UNC paths are parsed as URI
When I use `gpd.read_file` with a Windows UNC path that contains an exclamation mark (e.g. `\\server\!test\example.shp`), the path is mishandled. It is sent through `_parse_uri` which splits on the "!" character and drops the beginning of the path.

theroggy:
TODO: on linux, the same issue occurs but this isn't being fixed yet in the current state of the PR...

Resolves #632


theroggy:
> @theroggy have you seem my comment on the issue? Because if this is in fact not specific to Windows, we might want to fix it differently, such that it is also fixed in the code path that uses `_parse_uri`

Yes, I saw, but didn't get to it yet to have a closer look...

jorisvandenbossche:
(pushing this to the next release, so I can start the current one)

PR #649, opened 2026-04-24
BUG: Allow building with pre-release versions of GDAL

### LINKED ISSUE #648: BUG: Can't build with pre-release versions of GDAL
`setup.py` checks that the GDAL version is above the minimum, but the parsing logic doesn't seem to work for the current GDAL versioning scheme.

Over the last year we've seen these naming conventions for GDAL pre-releases:

- v3.x.xbeta0
- v3.x.xbeta1
- v3.x.xbeta2
- v3.x.xRC1
- v3.x.xRC2
- v3.x.xrc0
- v3.x.x-r1

However, we're only doing `.strip("dev")` before parsing:

https://github.com/geopandas/pyogrio/blob/e3eb27f955668975e298f38aa0ba7c37bbae1e4f/setup.py#L156

Here's a minimal example:

```python
Python 3.15.0a8 (tags/v3.15.0a8:55ea59e, Apr 23 2026, 18:27:26) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import gdal
>>> gdal.__version__
'3.13.0beta2'
>>> gdal_version_str = gdal.__version__
>>> gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
  File "<python-input-3>", line 1, in <module>
    gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
                   ^^^^^
  File "<python-input-3>", line 1, in <genexpr>
    gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
                         ~~~^^^
ValueError: invalid literal for int() with base 10: '0beta2'
>>> 
```

p-vdp:
Fixes  #648

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,6 +129,35 @@ def get_gdal_config():
             raise e
 
 
+def parse_gdal_version(version_str: str) -> tuple[int, int, int]:
+    # assume three elements separated by periods
+    major, minor, patch = version_str.split(".")[:3]
+    ver_strs = {"major": major, "minor": minor, "patch": patch}
+
+    # remove non-digits and convert to int
+    ver_ints = dict()
+    for k, v in ver_strs.items():
+        try:
+            ver_ints[k] = int(v)
+        except ValueError:
+            v_numeric = list()
+            for i, c in enumerate(v):
+                if c.isdigit():
+                    v_numeric.append(c)
+                elif len(v_numeric) > 0:
+                    break
+                else:
+                    pass
+            ver_ints[k] = int("".join(v_numeric))
+
+    # check output before returning
+    assert len(ver_ints) == 3
+    for v in ver_ints.values():
+        assert isinstance(v, int)
+
+    return ver_ints["major"], ver_ints["minor"], ver_ints["patch"]
+
+
 ext_modules = []
 package_data = {}
 
@@ -153,7 +182,7 @@ else:
 
     ext_options, gdal_version_str = get_gdal_config()
 
-    gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
+    gdal_version = parse_gdal_version(gdal_version_str)
     if not gdal_version >= MIN_GDAL_VERSION:
         sys.exit(f"GDAL must be >= {'.'.join(map(str, MIN_GDAL_VERSION))}")
 
--- a/setup.py
+++ b/setup.py
@@ -130,18 +130,18 @@ def get_gdal_config():
 
 
 def parse_gdal_version(version_str: str) -> tuple[int, int, int]:
-    # assume three elements separated by periods
+    # assume (at least) three elements separated by periods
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -153,7 +186,7 @@ def get_gdal_config():
 
     ext_options, gdal_version_str = get_gdal_config()
 
-    gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
+    gdal_version = parse_gdal_version(gdal_version_str)
```
martinfleis:
```suggestion
    gdal_version = Version(gdal_version_str).base_version
```

This is a solved problem. This should work if you add this on top. Then there's no custom code needed.

```py
from packaging.version import Version
```

This is a standard way of parsing versions across the ecosystem.

martinfleis (commented):
I think that relying on `packaging`, which we already depend on should be sufficient, no? Are there some forms of GDAL version string that are not parsable by `Version`?

setup.py
p-vdp: Excellent point, thanks!

setup.py
```diff
@@ -153,7 +154,7 @@ def get_gdal_config():
 
     ext_options, gdal_version_str = get_gdal_config()
 
-    gdal_version = tuple(int(i) for i in gdal_version_str.strip("dev").split("."))
+    gdal_version = (Version(gdal_version_str).major, Version(gdal_version_str).minor, Version(gdal_version_str).micro)
```
theroggy:
```suggestion
    gdal_version = Version(gdal_version_str).release
```
This is an even simpler way to get the version tuple...

theroggy (commented):
Nice... already a lot more readable using `Version`... 

This is a bit simpler still I think...

theroggy (approved):
Thanks!

martinfleis (approved):

jorisvandenbossche:
> I think that relying on `packaging`, which we already depend on should be sufficient, no? Are there some forms of GDAL version string that are not parsable by `Version`?

`packaging` is for Python packaging, and while versions are standardized for that, I don't think that GDAL is "obliged" to follow that (and no idea if there are other software ecosystems which are less strict about versioning schemes, or use other identifiers, ..), so in theory they could publish a package with a Python-incompatible version. 

In practice that is maybe quite unlikely, and looking at the tags listed in the issue https://github.com/geopandas/pyogrio/issues/648, that seems all good.

jorisvandenbossche:
@p-vdp thanks for the PR!

p-vdp:
@jorisvandenbossche yeah I suppose there's nothing to stop the GDAL devs from going off the rails and abandoning this versioning scheme for the mainline GDAL binaries, but as long as they have Python bindings on PyPI it would be very strange for them to do so. 

Thanks for the merge!

PR #650, opened 2026-05-28
BUG: fix attribute data being written incorrectly  with KML driver

### LINKED ISSUE #3609: BUG: to_file(filename, driver='kml') results in offset columns
- [x ] I have checked that this issue has not already been reported.

- [x ] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
from shapely.geometry import Point

# build a dict of columns: 2 text, 1 int, 1 float, and the geometry
data = {
    "text_a":   ["foo",   "bar",   "baz"],
    "text_b":   ["lorem", "ipsum", "dolor"],
    "int_val":  [1,       2,       3],
    "float_val":[0.1,     0.2,     0.3],
    "geometry": [
        Point(-105.0, 40.0),
        Point(-110.0, 45.0),
        Point(-115.0, 50.0),
    ],
}

# create the GeoDataFrame (with WGS84 lon/lat)
gdf = gpd.GeoDataFrame(data, crs="EPSG:4326")

gdf.to_file("blah.kml", driver="kml")
with open("blah.kml") as ff:
    print(ff.read())

```

#### Problem description

In the ExtendedData tag, the first element has the name of the first column but it has the value of the 3rd column.
```xml
  <Placemark id="blah.1">
	<name>foo</name>
	<description>lorem</description>
	<ExtendedData><SchemaData schemaUrl="#blah">
		<SimpleData name="text_a">1</SimpleData>
		<SimpleData name="text_b">0.1</SimpleData>
	</SchemaData></ExtendedData>
      <Point><coordinates>-105,40</coordinates></Point>
  </Placemark>
```

#### Expected Output
The column names should match their values and not exclude the last two columns.

```xml
  <Placemark id="blah.1">
	<name>foo</name>
	<description>lorem</description>
	<ExtendedData><SchemaData schemaUrl="#blah">
		<SimpleData name="text_a">foo</SimpleData>
		<SimpleData name="text_b">lorem</SimpleData>
        <SimpleData name="int_val">1</SimpleData>
        <SimpleData name="float_val">0.1</SimpleData>
	</SchemaData></ExtendedData>
      <Point><coordinates>-105,40</coordinates></Point>
  </Placemark>
```

Since the first two columns are represented in the name and description tag, I'm OK with excluding them from ExtendedData but the important thing is that it isn't naming the first SimpleData by the first column but putting the third column's values there.

Using Fiona works though.

If I do
```python
import fiona
fiona.drvsupport.supported_drivers['KML'] = 'rw'
gdf.to_file("blah.kml", driver="KML", engine="fiona")
with open("blah.kml") as ff:
    print(ff.read())
```

```xml
	<ExtendedData><SchemaData schemaUrl="#blah">
		<SimpleData name="text_a">foo</SimpleData>
		<SimpleData name="text_b">lorem</SimpleData>
		<SimpleData name="int_val">1</SimpleData>
		<SimpleData name="float_val">0.1</SimpleData>
	</SchemaData></ExtendedData>
```
fiona doesn't populate a name or description tag which is fine with me.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.12.2 (main, Feb 25 2024, 16:35:05) [GCC 11.4.0]
executable : /home/dean/pyjs/generic/.venv/bin/python
machine    : Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.35

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.4
GEOS lib   : None
GDAL       : 3.10.3
GDAL data dir: /home/dean/pyjs/generic/.venv/lib/python3.12/site-packages/pyogrio/gdal_data/
PROJ       : 9.3.0
PROJ data dir: /home/dean/pyjs/generic/.venv/lib/python3.12/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.1.1
numpy      : 2.2.3
pandas     : 2.2.2
pyproj     : 3.6.1
shapely    : 2.0.7
pyogrio    : 0.11.0
geoalchemy2: 0.15.2
geopy      : None
matplotlib : 3.9.2
mapclassify: None
fiona      : 1.10.1
psycopg    : 3.2.3
psycopg2   : 2.9.10 (dt dec pq3 ext lo64)
pyarrow    : 20.0.0

</details>

theroggy:
Resolves https://github.com/geopandas/geopandas/issues/3609

### DIFF
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -3059,6 +3059,9 @@ def test_write_kml(tmp_path, kml_driver, use_arrow):
 
     Test added when fixing https://github.com/geopandas/geopandas/issues/3609
     """
+    if kml_driver not in list_drivers():
+        pytest.skip(f"{kml_driver} driver not available")
+
     df = gp.GeoDataFrame(
         {"col_1": [1.0, 2.0], "col_2": [3.0, 4.0], "col_3": [5.0, 6.0]},
         geometry=[Point(0, 0), Point(1, 1)],
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -3080,7 +3080,11 @@ def test_write_kml(tmp_path, kml_driver, use_arrow):
     try:
         result_df = result_df[df.columns]
     except KeyError as ex:
-        raise RuntimeError(f"Keyerror for {df.columns=} on {result_df=}") from ex
+        with open(output_path) as f:
+            content = f.read()
+        raise RuntimeError(
+            f"Keyerror for {df.columns=} on {result_df=}, {content=}"
+        ) from ex
     assert_geodataframe_equal(result_df, df, check_index_type=False)
 
 
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -1198,7 +1198,8 @@ def test_write_read_column_names_casing(tmp_path, ext, use_arrow, expected_resul
 
     filename = tmp_path / f"test_duplicate_columns{ext}"
     if expected_result == "error":
-        with pytest.raises(Exception, match="Error .* field .*'COL' .*"):
+        with pytest.raises(Exception, match="COL"):
+            # The message depends on the driver and GDAL version, keep match simple.
             write_dataframe(gdf, filename, use_arrow=use_arrow)
         return
 
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -3045,6 +3045,39 @@ def test_arrow_enable_with_environment_variable(tmp_path):
             _ = read_dataframe(test_path, encoding="cp1252")
 
 
+@pytest.mark.requires_arrow_write_api
+@pytest.mark.parametrize("kml_driver", ["LIBKML", "KML"])
+def test_write_kml(tmp_path, kml_driver, use_arrow):
+    """Test writing a KML file.
+
+    A KML file is a bit of a special case, because when it is written, some extra
+    columns are added automatically in the layer definition: "Name" and "Description".
+    Because these extra columns are the first columns in the layer definition, it is
+    important to explicitly check the column index when writing values to fields as
+    you cannot rely on the index to be the same as the order fields were added to a
+    layer.
+
+    Test added when fixing https://github.com/geopandas/geopandas/issues/3609
+    """
+    df = gp.GeoDataFrame(
+        {"col_1": [1.0, 2.0], "col_2": [3.0, 4.0], "col_3": [5.0, 6.0]},
+        geometry=[Point(0, 0), Point(1, 1)],
+        crs="EPSG:4326",
+    )
+
+    output_path = tmp_path / "test.kml"
+    write_dataframe(df, output_path, driver=kml_driver, use_arrow=use_arrow)
+
+    assert output_path.exists()
+
+    result_df = read_dataframe(output_path)
+
+    # In a KML, there are several columns that are added automagically... so only check
+    # the columns we wrote.
+    result_df = result_df[df.columns]
+    assert_geodataframe_equal(result_df, df, check_index_type=False)
+
+
 @pytest.mark.filterwarnings("ignore:File /vsimem:RuntimeWarning")
 @pytest.mark.parametrize("driver", ["GeoJSON", "GPKG"])
 def test_write_memory(naturalearth_lowres, driver):
--- a/pyogrio/tests/test_geopandas_io.py
+++ b/pyogrio/tests/test_geopandas_io.py
@@ -292,12 +292,13 @@ def test_read_geojson_error(naturalearth_lowres_geojson, use_arrow):
 
 @pytest.mark.skipif(
     "LIBKML" not in list_drivers(),
-    reason="LIBKML driver is not available and is needed to read simpledata element",
+    reason="LIBKML driver is not available and is needed to read attribute columns",
 )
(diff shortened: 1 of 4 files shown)


pyogrio/tests/test_geopandas_io.py
```diff
@@ -3045,6 +3099,46 @@ def test_arrow_enable_with_environment_variable(tmp_path):
...
+    output_path = tmp_path / "test.kml"
+    write_dataframe(df, output_path, driver=kml_driver, use_arrow=use_arrow)
+
+    assert output_path.exists()
+
+    result_df = read_dataframe(output_path)
+
+    # In a KML, there are several columns that are added automagically... so only check
+    # the columns we wrote.
+    result_df = result_df[df.columns]
```
jorisvandenbossche, line 3215: Do we still want to verify that those other columns are indeed present? (or does that vary across gdal / libkml versions?)

pyogrio/tests/test_geopandas_io.py
```diff
@@ -3045,6 +3099,46 @@ def test_arrow_enable_with_environment_variable(tmp_path):
             _ = read_dataframe(test_path, encoding="cp1252")
 
 
+@pytest.mark.requires_arrow_write_api
+@pytest.mark.parametrize("kml_driver", ["LIBKML", "KML"])
+@pytest.mark.skipif(
+    "LIBKML" not in list_drivers(),
+    reason="LIBKML driver is not available and is needed to read simpledata element",
+)
```
jorisvandenbossche, line 3184: That is because we don't build with libexpat, and so the built-in KML driver is then write-only?

pyogrio/tests/test_geopandas_io.py
```diff
@@ -1169,6 +1169,60 @@ def test_write_read_datetime_utc(
...
+        (".shp", True, "second_column_dropped"),
+    ],
+)
+def test_write_read_column_names_casing(tmp_path, ext, use_arrow, expected_result):
+    """Test writing and reading a file with column names that only differ in casing.
+
+    Probably never a good idea to use multiple columns with the same name but different
+    casing, but at least this test documents the current behaviour.
+
+    With arrow, this never seems to be supported.
```
jorisvandenbossche, line 1195: Worth raising to GDAL? (if a format like JSON supports case-sensitive column names, there is nothing inherent to the Arrow interface that should prevent using this)

jorisvandenbossche (commented):
Looks good!

pyogrio/tests/test_geopandas_io.py
theroggy: No, with our wheels reading is fine because if the LIBKML driver is present, GDAL defaults to LIBKML when reading .kml files. (And because there is no `driver` parameter exposed in `read_dataframe` we cannot force to use the "KML" driver to test this explicitly).

In the GDAL docker images however, LIBKML is apparently not available... (but libexpat is), so you can read the .kml file but the "simpledata elements" (=the alphanumeric columns in the dataset) are just ignored when reading, even though they are written correctly with the "KML" driver.

I had started to create an issue in GDAL to either suggest adding support for it or at least documenting this behaviour, but because I accidentally lost the text I had written and because the LIBKML driver does it fine I didn't report it in the end. I'll try to create it anyway on of the coming days as it is quite odd behaviour.

jorisvandenbossche: >  but the "simpledata elements" (=the alphanumeric columns in the dataset) are just ignored when reading

So the KML driver does not read any typical "attribute" information? (and so with that driver we can't test that those were written)

theroggy: Indeed. I opened an issue in GDAL: https://github.com/OSGeo/gdal/issues/14698

pyogrio/tests/test_geopandas_io.py
theroggy: The LIBKML driver returns a lot more columns than the KML driver...:
- LIBKML: 'id', 'Name', 'description', 'timestamp', 'begin', 'end', 'altitudeMode', 'tessellate', 'extrude', 'visibility', 'drawOrder', 'icon'
- KML: 'Name', 'description'

Hence, I think testing on those will only create a maintenance burden if there are changes in the columns being added change and as we don't do anything with those columns we would only be "man in the middle"... as any fixes will nead to be done in GDAL?

jorisvandenbossche: Yep, then for sure leave it as is

pyogrio/tests/test_geopandas_io.py
```diff
@@ -3045,6 +3099,46 @@ def test_arrow_enable_with_environment_variable(tmp_path):
             _ = read_dataframe(test_path, encoding="cp1252")
 
 
+@pytest.mark.requires_arrow_write_api
+@pytest.mark.parametrize("kml_driver", ["LIBKML", "KML"])
+@pytest.mark.skipif(
+    "LIBKML" not in list_drivers(),
+    reason="LIBKML driver is not available and is needed to read simpledata element",
```
jorisvandenbossche:
```suggestion
    reason="LIBKML driver is not available and is needed to read attribute columns",
```

(for me that would be clearer, because "simpledata element" is an internal format detail on how those attribute columns are represented)

theroggy: OK, changed

jorisvandenbossche (approved):

jorisvandenbossche:
@theroggy thanks!

Issue #103, opened 2022-05-14
Expose GDAL driver metadata


jorisvandenbossche:
Would we want something like this in fiona: https://github.com/Toblerity/Fiona/pull/950 ?

It would allow to check the extension of a driver, the open options, whether the driver supports certain data types, etc (but it would probably be mostly copy/paste from fiona ..)

theroggy:
Sounds like it is at least somewhat linked to #81 and #71

jorisvandenbossche:
> linked to https://github.com/geopandas/pyogrio/issues/81

Yes, if we have this, it could indeed be used for #81. But I think it's also certainly fine to already do something about #81 without this feature (this issue is also meant for user facing API, while for the tests we can hard-code a bunch of things on the short term)

theroggy:
I wonder if it would be possible to get more driver metadata into gdal... this way it is available centrally instead of having copies of this metadata in every python library + in every other programming language.

theroggy:
> Which additional information are you thinking about? (the current GDAL metadata already exposes quite a lot: extensions, which data types are supported, whether nulls are supported, ..)

Ah, OK, I wasn't aware of that. In that case, ignore my remark :-).

jorisvandenbossche:
I started using the GDAL driver metadata internally in https://github.com/geopandas/pyogrio/pull/189, but without publicly exposing it for now.

PR #656, opened 2026-06-03
ENH: add list_drivers_details function

### LINKED ISSUE #647: Add a function to retrieve human-readable driver names for geodata export
I’m building a small desktop app that allows exporting geodata files, and I want to provide users with the maximum number of writing options.

Currently, list_drivers() returns short names, which is fine for most use cases, but not ideal when you want to display easily readable names to users.

It would be helpful to have another function, with another data structure that provides more readable driver names alongside the short names.

### LINKED ISSUE #657: BUG: PGDUMP labeled as 'rw' in list_driver
In list_driver PG_DUMP is declared read and write but it is not the case in gdal.GetMetadata
```
import pyogrio as pyg
from osgeo import ogr
import pprint as pri
drivers = pyg.list_drivers()

for driver in drivers:
    driver_gdal = ogr.GetDriverByName(driver)
    create = driver_gdal.GetMetadataItem("DCAP_CREATE")
    open_ = driver_gdal.GetMetadataItem("DCAP_OPEN")
    test = create=='YES' and open_ is None
    if  test:
        print(f"""
driver short name : {driver}
define in list_drivers() as : {drivers[driver]}
"""
)
```
it depends on ogr_list_driver  which doesn't check only if a driver support writing, default is read-only: 
```
def ogr_list_drivers():
    """List all available OGR drivers with read/write mode.

    Returns
    -------
    dict
        Dictionary with the driver name as key and the supported modes as value
        ("r" or "rw")

    """
    cdef OGRSFDriverH driver = NULL
    cdef int i
    cdef char *name_c

    drivers = dict()
    for i in range(OGRGetDriverCount()):
        driver = OGRGetDriver(i)
        name_c = <char *>OGR_Dr_GetName(driver)

        name = get_string(name_c)

        if ogr_driver_supports_write(name):
              """ Return True if _get_driver_metadata_item(driver, "DCAP_CREATE") == "YES """

            drivers[name] = "rw"

        else:
            drivers[name] = "r"

    return drivers
```

theroggy:
Resolves #647 
Resolves #657

Reference #103


pyogrio/_ogr.pyx
```diff
@@ -211,6 +233,37 @@ def ogr_list_drivers():
...
+        driver = OGRGetDriver(i)
+
+        name_c = <char *>OGR_Dr_GetName(driver)
+        name = get_string(name_c)
+
+        drivers[name] = {
+            "long_name": _get_driver_metadata_item(name, "DMD_LONGNAME"),
+            "create": ogr_driver_supports_write(name),
+            "update": ogr_driver_supports_update(name),
+            "append": ogr_driver_supports_append(name),
```
martinfleis: reading support is implied as always True?

armadillo-mononoke: since this code execute perfectly, yes I suppose.
```
import pyogrio as ogr
foo = list(ogr.list_drivers().values())
foo.sort()
bar = [driver for driver in ogr.list_drivers().values() if "r" in driver]
bar.sort()
assert foo == bar
```

theroggy: Good point. The existing code in list_drivers assumed this was the case, but I just doublechecked, but apparently there is indeed a write-only driver, PGDUMP, so I added a test on the "open" driver capability...

pyogrio/_ogr.pyx
```diff
@@ -199,14 +243,54 @@ def ogr_list_drivers():
...
         name_c = <char *>OGR_Dr_GetName(driver)
-
         name = get_string(name_c)
 
+        acces_modes = ""
+        if ogr_driver_supports_open(name):
+            acces_modes += "r"
         if ogr_driver_supports_write(name):
-            drivers[name] = "rw"
+            acces_modes += "w"
```
jorisvandenbossche: Personally, I would still add the "append" here as well, as you did for the original PR (it is a change to an existing function, but generally people should check for "r" or "w" being present in the string (which should still work), and this makes the output here consistent with what fiona returns?)

pyogrio/_ogr.pyx
```diff
@@ -199,14 +243,54 @@ def ogr_list_drivers():
...
+    drivers = dict()
+    for i in range(OGRGetDriverCount()):
+        driver = OGRGetDriver(i)
+
+        name_c = <char *>OGR_Dr_GetName(driver)
+        name = get_string(name_c)
+
+        extensions = _get_driver_metadata_item(name, "DMD_EXTENSIONS")
+        if extensions is not None:
+            extensions = [f".{ext}" for ext in extensions.split(" ")]
```
jorisvandenbossche, line 318: Do we need to include the dot?

pyogrio/core.py
```diff
@@ -72,6 +73,32 @@ def list_drivers(read=False, write=False):
...
+    For each driver, the following properties are included:
+
+    - long_name: the long name of the driver.
+    - open: a boolean indicating if the driver supports opening and reading an
+      existing file.
+    - create: a boolean indicating if the driver supports creation of new files.
+    - update: a boolean indicating if the driver supports updating an existing file,
+      including appending new rows. This property is None if GDAL < 3.11.
+    - append: a boolean indicating if the driver supports appending even though
+      it does not support updating. This property is None if GDAL < 3.12.
```
jorisvandenbossche:
```suggestion
      it does not support updating in general. This property is None if GDAL < 3.12.
```

(this is still vague, but otherwise it sounds strange as updating also includes appending)

pyogrio/core.py
```diff
@@ -72,6 +73,32 @@ def list_drivers(read=False, write=False):
...
+    - help_topic_url: a relative URL in the GDAL documentation to the help topic for
```
jorisvandenbossche: I think it would be nice to make this an actual working url (so prepend the relative URL with the base url for the stable docs?), so one can actually click it directly

jorisvandenbossche (commented):
Looks good!

pyogrio/_ogr.pyx
theroggy: @jorisvandenbossche  OK, I reopened the original PR, #559, for this as that will be easier because the specific tests,... for this change are there already. Once that one is merged I'll continue on this PR.

pyogrio/_ogr.pyx
theroggy: Having the dot in the extension/suffix is kind of the default in python I think, e.g. both the "old" [os.path.splitext](https://docs.python.org/3/library/os.path.html#os.path.splitext) and the new [Path.suffix](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix) include the dot?

docs/source/supported_formats.md
```diff
@@ -1,30 +1,34 @@
...
-## Support for other formats
+You can also use [Virtual File System](https://gdal.org/user/virtual_file_systems.html#virtual-file-systems),
+paths to read e.g. remote or zipped data sources and directories.
 
-Other vector formats that are registered within your particular installation of
-GDAL may be supported. Please be aware that these likely have not been tested
-for compatibility with Pyogrio and you may encounter specific issues with these
-formats and / or their constituent geometry or field data types.
+Most Pyogrio functionalities will most likely work for the other vector formats
+available in your local installation of GDAL. Nonetheles, please be aware that
```
jorisvandenbossche:
```suggestion
available in your local installation of GDAL. Nonetheless, please be aware that
```

pyogrio/core.py
```diff
@@ -82,6 +83,32 @@ def list_drivers(read=False, write=False, append=False):
...
+def list_drivers_details() -> dict[str, dict]:
+    """List all available drivers with detailed information.
+
+    For each driver, the following properties are included:
+
+    - long_name: the long name of the driver.
+    - read: a boolean indicating if the driver supports opening and reading an
+      existing file.
+    - append: a boolean indicating if the driver supports appending rows to an
+      existing file. This property is None if GDAL < 3.12.
```
jorisvandenbossche:
```suggestion
      existing file. This property is None if GDAL < 3.11.
```

(since this is now also "update", and I think we can disregard the detail it is only filled for < 3.12 for geojsonseq here, to avoid too much detail)

pyogrio/_ogr.pyx
```diff
@@ -247,15 +273,53 @@ def ogr_list_drivers():
...
+
+        drivers[name] = {
+            "long_name": _get_driver_metadata_item(name, "DMD_LONGNAME"),
+            "read": ogr_driver_supports_open(name),
+            "append": ogr_driver_supports_update(name) or ogr_driver_supports_append(name),
+            "write": ogr_driver_supports_write(name),
+            "supports_vsi": ogr_driver_supports_vsi(name),
+            "help_topic_url": _get_driver_metadata_item(name, "DMD_HELPTOPIC"),
```
jorisvandenbossche:
```suggestion
            "help_topic_url": "https://gdal.org/en/stable/" + _get_driver_metadata_item(name, "DMD_HELPTOPIC"),
```

pyogrio/tests/test_core.py
```diff
@@ -217,6 +231,47 @@ def test_list_drivers():
...
+            drivers[name]["help_topic_url"], str
+        )
+        assert drivers[name]["extensions"] is None or isinstance(
+            drivers[name]["extensions"], list
+        )
+        if drivers[name]["extensions"] is not None:
+            assert all(ext.startswith(".") for ext in drivers[name]["extensions"])
+
+
+def test_list_drivers_details_help_topic_url():
```
jorisvandenbossche, line 266:
```suggestion
@pytest.mark.network
def test_list_drivers_details_help_topic_url():
```

jorisvandenbossche (approved):
Thanks for the updates!

jorisvandenbossche:
For posterity: this PR originally included separately "update" and "append" capabilities in the details per driver, but @theroggy and I discussed that yesterday: since pyogrio's write API currently only support bulk writing or appending, we essentially don't have a differentiation between GDAL's update vs append (i.e. we don't (yet) allow updating existing rows, or "upsert"). So therefore we decided to expose the combo of both as "append" here. 
(in the future, we could add another field like "update" or "upsert" if that functionality would be added to the `write` path)

PR #663, opened 2026-06-06
MAINT: update ruff + fixes

theroggy:


pyogrio/tests/test_arrow.py
```diff
@@ -878,7 +885,7 @@ def test_write_non_arrow_tabular_data(tmp_path):
     data = pa.chunked_array([[1, 2, 3], [4, 5, 6]])
     with pytest.raises(
         DataLayerError,
-        match=".*should be called on a schema that is a struct of fields",
+        match=r".*should be called on a schema that is a struct of fields",
```
jorisvandenbossche: I don't fully understand why this is turned into a raw string, because in this case we do want the `.*` to be interpreted as regex pattern and not as a literal string?

(also, what is the rationale for when to use `re.escape` and when `r""`?)

jorisvandenbossche: OK, after reading the ruff page about this, I get it: the `r"` is to ensure it is unambiguous that it indeed _is_ a plain regex pattern, and not to be escaped (otherwise it is ambiguous the `.` is a regex meta-character or a plain character)

However, that made me also realize that I think in this case we don't even need the `.*`, as the match argument is passed to `search` and matches anywhere in the string (https://docs.pytest.org/en/latest/how-to/assert.html#matching-exception-messages):


```suggestion
        match="should be called on a schema that is a struct of fields",
```

jorisvandenbossche: (and same for some of the cases below)

theroggy: I simplified the match strings where possible...

jorisvandenbossche (approved):
Thanks!

PR #667, opened 2026-06-12
TST: update tests as reading .kml attribute columns is being supported from GDAL 3.14

theroggy:
Based on https://github.com/OSGeo/gdal/issues/14698, support to read .kml attribute columns was added to the GDAL "KML" driver.

This PR removes some skips in the tests for GDAL >= 3.14 so this behaviour is tested.


jorisvandenbossche:
I merged the other PR impacting GDAL 3.14 CI, so if you update here, we can see if CI is green now

theroggy:
I added the "skipped" summary to the conda tests as well to verify there as well if the tests aren't skipped anymore with conda-master. Afterwards I saw (and remembered) that there is no skipping there as the LIBKML driver is always available there... but I think it is still an improvement (even though it doesn't belong in this PR).

While busy, I added "xpassed" to the summary as well and synced the pytest command so the docker and conda tests show the same reporting.

jorisvandenbossche:
> added the "skipped" summary to the conda tests as well to verify ..
> ..
> While busy, I added "xpassed" to the summary as well

Sounds good! I was recently thinking to update that as well, also to add the summary of failures (which can be handy when there are failures, to quickly see all paths of failing tests)

jorisvandenbossche:
Thanks!

PR #681, opened 2026-07-03
MNT: Improve license metadata (PEP 639)

mwtoews:
This improves the licence metadata to align with [PEP 639](https://peps.python.org/pep-0639/):

- Specify license as a valid SPDX text identifier (the former table structure is deprecated)
- Remove license classifier, which is deprecated

These changes clear-up warnings from the package build. It also requires setuptools>=77.


jorisvandenbossche (approved):
Thanks!

PR #682, opened 2026-07-03
MNT: use dependency groups (PEP 735) + relax versioneer version requirement

mwtoews:
Use dependency groups ([PEP 735](https://peps.python.org/pep-0735/)). This is a standard Python feature that is best used to describe dependencies for tests and development. The former "test" extra is renamed to a "tests" group (plural, to match the `pyogrio/tests` subdir). The "dev" group is expanded to also include pre-commit and the tests group. Note that `uv sync` will include "dev" by default; but in the CI refactor it is explicitly done with `--group dev` for other front-ends with different defaults.

The "geopandas" extra remains with the published package, i.e. `pip install pyogrio[geopandas]`

Lastly, ignore `*.lock` files, e.g. from uv, pixi and others.


pyproject.toml
```diff
@@ -32,11 +33,25 @@ requires-python = ">=3.10"
...
+    "Cython>=3.1",
+    "versioneer[toml]==0.28",
+    # tomli is used by versioneer
+    "tomli; python_version < '3.11'",
+]
+benchmarks = ["pytest-benchmark"]
+tests = ["pytest", "pytest-cov"]
+dev = [
+    "pre-commit",
+    "pyogrio[geopandas]",
```
jorisvandenbossche: To what extent is this a problem (or unexpected) that it will then also install a _released_ (I assume) version of pyogrio when installing the dev dependencies?

mwtoews: This is admittedly a bit of a "hack" which seems to work, but could potentially break as there is no mechanism described in PEP 735 for self-references to either `optional-dependencies` (or `build-system`). Given it's only one item in that extra, it's not biggie to explicitly list `geopandas`.

mwtoews:
This PR also closes #683 to expand the versioneer version range slightly.
Also `versioneer[toml]` is sufficient enough to also install `tomli` if needed ([ref](https://github.com/python-versioneer/python-versioneer/blob/28c613dbef5fce09dc3ba6b1baa811c2d76b2245/pyproject.toml#L36-L37)) so this does not need to be specified as well.

pyproject.toml
jorisvandenbossche: In practice, also listing `geopandas` explicitly has this strange circularity, because geopandas depends on pyogrio. 

But regardless of that, I think listing geopandas explicitly here is clearer.

jorisvandenbossche: Based on the logs (and testing this PR locally with pip), this does seem to go fine and actually directly installs the editable pyogrio (and never first installs a released one)

jorisvandenbossche (approved):
Thanks! (and sorry for the slow follow-up)

PR #691, opened 2026-08-07
fix: register GDAL error handler before drivers so registration warnings are suppressible (#690)

### LINKED ISSUE #690: Unable to hide warning when using GDAL_SKIP
When you use `GDAL_SKIP` to disable a driver, pyogrio outputs a warning message that is difficult to hide.

(Originally I commented this on an existing issue, see below)

> I've been coming up against this issue when disabling drivers using `GDAL_SKIP` environment variable. Here's an example from Windows CLI, but I get the same thing in scripts and on Linux too.
> 
> ```
> c:\>set GDAL_SKIP=netCDF
> 
> c:\>python -c "import pyogrio"
> Warning 1: Unable to find driver netCDF to unload from GDAL_SKIP environment variable.
> ```
> 
> I've tried `gdal.quiet_errors()`, `warnings.filterwarnings("ignore")`, and `contextlib.redirect_stderr(None)` and I cannot make the warnings go away. The only thing that works is redirecting stderr from the commandline, but that kills all error output, not just this one message.
> 
> ```
> $ python myscript.py 2>/dev/null
> ``` 

 _Originally posted by @jontwo in [#289](https://github.com/geopandas/pyogrio/issues/289#issuecomment-5178228022)_

For context, we need to disable netCDF due to [CVE-2026-49014](https://nvd.nist.gov/vuln/detail/CVE-2026-49014). We're unable to upgrade to GDAL 3.13.1 so are disabling the driver for now.

patchwright:
## Problem
Fixes #690. When `GDAL_SKIP` names a driver GDAL doesn't recognize, `import pyogrio` prints a raw warning straight to stderr, and it cannot be suppressed with `warnings.filterwarnings`, `-W ignore`, or `redirect_stderr` -- none of the usual mechanisms work. Confirmed on master HEAD 99793df.

## Root cause
In `pyogrio/core.py`, `_register_drivers()` runs before `_register_error_handler()`. `_register_drivers()` calls `GDALAllRegister()`, which internally calls `AutoSkipDrivers()` and can emit a GDAL warning (e.g. "Unable to find driver X to unload from GDAL_SKIP environment variable") -- but at that point our error handler isn't installed yet, so GDAL's default handler prints it directly to stderr instead of routing it through `pyogrio._err.error_handler`, which is what turns GDAL warnings into catchable Python `RuntimeWarning`s.

## Fix
Swap the two calls so the error handler is registered first. `_register_error_handler()` (`CPLPushErrorHandler`) has no dependency on drivers being registered, so this is safe. 2 lines changed, no API changes.

## Backward compatibility
No breaking changes. One small behavior note: any other warning `GDALAllRegister()` might emit during driver registration (e.g. a plugin failing to load) will now also be routed through Python's warnings system as a `RuntimeWarning` instead of printing raw to stderr -- this is the intended effect of the fix, consistent with how every other GDAL warning in this library is already handled (see `test_error_handling_warning`).

## How to test
```bash
$ GDAL_SKIP=nonexistent_driver_xyz python -c "import pyogrio"
Warning 1: Unable to find driver nonexistent_driver_xyz to unload from GDAL_SKIP environment variable.

$ GDAL_SKIP=nonexistent_driver_xyz python -W ignore -c "import pyogrio"
# (nothing -- now suppressible)
```
Added `test_gdal_skip_unknown_driver_warning_suppressible` in `pyogrio/tests/test_core.py`, which fails on unpatched `main` (raw stderr output leaks through `-W ignore`) and passes with this fix. Full `pyogrio/tests/test_core.py` suite: 108 passed, 24 skipped, no regressions. Broader suite (excluding geopandas/arrow extras not installed locally): 313 passed, 71 skipped. `ruff format`/`ruff check` clean on the changed files (2 pre-existing `PLR0917` findings elsewhere in the file are unrelated to this diff).

---
Assisted-by: Claude (code generation, reviewed and tested locally)


fallenmi (changes requested):
The new GDAL_SKIP regression passes with this order, but moving the existing global handler before GDALAllRegister also hides registration failures.

Pyogrio's error_handler intentionally discards CE_Failure because normal call sites translate a returned error code into a Python exception. Here, _register_drivers wraps the void-returning GDALAllRegister, so there is no later translation. GDAL's AutoLoadDrivers emits CE_Failure when a discovered plugin cannot be loaded.

This was reproduced against the exact current base core.py (also the 0.13.0 wheel, GDAL 3.12.4) with an invalid gdal_broken.so in GDAL_DRIVER_PATH:

- Base order: import returns 0 and stderr reports the dlopen failures.
- PR order: import returns 0 and stderr is empty.

That turns a diagnosable broken plugin configuration into a silently missing driver. Please preserve CE_Failure output during GDALAllRegister while routing CE_Warning through Python warnings, for example with a registration-specific handler, and add a regression for a plugin-load failure. The proposed GDAL_SKIP test itself was independently confirmed to pass with the new order.

Review analysis and reproduction were performed with OpenAI Codex assistance under the account owner's authorization.

Issue #683, opened 2026-07-13
Is versioneer 0.28 still required?


mathomp4:
Sorry for the generic issue. I'm currently in spack concretization hell and the current issue is that pyogrio requires versioneer 0.28.

I'm building Pyogrio from source in a Spack-based scientific software stack.

Pyogrio 0.13.0 declares this build requirement:

```
  "versioneer[toml]==0.28",
```

This exact pin conflicts with other current Python packages in the same environment. For example, modern Dask pulls in Partd 1.4.2+, which requires versioneer[toml]==0.29.

Because both requirements are exact, package managers that try to share build dependencies must either install two Versioneer versions or fail dependency resolution. In our case, this also produces duplicate build-tool installations across a large stack.

Is the exact 0.28 pin still necessary for Pyogrio? If Pyogrio builds correctly with Versioneer 0.29, would you consider relaxing it to a compatible range, for example:
```
  "versioneer[toml]>=0.28,<0.30",
```
or otherwise updating the build requirement?

I'd gladly test but, well, this is the first time I've heard of pyogrio and even your README is like a magic text to me. 😄

jorisvandenbossche:
@mathomp4 thanks for asking! I don't think the exact pin is needed, will be relaxed with https://github.com/geopandas/pyogrio/pull/682

jorisvandenbossche:
Closed by https://github.com/geopandas/pyogrio/pull/682

PR #696, opened 2026-09-16
Expand allowable "versioneer" versions.

eap:
Python's versioneer is mature and has not produced a release in several years. Versioneer 0.28 and 0.29 are functionally identical for projects like this one (using the vendored "_version.py" and including a setup.py). So I have expanded the dependency version spec to allow for versioneer versions `>=0.28`

## Why?

I am installing pyogrio via its [spack-package](https://packages.spack.io/package.html?name=py-pyogrio) into a "[unified](https://spack.readthedocs.io/en/latest/environments.html#spec-concretization)" spack enviornment. Tightly constrained dependency versions (including build-time dependencies) create multi-version conflicts that have to be evaluated and whitelisted.

I have offered this change so that, if it is accepted, myself and other spack users can avoid whitelisting versioneer multi-version conflicts when installing pyogrio in the future. I have sent a [similar PR](https://github.com/dask/partd/pull/84) to `dask/partd`.


jorisvandenbossche:
@eap apologies! I merged https://github.com/geopandas/pyogrio/pull/682 (which also included a fix for this) without first checking other new PRs, so didn't notice your targeted PR for this.

But thanks for the contribution! Going to close this since this is already resolved now.

For reference, this was also reported in https://github.com/geopandas/pyogrio/issues/683

eap:
> @eap apologies! I merged #682 (which also included a fix for this) without first checking other new PRs, so didn't notice your targeted PR for this.

No apologies necessary; my bad for not noticing that this was a WIP. Anyhow, thank you for the fix!

PR #697, opened 2026-09-21
MAINT: Use PROJ_DATA instead of deprecated PROJ_LIB environment variable

mwtoews:
PROJ 9.1 introduced `PROJ_DATA` to deprecate `PROJ_LIB`, and plans to remove use of the older environment variable by PROJ 10.

Offer some backwards compatibility with a warning message, as it's possible to use older PROJ versions with modern GDAL versions.


jorisvandenbossche (approved):
Thanks!

